PR #45, opened 2026-02-02
Update GETTING_STARTED.md

johangithub:
I just went through the onboarding and noticed some road blocks, so updating the onboarding doc to reflect the fixes. A couple things I wasn't sure about
1. Do we want modal installed globally with pipx or do uv run modal in the repo?
2. What I got from the instruction was we set two separate R2 tokens, is that correct?
3. What do you think about using ENV variable and using a script to generate .tfvars?

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -134,6 +134,9 @@ Create an R2 API Token:
 2. Create an API key
 3. Note the **API Key** (starts with `sk-ant-`)
 
+> **Want to use your OpenAI ChatGPT subscription?** See [Using OpenAI Models](OPENAI_MODELS.md) for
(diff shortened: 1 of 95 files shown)


docs/GETTING_STARTED.md
```diff
@@ -138,7 +147,7 @@ access.
    - **Name**: `Open-Inspect-YourName` (must be globally unique)
    - **Homepage URL**: `https://open-inspect-{your-deployment-name}.vercel.app` (or your custom
      domain)
-   - **Webhook**: Uncheck "Active" (not needed)
+   - **Webhook**: Leave "Active" **unchecked**
```
ColeMurray: this shouldn't be needed

PR #71, opened 2026-02-06
Strengthen source-control provider foundation

ColeMurray:
## Summary
- add deployment-level `SCM_PROVIDER` resolution with a default of `github`
- route provider selection through the source-control factory and return explicit `501` for unimplemented providers on non-public routes
- document single-provider boundaries with an ADR and a provider contribution checklist
- add focused tests for provider env resolution and provider factory behavior

## Validation
- `npm -w @open-inspect/control-plane exec vitest run src/source-control/config.test.ts src/source-control/providers/index.test.ts`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,6 +67,13 @@ Use clear, descriptive commit messages:
 4. Update documentation if needed
 5. Provide a clear description of your changes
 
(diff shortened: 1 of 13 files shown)


ColeMurray:
Closing in favor of #72 (same change set, clean base on current main).

PR #119, opened 2026-02-17
feat: web security hardening (CSP headers, auth middleware, URL encoding)

wala0003:
## Summary

Adds several security improvements to the web package:

1. **URL parameter encoding** — Wraps dynamic route parameters with `encodeURIComponent()` to prevent path traversal/injection when constructing control-plane URLs
2. **Security response headers** — Adds CSP, X-Frame-Options (DENY), X-Content-Type-Options, Referrer-Policy, and Permissions-Policy headers via Next.js config
3. **Auth hardening** — Fail-fast validation for `NEXTAUTH_SECRET` at module load, plus GitHub token expiry detection with appropriate error responses
4. **Auth middleware** — New `middleware.ts` that enforces authentication at the edge for all non-public routes

## Motivation

These are defense-in-depth measures that prevent common web vulnerabilities:
- Path traversal via malicious session IDs
- Clickjacking (X-Frame-Options: DENY)
- XSS via MIME sniffing (X-Content-Type-Options: nosniff)
- Silent auth failures from expired tokens
- Unauthorized access to session pages when auth cookies are missing

### Commit 1: `encodeURIComponent` on dynamic URL parameters
- `packages/web/src/app/api/sessions/[id]/archive/route.ts`
- `packages/web/src/app/api/sessions/[id]/unarchive/route.ts`
- `packages/web/src/app/api/sessions/[id]/prompt/route.ts`
- `packages/web/src/app/api/sessions/[id]/ws-token/route.ts`
- `packages/web/src/components/secrets-editor.tsx` (apiBase + delete key)

### Commit 2: Security response headers
- `packages/web/next.config.ts` — CSP, X-Frame-Options, etc.

### Commit 3: Auth hardening
- `packages/web/src/lib/auth.ts` — fail-fast NEXTAUTH_SECRET check
- `packages/web/src/app/api/sessions/[id]/ws-token/route.ts` — token expiry warning
- `packages/web/src/app/api/sessions/route.ts` — token expiry 401 response

### Commit 4: Auth middleware
- `packages/web/src/middleware.ts` — new file, enforces auth on all protected routes

## Testing

- All 4 workspace packages pass `npm run typecheck`
- No breaking changes — all existing functionality preserved
- Security headers are additive (new behavior only)
- Middleware allows all currently-public paths through

### DIFF
--- a/packages/web/next.config.ts
+++ b/packages/web/next.config.ts
@@ -17,7 +17,7 @@ const nextConfig: NextConfig = {
             "script-src 'self' 'unsafe-inline' 'unsafe-eval'",
             "style-src 'self' 'unsafe-inline'",
             "img-src 'self' data: https://avatars.githubusercontent.com",
-            "font-src 'self' https://fonts.gstatic.com",
+            "font-src 'self'",
             `connect-src 'self' ${process.env.NEXT_PUBLIC_WS_URL || ""} ${process.env.CONTROL_PLANE_URL || ""}`,
             "frame-ancestors 'none'",
             "base-uri 'self'",


packages/web/src/middleware.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * All other routes require a valid NextAuth session.
+ */
+export async function middleware(request: NextRequest) {
+  const { pathname } = request.nextUrl;
+
+  // Allow public routes
+  if (
+    pathname.startsWith("/access-denied") ||
+    pathname.startsWith("/api/auth") ||
```
ColeMurray, line 22: [Automated Review] `pathname.startsWith("/api/auth")` and `pathname.startsWith("/access-denied")` are prefix checks, so future routes like `/api/authz` or `/access-denied-preview` would become unintentionally public. For auth boundaries this should be segment-aware and fail-closed, e.g. `pathname === "/api/auth" || pathname.startsWith("/api/auth/")` (and similarly for `/access-denied`), or moved into explicit matcher exclusions.

packages/web/src/lib/auth.ts
```diff
@@ -2,6 +2,14 @@ import type { NextAuthOptions } from "next-auth";
 import GitHubProvider from "next-auth/providers/github";
 import { checkAccessAllowed, parseAllowlist } from "./access-control";
 
+// Validate critical secrets at module load time to fail fast
+if (!process.env.NEXTAUTH_SECRET) {
```
ColeMurray, line 6: [Automated Review] Requiring `NEXTAUTH_SECRET` is the right security posture for production, but enforcing it with a top-level throw here is too early: this runs at module evaluation time (including `next build`), and many deployments inject secrets only at runtime. That turns a runtime config check into a build-time failure. Recommendation: remove the module-load throw, rely on Auth.js/NextAuth runtime enforcement in production, and enforce the secret via deployment config/health checks. That preserves security without introducing build regressions.

packages/web/src/middleware.ts
```diff
@@ -0,0 +1,64 @@
...
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    // Allow the home page to be visible (it shows sign-in UI)
+    if (pathname === "/") {
+      return NextResponse.next();
+    }
+
+    const signInUrl = new URL("/", request.url);
+    return NextResponse.redirect(signInUrl);
```
ColeMurray, line 48: [Automated Review] **[Architecture]** This middleware introduces a second layer of auth checking on every API request. Every API route handler already calls `getServerSession(authOptions)` and returns 401 if unauthenticated (e.g., `sessions/route.ts:14`, `repos/route.ts:15`, `secrets/[key]/route.ts:11`, etc.).

The request flow becomes:
1. **Middleware**: `getToken()` → decodes JWT from cookie → 401 if missing
2. **Route handler**: `getServerSession()` → also validates the session → 401 if missing

This isn't harmful — defense-in-depth is a valid pattern — but it does add latency to every authenticated API request (two JWT decode operations).

A few options to consider:

- **(a) Keep both, document the intent** — Add a comment here noting this is intentional defense-in-depth, and a comment in the route handlers explaining they retain their own auth checks as a fallback. Safest option for a security PR.
- **(b) Remove per-route `getServerSession` → 401 from API routes now covered by middleware** — Reduces latency but is riskier; a middleware misconfiguration or `matcher` gap would leave routes unprotected.
- **(c) Scope middleware to page routes only** — Let middleware handle redirects for unauthenticated page navigation (its main UX win), and let API routes continue to self-protect. This is arguably the most common Next.js pattern and avoids the double-decode. The middleware would skip `/api/*` paths, and API routes keep their existing `getServerSession` checks.

I'd recommend **(a)** or **(c)**. Just want to make sure the double-check is a conscious architectural choice rather than an oversight.

packages/web/next.config.ts
```diff
@@ -2,6 +2,31 @@ import type { NextConfig } from "next";
...
+      headers: [
+        { key: "X-Frame-Options", value: "DENY" },
+        { key: "X-Content-Type-Options", value: "nosniff" },
+        { key: "Referrer-Policy", value: "strict-origin-when-cross-origin" },
+        { key: "Permissions-Policy", value: "camera=(), microphone=(), geolocation=()" },
+        {
+          key: "Content-Security-Policy",
+          value: [
+            "default-src 'self'",
+            "script-src 'self' 'unsafe-inline' 'unsafe-eval'",
```
ColeMurray, line 17: [Automated Review] **[Security — Critical]** `'unsafe-eval'` significantly weakens this CSP. It allows `eval()`, `Function()`, and similar dynamic code execution, which is a primary XSS exploitation vector.

Next.js in production does **not** require `'unsafe-eval'`. It's only needed in development mode (React Fast Refresh / hot module replacement). Including it here negates much of the protection the CSP is meant to provide.

Recommendation: Remove `'unsafe-eval'` entirely. If something breaks in production, that's a signal to investigate which specific script needs it and address it with a nonce or hash rather than blanket-allowing eval.

```suggestion
            "script-src 'self' 'unsafe-inline'",
```

Note: `'unsafe-inline'` is also not ideal long-term (it allows inline `<script>` injection), but it's necessary for Next.js's `__NEXT_DATA__` script and inline styles. A future improvement would be to use Next.js's [built-in nonce support](https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy) to eliminate `'unsafe-inline'` as well.

packages/web/next.config.ts
```diff
@@ -2,6 +2,31 @@ import type { NextConfig } from "next";
...
+            "style-src 'self' 'unsafe-inline'",
+            "img-src 'self' data: https://avatars.githubusercontent.com",
+            "font-src 'self' https://fonts.gstatic.com",
```
ColeMurray: [Automated Review] **[Security — Critical]** `https://fonts.gstatic.com` is allowed here, but the codebase does not use Google Fonts — there are zero references to `fonts.gstatic.com`, `fonts.googleapis.com`, or any Google Font imports anywhere in `packages/web/src`.

Every origin added to the CSP expands the trust boundary. Allowing an unused origin means that if an attacker can inject a `<link>` or `@font-face` rule, they can exfiltrate data to `fonts.gstatic.com` via font request URLs (a known CSP bypass technique).

```suggestion
            "font-src 'self'",
```

packages/web/next.config.ts
```diff
@@ -2,6 +2,31 @@ import type { NextConfig } from "next";
...
+            `connect-src 'self' ${process.env.NEXT_PUBLIC_WS_URL || ""} ${process.env.CONTROL_PLANE_URL || ""}`,
```
ColeMurray, line 21: [Automated Review] **[Functionality — Critical]** Two issues with this `connect-src` directive:

**1. `CONTROL_PLANE_URL` will always be empty at build time.** This env var lacks the `NEXT_PUBLIC_` prefix, so Next.js will not inline it during the build. The `next.config.ts` `headers()` function is evaluated at build time (for static headers) or at server startup, but either way `process.env.CONTROL_PLANE_URL` is a server-side secret — the resulting CSP header sent to the browser will be `connect-src 'self' wss://... ` (with the control plane URL missing). This is likely fine in practice since API routes are same-origin (`/api/*`), but the intent doesn't match the implementation.

**2. Trailing whitespace from empty fallbacks.** When either env var is unset, the `|| ""` produces `connect-src 'self'  ` with extra spaces. While browsers tolerate this, it's messy.

Recommendation: Filter out empty values and only include env vars that are actually `NEXT_PUBLIC_*` (browser-visible):

```suggestion
            `connect-src 'self' ${[process.env.NEXT_PUBLIC_WS_URL].filter(Boolean).join(" ")}`,
```

If the browser also needs to reach the control plane directly (not just via same-origin `/api/*` proxy), then the control plane URL should be exposed as a `NEXT_PUBLIC_` env var.

packages/web/src/app/api/sessions/[id]/ws-token/route.ts
```diff
@@ -35,8 +35,13 @@ export async function POST(request: NextRequest, { params }: { params: Promise<{
     const jwt = await getToken({ req: request });
     const jwtMs = Date.now() - jwtStart;
 
+    // Warn if GitHub token has expired (control plane may still proceed with app token)
+    if (jwt?.accessTokenExpiresAt && Date.now() > (jwt.accessTokenExpiresAt as number)) {
+      console.warn(`[ws-token] GitHub access token expired for session=${sessionId}`);
+    }
```
Comment (ColeMurray) [inline packages/web/src/app/api/sessions/[id]/ws-token/route.ts:41]: [Automated Review] **[Consistency]** Token expiry is handled differently between the two routes this PR touches:

- `sessions/route.ts:53` — expired token → **hard 401**, request blocked
- `ws-token/route.ts:39` — expired token → **warning log**, request continues

This should use the same approach. I'd recommend matching the `sessions/route.ts` pattern and returning a 401 here as well:

```suggestion
    // Reject if GitHub token has expired — client should re-authenticate
    if (jwt?.accessTokenExpiresAt && Date.now() > (jwt.accessTokenExpiresAt as number)) {
      console.warn(`[ws-token] GitHub access token expired for session=${sessionId}`);
      return NextResponse.json(
        { error: "GitHub token expired. Please sign out and sign in again." },
        { status: 401 }
      );
    }
```

An expired token means the user needs to re-authenticate regardless of which endpoint they're hitting. Silently continuing with a degraded token masks auth issues from the user and makes the expiry handling harder to reason about across the codebase.

packages/web/src/app/api/sessions/[id]/archive/route.ts
```diff
@@ -15,7 +15,7 @@ export async function POST(request: NextRequest, { params }: { params: Promise<{
   const userId = session.user.id || session.user.email || "anonymous";
 
   try {
-    const response = await controlPlaneFetch(`/sessions/${id}/archive`, {
+    const response = await controlPlaneFetch(`/sessions/${encodeURIComponent(id)}/archive`, {
```
Comment (ColeMurray) [inline packages/web/src/app/api/sessions/[id]/archive/route.ts:18]: [Automated Review] **[Consistency]** Noting this for all the `encodeURIComponent` additions across this PR (archive, unarchive, prompt, ws-token, secrets-editor):

**Existing pattern**: The codebase already uses `encodeURIComponent` deliberately — on user-controlled path segments like repo owner/name (`repos/[owner]/[name]/secrets/route.ts:20`) and secret keys (`secrets/[key]/route.ts:19`). Session IDs are intentionally left unencoded because they're machine-generated identifiers from the control plane (Durable Object IDs) that only contain URL-safe characters.

**Server-side session routes**: Since session IDs are server-generated and URL-safe, encoding them doesn't mitigate a practical risk. A user can put arbitrary strings in the `[id]` URL segment, but Next.js `[id]` segments match a single path segment (can't contain `/`), and an invalid session ID simply 404s on the control plane regardless of encoding.

**Client-side secrets-editor**: The `encodeURIComponent` on `owner`/`name` in `fetch()` URLs is redundant — the server-side route handler already encodes these before forwarding to the control plane. For `normalizedKey`, the value is validated against `/^[A-Za-z_][A-Za-z0-9_]*$/`, so it can only contain characters that never need encoding.

**Inconsistency**: The PR doesn't cover client-side session ID interpolations (e.g., `use-session-socket.ts:376`, `use-session-socket.ts:433`), so even within this PR the encoding isn't applied uniformly.

We appreciate the security-minded thinking, but we don't think these changes provide enough practical value to merge. The existing codebase already encodes at the boundaries where it matters (user-controlled strings forwarded to the control plane), and the additions here either target values that can't contain unsafe characters or duplicate encoding that already happens server-side.

PR #147, opened 2026-02-18
feat: Linear bot integration

bencered:
## Summary

Adds a new `packages/linear-bot` Cloudflare Worker that integrates [Linear](https://linear.app) with Open-Inspect — triggering background coding agent sessions directly from Linear issues.

### Flow
```
Linear Issue → Webhook → Linear Bot → Control Plane → Sandbox → PR
     ↑                                                            │
     └────────── Comment with PR link ←── Completion callback ────┘
```

### Features
- **Webhook signature verification** (HMAC-SHA256)
- **Configurable triggers**: label-based (`agent` label), assignee-based, or auto-create on all new issues
- **Team → repo mapping** via KV (`PUT /config/team-repos`)
- **Deduplication** of webhook events (KV with 1h TTL)
- **Completion callbacks** — posts results back as Linear comments with PR links
- **Full Terraform infrastructure** — Worker, KV namespace, service bindings, secrets

### Control Plane Changes
- **Generic callback routing**: Replaced hardcoded `notifySlackBot()` with `notifyCallbackClient()` that routes based on message `source` field (`slack`, `linear`, etc.)
- **Relaxed `callbackContext` type** from Slack-specific interface to `Record<string, unknown>` — enabling any integration to pass its own context
- **Repository query** now includes `source` column for callback routing
- **Added `LINEAR_BOT` optional service binding** to control plane env

### Architecture
Mirrors the existing `slack-bot` package for consistency:
- Same structured JSON logger pattern
- Same internal auth via `@open-inspect/shared` token generation
- Same service binding pattern to control plane
- Same Terraform module usage (`cloudflare-worker`, `cloudflare-kv`)

### New Files
| File | Purpose |
|------|---------|
| `src/index.ts` | Hono app, webhook handler, trigger logic, config endpoints |
| `src/types.ts` | TypeScript interfaces (Env, webhook payloads, configs) |
| `src/callbacks.ts` | Completion callback handler |
| `src/utils/linear-client.ts` | Linear GraphQL API client |
| `src/utils/internal.ts` | Internal auth re-export |
| `src/logger.ts` | Structured JSON logger |
| `README.md` | Setup and usage documentation |
| `INTEGRATION.md` | Control plane integration notes |

### Setup
1. Add Linear API key and webhook secret to `terraform.tfvars`
2. `terraform apply`
3. Configure webhook URL in Linear Settings → API → Webhooks
4. Set team → repo mapping: `PUT /config/team-repos`
5. Add `agent` label to a Linear issue → agent starts working

### Build & Typecheck
```
✅ esbuild: dist/index.js (76.7kb)
✅ tsc --noEmit: clean
✅ eslint + prettier: clean (via lint-staged)
```

### DIFF
--- a/packages/control-plane/src/session/pull-request-service.test.ts
+++ b/packages/control-plane/src/session/pull-request-service.test.ts
@@ -186,15 +186,18 @@ describe("SessionPullRequestService", () => {
     });
   });
 
-  it("returns manual fallback when prompting auth is unavailable", async () => {
+  it("creates PR with app auth when prompting auth is unavailable", async () => {
     const result = await harness.service.createPullRequest(createInput({ promptingAuth: null }));
 
-    expect(result.kind).toBe("manual");
-    if (result.kind === "manual") {
-      expect(result.createPrUrl).toContain("/pull/new/");
-      expect(result.headBranch).toBe("open-inspect/session-name-1");
-      expect(result.baseBranch).toBe("main");
-    }
+    expect(result).toEqual({
+      kind: "created",
+      prNumber: 42,
+      prUrl: "https://github.com/acme/web/pull/42",
+      state: "open",
+    });
+    const createPrCall = (harness.provider.createPullRequest as ReturnType<typeof vi.fn>).mock
+      .calls[0];
+    expect(createPrCall[0]).toEqual({ authType: "app", token: "app-token" });
     expect(harness.deps.broadcastArtifactCreated).toHaveBeenCalledTimes(1);
   });
 
@@ -224,7 +227,7 @@ describe("SessionPullRequestService", () => {
     });
   });
 
-  it("reuses existing manual artifact URL for same branch", async () => {
+  it("ignores prior manual branch artifact and creates PR", async () => {
     harness.artifacts.push({
       id: "branch-artifact-1",
       type: "branch",
@@ -240,11 +243,11 @@ describe("SessionPullRequestService", () => {
     const result = await harness.service.createPullRequest(createInput({ promptingAuth: null }));
 
     expect(result).toEqual({
-      kind: "manual",
-      createPrUrl: "https://existing.example.com/manual-pr",
-      headBranch: "open-inspect/session-name-1",
-      baseBranch: "main",
+      kind: "created",
+      prNumber: 42,
+      prUrl: "https://github.com/acme/web/pull/42",
+      state: "open",
     });
-    expect(harness.deps.broadcastArtifactCreated).not.toHaveBeenCalled();
+    expect(harness.provider.createPullRequest).toHaveBeenCalledTimes(1);
   });
 });
--- a/packages/control-plane/src/session/repository.test.ts
+++ b/packages/control-plane/src/session/repository.test.ts
@@ -868,16 +868,17 @@ describe("SessionRepository", () => {
 
   describe("getMessageCallbackContext", () => {
     it("returns null for unknown message", () => {
-      mock.setData(`SELECT callback_context FROM messages WHERE id = ?`, []);
+      mock.setData(`SELECT callback_context, source FROM messages WHERE id = ?`, []);
       expect(repo.getMessageCallbackContext("unknown")).toBeNull();
     });
 
     it("returns callback context", () => {
-      mock.setData(`SELECT callback_context FROM messages WHERE id = ?`, [
-        { callback_context: '{"channel":"C123"}' },
+      mock.setData(`SELECT callback_context, source FROM messages WHERE id = ?`, [
+        { callback_context: '{"channel":"C123"}', source: "slack" },
       ]);
       expect(repo.getMessageCallbackContext("msg-1")).toEqual({
         callback_context: '{"channel":"C123"}',
+        source: "slack",
       });
     });
   });
--- a/packages/control-plane/test/integration/create-pr.test.ts
+++ b/packages/control-plane/test/integration/create-pr.test.ts
@@ -137,7 +137,7 @@ describe("POST /internal/create-pr", () => {
     );
   });
 
-  it("returns manual fallback and stores branch artifact when prompting user has no OAuth token", async () => {
+  it("creates PR with app auth when prompting user has no OAuth token", async () => {
     const { stub } = await initSession({ userId: "user-1" });
 
     const participants = await queryDO<{ id: string }>(
@@ -172,9 +172,14 @@ describe("POST /internal/create-pr", () => {
           isPrivate: true,
           providerRepoId: 12345,
         }),
-        createPullRequest: async () => {
-          throw new Error("createPullRequest should not be called for manual fallback");
-        },
+        createPullRequest: async () => ({
+          id: 42,
+          webUrl: "https://github.com/acme/web-app/pull/42",
+          apiUrl: "https://api.github.com/repos/acme/web-app/pulls/42",
+          state: "open" as const,
+          sourceBranch: "open-inspect/test-session",
+          targetBranch: "main",
+        }),
         buildManualPullRequestUrl: (config: {
           owner: string;
           name: string;
@@ -207,22 +212,20 @@ describe("POST /internal/create-pr", () => {
 
     expect(res.status).toBe(200);
     const body = await res.json<{
-      status: string;
-      createPrUrl: string;
-      headBranch: string;
-      baseBranch: string;
+      prNumber: number;
+      prUrl: string;
+      state: string;
     }>();
-    expect(body.status).toBe("manual");
-    expect(body.createPrUrl).toContain("/pull/new/");
-    expect(body.headBranch.length).toBeGreaterThan(0);
-    expect(body.baseBranch).toBe("main");
+    expect(body.prNumber).toBe(42);
+    expect(body.prUrl).toBe("https://github.com/acme/web-app/pull/42");
+    expect(body.state).toBe("open");
 
     const artifacts = await queryDO<{ type: string; metadata: string | null }>(
       stub,
       "SELECT type, metadata FROM artifacts ORDER BY created_at DESC LIMIT 1"
     );
-    expect(artifacts[0]?.type).toBe("branch");
-    expect(artifacts[0]?.metadata).toContain('"mode":"manual_pr"');
+    expect(artifacts[0]?.type).toBe("pr");
+    expect(artifacts[0]?.metadata).toContain('"number":42');
   });
 
   it("returns 409 when a PR artifact already exists", async () => {


packages/linear-bot/src/types.ts
```diff
@@ -0,0 +1,255 @@
...
+  LINEAR_CLIENT_ID: string;
+  LINEAR_CLIENT_SECRET: string;
+
+  // Worker public URL (for OAuth callback)
+  WORKER_URL: string;
+
+  // Secrets
+  LINEAR_WEBHOOK_SECRET: string;
+  LINEAR_API_KEY: string; // kept for backward compat / fallback
+  ANTHROPIC_API_KEY: string;
```
ColeMurray, line 31: [Automated Review] **Critical — Missing Terraform secret will cause runtime failure.**

`ANTHROPIC_API_KEY` is declared here as a required (`string`) binding, and is used by the classifier at `classifier/index.ts:187`:

```ts
const result = await callAnthropic(env.ANTHROPIC_API_KEY, prompt);
```

However, this secret is **absent from the `linear_bot_worker` secrets block** in `terraform/environments/production/main.tf` (lines 281–285). Compare with the `slack_bot_worker` module which correctly includes:

```hcl
{ name = "ANTHROPIC_API_KEY", value = var.anthropic_api_key },
```

Any issue that reaches the LLM classification fallback (step 4 in the repo resolution cascade) will fail with an undefined API key. Fix: add `{ name = "ANTHROPIC_API_KEY", value = var.anthropic_api_key }` to the `linear_bot_worker` secrets list in `main.tf`.

packages/linear-bot/src/index.ts
```diff
@@ -0,0 +1,825 @@
...
+  log.debug("webhook.skipped", { trace_id: traceId, type: eventType, action });
+  return c.json({ ok: true, skipped: true, reason: `unhandled event type: ${eventType}` });
+});
+
+// ─── Config Auth Middleware ───────────────────────────────────────────────────
+
+import { verifyInternalToken } from "@open-inspect/shared";
+
+app.use("/config/*", async (c, next) => {
+  if (c.req.method === "GET") return next(); // reads are open
```
ColeMurray: [Automated Review] **Security — Config GET endpoints are unauthenticated.**

The middleware skips auth for all GET requests:

```ts
if (c.req.method === "GET") return next(); // reads are open
```

This exposes the following to anyone who can reach the worker URL (no auth required):
- `GET /config/team-repos` — team → repo mappings
- `GET /config/project-repos` — project → repo mappings
- `GET /config/triggers` — trigger configuration
- `GET /config/user-prefs/:userId` — per-user model preferences (enumerable if user IDs are guessable)

While this may be acceptable for a single-tenant deployment behind a private network, Cloudflare Workers are publicly routable by default. Consider requiring auth on GET requests as well, or at minimum documenting this as an intentional design decision. The write path already validates `INTERNAL_CALLBACK_SECRET` — applying the same check to reads would be a one-line change (remove the early return).

packages/linear-bot/src/callbacks.ts
```diff
@@ -0,0 +1,198 @@
...
+        org_id: context.organizationId,
+      });
+    }
+
+    // Fallback: post a comment
+    const commentBody = payload.success
+      ? `## 🤖 Open-Inspect completed\n\n${message}`
+      : `## ⚠️ Open-Inspect encountered an issue\n\n${message}`;
+
+    const result = await postIssueComment(env.LINEAR_API_KEY, context.issueId, commentBody);
```
ColeMurray, line 289: [Automated Review] **Critical — `LINEAR_API_KEY` missing from Terraform secrets; fallback path will fail.**

`env.LINEAR_API_KEY` is used here as the comment-posting fallback when the OAuth token path fails (no `agentSessionId`, no OAuth token, etc.):

```ts
const result = await postIssueComment(env.LINEAR_API_KEY, context.issueId, commentBody);
```

However, `LINEAR_API_KEY` is **not included in the `linear_bot_worker` secrets block** in `terraform/environments/production/main.tf` (lines 281–285), nor is there a corresponding Terraform variable defined for it. At runtime this binding will be `undefined`, and `postIssueComment` will send an `undefined` Authorization header to the Linear API.

Two options:
1. **If this is a real fallback path**: add `LINEAR_API_KEY` to the Terraform secrets and define a variable for it.
2. **If OAuth is the only supported path**: make the type optional (`LINEAR_API_KEY?: string`) in `types.ts:30`, guard with a runtime check here before calling `postIssueComment`, and log when the fallback is skipped due to missing config.

packages/control-plane/src/session/durable-object.ts
```diff
@@ -2185,13 +2217,7 @@ export class SessionDO extends DurableObject<Env> {
...
         reasoningEffort?: string;
         attachments?: Array<{ type: string; name: string; url?: string }>;
-        callbackContext?: {
-          channel: string;
-          threadTs: string;
-          repoFullName: string;
-          model: string;
-          reactionMessageTs?: string;
-        };
+        callbackContext?: Record<string, unknown>;
```
ColeMurray, line 2284: [Automated Review] **Suggestion — Relaxing `callbackContext` to `Record<string, unknown>` removes compile-time safety.**

The previous typed interface caught structural issues at compile time (missing `channel`, wrong `threadTs` type, etc.). With `Record<string, unknown>`, any object is accepted — a typo like `isueId` instead of `issueId` in the Linear bot's `CallbackContext` would silently pass and only surface as a bug at runtime when the callback handler can't find the expected field.

Consider a discriminated union instead:

```ts
interface SlackCallbackContext {
  source: "slack";
  channel: string;
  threadTs: string;
  repoFullName: string;
  model: string;
  reactionMessageTs?: string;
}

interface LinearCallbackContext {
  source: "linear";
  issueId: string;
  issueIdentifier: string;
  issueUrl: string;
  repoFullName: string;
  model: string;
  agentSessionId?: string;
  organizationId?: string;
}

type CallbackContext = SlackCallbackContext | LinearCallbackContext;
```

This preserves the extensibility goal (new integrations add a new variant) while keeping type safety for each integration's required fields. The `source` discriminant also aligns with the routing logic in `getCallbackBinding()`.

terraform/environments/production/main.tf
```diff
@@ -12,14 +12,15 @@ locals {
   name_suffix = var.deployment_name
 
   # URLs for cross-service configuration
-  control_plane_host = "open-inspect-control-plane-${local.name_suffix}.${var.cloudflare_worker_subdomain}.workers.dev"
+  control_plane_host = "open-inspect-control-plane-${local.name_suffix}.${var.cloudflare_worker_subdomain}"
```
ColeMurray: [Automated Review] **Suggestion — This breaking change to URL construction should probably be reverted.**

This line changed from:
```hcl
control_plane_host = "...${var.cloudflare_worker_subdomain}.workers.dev"
```
to:
```hcl
control_plane_host = "...${var.cloudflare_worker_subdomain}"
```

The `.workers.dev` suffix is a fixed part of the Cloudflare Workers platform — every workers.dev URL follows `<worker-name>.<account-subdomain>.workers.dev`. It is not configurable, so absorbing it into the variable does not add any real flexibility.

The underlying goal seems to be fixing the module `worker_url` output to include the account subdomain (previously it was just `https://<name>.workers.dev`). That is a valid fix, but it can be done while keeping `.workers.dev` hardcoded as a platform constant:

```hcl
# Module output — keeps .workers.dev where it belongs
value = "https://${cloudflare_worker.this.name}.${var.worker_subdomain}.workers.dev"

# main.tf — no change needed to existing variable
control_plane_host = "...${var.cloudflare_worker_subdomain}.workers.dev"
```

As-is, this silently breaks existing deployments where `cloudflare_worker_subdomain = "myaccount"` — all worker URLs, the control plane URL, and WebSocket URL will be malformed. The variable also becomes more error-prone since someone will inevitably omit the `.workers.dev` suffix.

Is there a specific reason for including `.workers.dev` in the variable value rather than keeping it hardcoded? If not, I would recommend reverting this part of the change.

packages/linear-bot/src/index.ts
```diff
@@ -0,0 +1,825 @@
...
+    const { orgName } = await exchangeCodeForToken(c.env, code);
+    return c.html(`
+      <html>
+        <head><title>OAuth Success</title></head>
+        <body>
+          <h1>Open-Inspect Agent Installed!</h1>
+          <p>Successfully connected to workspace: <strong>${orgName}</strong></p>
+          <p>You can now @mention or assign the agent on Linear issues.</p>
+        </body>
+      </html>
```
ColeMurray, line 256: [Automated Review] **Security — XSS in OAuth callback via unescaped `orgName`.**

`orgName` is interpolated directly into the HTML response without escaping:

```ts
<p>Successfully connected to workspace: <strong>${orgName}</strong></p>
```

`orgName` comes from the Linear API (`viewer.organization.name`). While it's a relatively trusted source, if a workspace is named something like ``, this executes in the user's browser. OAuth callback pages are particularly sensitive since they run in an authenticated context.

Fix — escape HTML entities before interpolating:

```ts
function escapeHtml(s: string): string {
  return s.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
}

// then:
<strong>${escapeHtml(orgName)}</strong>
```

Alternatively, return a JSON response and let the caller handle display, which avoids HTML rendering entirely.

packages/linear-bot/src/index.ts
```diff
@@ -0,0 +1,825 @@
...
+    return c.json({ ok: true });
+  }
+
+  log.debug("webhook.skipped", { trace_id: traceId, type: eventType, action });
+  return c.json({ ok: true, skipped: true, reason: `unhandled event type: ${eventType}` });
+});
+
+// ─── Config Auth Middleware ───────────────────────────────────────────────────
+
+import { verifyInternalToken } from "@open-inspect/shared";
```
ColeMurray: [Automated Review] **Code Quality — Mid-file import.**

This `import` is placed between route definitions rather than at the top of the file with the other imports (lines 11–38). Move it up to the import block for consistency and to make all dependencies visible at a glance.

```ts
// Move to top of file with other imports
import { verifyInternalToken } from "@open-inspect/shared";
```

packages/linear-bot/src/callbacks.ts
```diff
@@ -0,0 +1,198 @@
...
+    { name: "HMAC", hash: "SHA-256" },
+    false,
+    ["sign"]
+  );
+  const signatureData = encoder.encode(JSON.stringify(data));
+  const expectedSig = await crypto.subtle.sign("HMAC", key, signatureData);
+  const expectedHex = Array.from(new Uint8Array(expectedSig))
+    .map((b) => b.toString(16).padStart(2, "0"))
+    .join("");
+  return signature === expectedHex;
```
ColeMurray: [Automated Review] **Security — Non-timing-safe signature comparison.**

This uses a plain string equality check:

```ts
return signature === expectedHex;
```

Compare with `verifyLinearWebhook` in `linear-client.ts:371-376` which correctly implements constant-time comparison:

```ts
if (signature.length !== expectedHex.length) return false;
let result = 0;
for (let i = 0; i < signature.length; i++) {
  result |= signature.charCodeAt(i) ^ expectedHex.charCodeAt(i);
}
return result === 0;
```

The callback signature guards access to the completion handler — it should use the same timing-safe pattern to prevent timing oracle attacks. Consider extracting a shared `timingSafeEqual(a, b)` helper that both verification functions can use.

packages/linear-bot/src/index.ts
```diff
@@ -0,0 +1,825 @@
...
+  return headers;
+}
+
+async function getTeamRepoMapping(env: Env): Promise<TeamRepoMapping> {
+  try {
+    const data = await env.LINEAR_KV.get("config:team-repos", "json");
+    if (data && typeof data === "object") return data as TeamRepoMapping;
+  } catch {
+    /* fall through */
+  }
```
ColeMurray, line 106: [Automated Review] Nit: Several KV helper functions (`getTeamRepoMapping`, `getProjectRepoMapping`, `getTriggerConfig`, `getUserPreferences`, `lookupIssueSession`) use empty `catch { /* fall through */ }` blocks. Consider logging at `debug` level instead — silent swallows make it harder to diagnose issues when KV reads fail unexpectedly (e.g., deserialization errors, binding misconfiguration).

```ts
} catch (e) {
  log.debug("kv.get", { key: "config:team-repos", error: e instanceof Error ? e : new Error(String(e)) });
}
```

packages/linear-bot/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@open-inspect/linear-bot",
+  "version": "0.1.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "build": "esbuild src/index.ts --bundle --format=esm --outfile=dist/index.js --platform=browser --target=es2022 --external:cloudflare:* --external:node:*",
+    "test": "vitest",
```
ColeMurray, line 8: [Automated Review] **Suggestion — No tests included despite vitest being configured.**

`"test": "vitest"` is wired up here, but the package has no test files. The classifier logic, webhook signature verification, callback handling, model extraction from labels, and static repo resolution all have multiple code paths with testable branching.

Even a small set of unit tests for the pure functions would add confidence:
- `extractModelFromLabels` — valid label, unknown label, no model labels
- `resolveStaticRepo` — label match, fallback to no-label entry, empty mapping
- `verifyLinearWebhook` — valid signature, invalid signature, missing signature
- `verifyCallbackSignature` — valid/invalid
- `isValidPayload` — complete payload, missing fields

These are all pure or near-pure functions that can be tested without mocking KV or service bindings.

PR #177, opened 2026-02-21
Add Bitbucket integration for auth, repo access, sandbox git flow, and PR creation

Abhinandan-Khurana:
**Summary**  
This PR adds end-to-end Bitbucket support to Background Agents by extending the existing source-control abstraction and session pipeline.

**What’s included**
- Added `BitbucketSourceControlProvider` and factory wiring for repo lookup, push auth/spec, and PR creation.
- Added Bitbucket OAuth flow in web auth and provider-neutral SCM token/session fields across web + control-plane.
- Added per-session `vcsProvider` persistence and provider-aware runtime resolution in Session DO.
- Added Bitbucket repo listing and repo validation using OAuth token fallback.
- Updated sandbox creation/restore and entrypoint git sync to use provider-neutral `VCS_*` credentials.
- Updated docs, env examples, and Terraform config for Bitbucket variables/secrets.
- Added security fix to avoid putting OAuth access tokens in URL query params during userinfo.

**Validation**
- Typecheck passed (`web`, `control-plane`, workspace)
- Control-plane tests passed
- Web tests passed
- Modal tests passed

### DIFF
--- a/packages/web/src/lib/auth.ts
+++ b/packages/web/src/lib/auth.ts
@@ -48,19 +48,7 @@ function BitbucketProvider(options: {
           return (await primary.json()) as Record<string, unknown>;
         }
 
-        // Some deployments are stricter about bearer header parsing; fallback to query auth.
-        const fallback = await fetch(
-          `https://api.bitbucket.org/2.0/user?access_token=${encodeURIComponent(token)}`,
-          {
-            headers: {
-              Accept: "application/json",
-            },
-          }
-        );
-        if (!fallback.ok) {
-          throw new Error(`Bitbucket userinfo failed (${primary.status}/${fallback.status})`);
-        }
-        return (await fallback.json()) as Record<string, unknown>;
+        throw new Error(`Bitbucket userinfo failed (${primary.status})`);
       },
     },
     clientId: options.clientId,


packages/web/src/components/sidebar-layout.tsx
```diff
@@ -61,13 +61,21 @@ export function SidebarLayout({ children }: SidebarLayoutProps) {
...
+            className="flex items-center gap-2 bg-primary text-primary-foreground px-6 py-3 font-medium hover:opacity-90 transition"
+          >
+            <GitHubIcon />
+            Sign in with GitHub
+          </button>
+          <button
+            onClick={() => signIn("bitbucket")}
+            className="flex items-center gap-2 bg-secondary text-secondary-foreground px-6 py-3 font-medium hover:opacity-90 transition"
+          >
+            Sign in with Bitbucket
```
ColeMurray, line 76: [Automated Review] Both sign-in buttons are rendered unconditionally, but the auth providers in `auth.ts` are conditionally registered based on env vars (`BITBUCKET_CLIENT_ID` / `BITBUCKET_CLIENT_SECRET`). Clicking "Sign in with Bitbucket" when no Bitbucket OAuth consumer is configured will cause a NextAuth runtime error.

Gate the Bitbucket button on provider availability (e.g., expose configured providers via the session or a server-side check) so it only appears when Bitbucket OAuth is actually configured.

terraform/environments/production/variables.tf
```diff
@@ -95,11 +121,6 @@ variable "enable_github_bot" {
   description = "Enable the GitHub bot worker. Requires github_webhook_secret and github_bot_username."
   type        = bool
   default     = false
-
-  validation {
-    condition     = var.enable_github_bot == false || (length(var.github_webhook_secret) > 0 && length(var.github_bot_username) > 0)
-    error_message = "When enable_github_bot is true, github_webhook_secret and github_bot_username must be non-empty."
-  }
 }
```
ColeMurray, line 124: [Automated Review] The validation blocks for `enable_github_bot` and `enable_linear_bot` (which enforced that required secrets are non-empty when the bot is enabled) have been removed. This weakens deploy-time validation and is unrelated to Bitbucket support.

Please revert these deletions or, if intentional, justify in the PR description and split into a separate commit.

packages/shared/src/types/index.ts
```diff
@@ -207,7 +208,7 @@ export interface ParticipantPresence {
 
 // Repository types for GitHub App installation
 export interface InstallationRepository {
```
ColeMurray, line 210: [Automated Review] Widening `id` from `number` to `string | number` is a breaking change to a shared interface consumed by multiple packages. Any downstream code that assumes `id` is a `number` (strict equality checks, arithmetic, database schemas expecting INTEGER) will break silently or fail at runtime for Bitbucket repos where `id` is a UUID string like `"{123}"`.

The `use-repos.ts` hook mirrors this change, so the break propagates to the web client as well.

**Recommendation:** Keep `id: number` and add a separate optional `providerRepoId?: string` field for Bitbucket's UUID identifier. This avoids forcing every consumer to handle a union type for what was previously a simple numeric ID.

packages/control-plane/src/routes/shared.ts
```diff
@@ -6,6 +6,16 @@ import type { CorrelationContext } from "../logger";
 import type { RequestMetrics } from "../db/instrumented-d1";
 import type { Env } from "../types";
 import { getGitHubAppConfig, getInstallationRepository } from "../auth/github-app";
+import { getBitbucketRepository, getBitbucketRepositoryWithOAuth } from "../auth/bitbucket";
+import { resolveScmProviderFromEnv, type SourceControlProviderName } from "../source-control";
+
+function stableRepoIdFromFullName(fullName: string): number {
+  let hash = 0;
```
ColeMurray, line 13: [Automated Review] This hash function has several problems:

1. **Overflow bug:** `(hash * 31 + charCode) | 0` forces 32-bit signed int. When `hash` equals `-2147483648` (valid 32-bit int), `Math.abs(-2147483648)` returns `2147483648` in JS, then `+ 1` gives `2147483649` — but in some edge cases `Math.abs` of the minimum 32-bit int is itself negative in engines that use 32-bit internally. This is fragile.

2. **Collision risk:** A 31-bit hash space (~2 billion values) makes collisions likely across repos. Since `repoId` is stored in D1 and used for session-to-repo association, a collision would silently associate sessions with the wrong repository.

3. **Applied to GitHub path too:** The fallback in `resolveInstalledRepo` (`typeof repo.id === "number" ? repo.id : stableRepoIdFromFullName(...)`) means this hash could be used for GitHub repos if `repo.id` comes through as a string for any reason — changing `repoId` values for existing sessions and orphaning them in the session index.

**Recommendation:** Use Bitbucket's UUID string directly as the repo identifier (see comment on `InstallationRepository.id`) rather than hashing it into a synthetic numeric ID.

packages/control-plane/src/router.ts
```diff
@@ -552,10 +542,22 @@ async function handleCreateSession(
   const repoOwner = body.repoOwner.toLowerCase();
   const repoName = body.repoName.toLowerCase();
 
-  let repoId: number;
+  const sessionScmProvider = body.vcsProvider ?? resolveDeploymentScmProvider(env);
+  const scmToken = body.scmToken ?? body.githubToken;
+
+  let repoId: number | null;
```
ColeMurray, line 548: [Automated Review] `repoId` was previously `number` (always set by `resolveInstalledRepo` for GitHub). Changing it to `number | null` means `null` can now flow into the Session DO's `upsertSession` call and be stored as `NULL` in the `repo_id` SQLite column.

This changes the data model for sessions — any queries filtering or joining on `repo_id` (e.g., the D1 session index, repo secrets lookups) will silently miss these sessions. Existing code downstream likely assumes `repo_id` is always a valid integer.

**Recommendation:** Ensure `resolveInstalledRepo` always returns a non-null `repoId` (even for Bitbucket), or add explicit null handling in all downstream consumers before this goes in.

packages/control-plane/src/router.ts
```diff
@@ -144,22 +144,7 @@ function enforceImplementedScmProvider(
...
-        trace_id: ctx.trace_id,
-      });
-      const response = error(
-        `SCM provider '${provider}' is not implemented in this deployment.`,
-        501
-      );
-      return withCorsAndTraceHeaders(response, ctx);
-    }
-
+    resolveDeploymentScmProvider(env);
```
ColeMurray, line 147: [Automated Review] The entire 501 guard for unimplemented SCM providers has been removed. Previously, setting `SCM_PROVIDER=bitbucket` without a working implementation would return a clear `501 Not Implemented` on all non-public routes. Now requests proceed and will fail with less obvious errors deeper in the stack if Bitbucket credentials are missing or misconfigured.

**Recommendation:** Replace the old guard with credential validation — e.g., if `provider === "bitbucket"` and none of `BITBUCKET_BOT_USERNAME`, `BITBUCKET_BOT_APP_PASSWORD`, `BITBUCKET_CLIENT_ID` are set, return a clear error early rather than letting requests fail in auth/repo-resolution layers with confusing messages.

packages/modal-infra/src/sandbox/bridge.py
```diff
@@ -23,9 +23,13 @@
 
 import httpx
 import websockets
-from websockets import ClientConnection, State
 from websockets.exceptions import InvalidStatus
 
+try:
+    from websockets import State
```
ColeMurray, line 29: [Automated Review] These changes are unrelated to Bitbucket and appear to be fixing a websockets library version compatibility issue. The `ClientConnection` type annotation changed to `Any` which weakens type safety.

Please split this into a separate commit/PR, and consider using a version-conditional type alias instead of `Any` to preserve type checking.

.gitignore
```diff
@@ -93,3 +93,7 @@ terraform/environments/production/plan.out
 terraform/environments/production/backend_override.tf
 terraform/environments/production/terraform.tfstate*
 terraform/environments/production/.terraform*
+terraform/environments/production/terraform.tfvars.backup-*
+
```
ColeMurray, line 97: [Automated Review] `research.md` and `plan.md` look like developer scratch files that should not be in the public repo's `.gitignore`. Consider removing these entries — they belong in a personal global gitignore (`~/.gitignore_global`) rather than the project-level ignore file.

packages/shared/src/index.ts
```diff
@@ -2,7 +2,7 @@
  * Shared types and utilities for Open-Inspect.
  */
 
-export * from "./types";
+export * from "./types/index";
```
ColeMurray, line 5: [Automated Review] This change from `"./types"` to `"./types/index"` is functionally equivalent but gratuitous — Node/TypeScript module resolution already resolves `"./types"` to `"./types/index"`. The explicit `/index` suffix could cause issues with certain bundler caching or TypeScript project reference configurations. Please revert unless there's a specific reason for the change.

packages/control-plane/src/session/participant-service.ts
```diff
@@ -152,60 +172,98 @@ export class ParticipantService {
...
+        });
+        accessToken = refreshed.access_token;
+        newRefreshToken = refreshed.refresh_token ?? null;
+        expiresInSeconds = refreshed.expires_in;
+      } else {
+        const githubClientId = this.env.GITHUB_CLIENT_ID;
+        const githubClientSecret = this.env.GITHUB_CLIENT_SECRET;
+        const refreshed = await refreshAccessToken(refreshToken, {
+          clientId: githubClientId as string,
+          clientSecret: githubClientSecret as string,
```
ColeMurray, line 235: [Automated Review] `refreshToken()` now calls `updateParticipantScmTokens()` instead of `updateParticipantTokens()`.

For existing GitHub sessions, this writes to `scm_*` columns AND conditionally to `github_*` columns (via the `CASE WHEN ? = 'github'` SQL). Existing participants have `scm_provider` defaulting to `'github'` from the migration, but `scm_refresh_token_encrypted` will be `null`. The `refreshToken()` method reads from `scm_refresh_token_encrypted ?? github_refresh_token_encrypted`, which is fine for fallback, but after the first refresh it writes to `scm_*` columns — so the `github_*` columns will be updated via the CASE statement, which is correct.

However, the old `updateParticipantTokens()` method still exists — it's not clear if any other code path still calls it. If so, those callers would write to `github_*` columns only, causing the `scm_*` and `github_*` columns to drift out of sync. Please either remove `updateParticipantTokens()` if it's dead code, or document which callers use which method and why.

Issue #207, opened 2026-02-22
Populate "Files changed" sidebar section from tool call events


ColeMurray:
## Summary

The session right sidebar has a fully-built "Files changed" section that never renders because it always receives an empty array. The component, types, and formatting utilities all exist — the only missing piece is a function that extracts changed file paths from the `tool_call` events already available on the client.

### What exists today

- **`FilesChangedSection`** component (`packages/web/src/components/sidebar/files-changed-section.tsx`) — renders file rows with name, additions count, and deletions count. Returns `null` when the list is empty.
- **`FileChange` type** (`packages/web/src/types/session.ts:27-31`) — `{ filename: string; additions: number; deletions: number }`.
- **`formatFilePath`** and **`formatDiffStat`** utilities (`packages/web/src/lib/format.ts`) — already handle display truncation and `+N`/`-N` formatting.
- **`SessionRightSidebarContent`** (`packages/web/src/components/session-right-sidebar.tsx:53-118`) — accepts an optional `filesChanged` prop (defaults to `[]`) and renders it inside a `CollapsibleSection` when non-empty. The empty-state text at line 109 even says *"Tasks and file changes will appear here as the agent works."*
- **`events` array** — the `useSessionSocket` hook returns all sandbox events to the session page, which already passes them to the sidebar (used today for task extraction).

### Why it's always empty

The session page (`packages/web/src/app/(app)/session/[id]/page.tsx`) renders `SessionRightSidebar` at three call sites (lines 631, 682, 710) but **never passes a `filesChanged` prop**. No code exists to derive file changes from events.

### Where the data lives

Every file-modifying tool call arrives as a `SandboxEvent` with:
- `type: "tool_call"`
- `tool: "Edit"` or `tool: "Write"`
- `args.filePath` (camelCase — see comment at `tool-formatters.ts:56`)
- `status: "completed"` for successful operations

For `Edit` tool calls, `args` also contains `oldString` and `newString`. For `Write` tool calls, `args` contains `content`. These can be used to estimate line-level diff stats.

Events are persisted to the DO's SQLite (with full `args`) for terminal statuses only (`completed`, `error`) via `shouldPersistToolCallEvent()` — so the data is available both during live streaming and on replay/reconnect.

### Established pattern to follow

[...]

PR #231, opened 2026-02-25
Populate 'Files changed' sidebar section from tool call events (#207)

pswitchy:
The "Files changed" section in the session right sidebar was previously non-functional because it always received an empty array. This PR implements a client-side utility to extract and compute file changes directly from [SandboxEvent](cci:2://file:///e:/background-agents/packages/web/src/components/session-right-sidebar.tsx:35:0-40:1) tool calls (`Edit` and [Write](cci:2://file:///e:/background-agents/packages/web/src/lib/tasks.ts:13:0-19:1)).

## Key Changes
- **New Utility**: Created [extractChangedFiles(events)](cci:1://file:///e:/background-agents/packages/web/src/lib/files.ts:23:0-68:1) in [packages/web/src/lib/files.ts](cci:7://file:///e:/background-agents/packages/web/src/lib/files.ts:0:0-0:0).
  - Filters for completed `Edit` and [Write](cci:2://file:///e:/background-agents/packages/web/src/lib/tasks.ts:13:0-19:1) tool calls.
  - Estimates additions/deletions based on line counts of `newString`/`oldString` (Edit) or `content` (Write).
  - Deduplicates by filename and accumulates stats across multiple operations.
  - Sorts alphabetically for stable display.
- **Sidebar Integration**: Modified [packages/web/src/components/session-right-sidebar.tsx](cci:7://file:///e:/background-agents/packages/web/src/components/session-right-sidebar.tsx:0:0-0:0).
  - Removed the unused `filesChanged` prop.
  - Sidebar now derives `filesChanged` internally using `useMemo` and the new extraction utility.
  - This automatically enables file change tracking across desktop, tablet, and mobile views.

## Verification
- **Automated Tests**: Added 15 comprehensive test cases in [packages/web/src/lib/files.test.ts](cci:7://file:///e:/background-agents/packages/web/src/lib/files.test.ts:0:0-0:0) covering path resolution, stat estimation, deduplication, and edge cases.
- **Test Results**: All 49 tests passed (including existing tests).
  ```bash
  Test Files  4 passed (4)
       Tests  49 passed (49)

### DIFF
--- a/packages/web/src/lib/files.ts
+++ b/packages/web/src/lib/files.ts
@@ -26,7 +26,10 @@ function countLines(str: unknown): number {
  * Compute accurate line-level additions/deletions between two strings
  * using the Myers diff algorithm via the `diff` package.
  */
-function computeEditStats(oldStr: unknown, newStr: unknown): { additions: number; deletions: number } {
+function computeEditStats(
+  oldStr: unknown,
+  newStr: unknown
+): { additions: number; deletions: number } {
   const old = typeof oldStr === "string" ? oldStr : "";
   const neu = typeof newStr === "string" ? newStr : "";
   if (old === "" && neu === "") return { additions: 0, deletions: 0 };
--- a/packages/web/src/lib/files.test.ts
+++ b/packages/web/src/lib/files.test.ts
@@ -136,8 +136,10 @@ describe("extractChangedFiles", () => {
       makeEvent({
         args: {
           filePath: "src/app.ts",
-          oldString: "import a from 'a';\nimport b from 'b';\nconst x = 1;\nconst y = 2;\nconst z = 3;",
-          newString: "import a from 'a';\nimport b from 'b';\nconst x = 42;\nconst y = 2;\nconst z = 3;",
+          oldString:
+            "import a from 'a';\nimport b from 'b';\nconst x = 1;\nconst y = 2;\nconst z = 3;",
+          newString:
+            "import a from 'a';\nimport b from 'b';\nconst x = 42;\nconst y = 2;\nconst z = 3;",
         },
       }),
     ];


packages/web/next-env.d.ts
```diff
@@ -1,6 +0,0 @@
-/// <reference types="next" />
```
ColeMurray, line 1: [Automated Review] **Unrelated change** — This auto-generated Next.js type declaration file is referenced in `tsconfig.json` (`"include": ["next-env.d.ts", ...]`). Next.js regenerates it on `next dev`/`next build`, so builds won't break, but deleting it from version control is unrelated to this PR and will cause the file to reappear as untracked after the next build.

This deletion should be reverted — or if intentional, it belongs in a separate commit with `next-env.d.ts` added to `.gitignore`.

packages/web/src/lib/files.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import type { FileChange } from "@/types/session";
+
+interface SandboxEvent {
+  type: string;
+  tool?: string;
+  args?: Record<string, unknown>;
+  status?: string;
+  timestamp: number;
+}
```
ColeMurray, line 14: [Automated Review] **Type safety gap** — This introduces a 5th local copy of `SandboxEvent` across the web package. Critically, the sidebar's `SandboxEvent` (`session-right-sidebar.tsx:37-42`) **lacks the `status` field** that `extractChangedFiles` depends on for filtering (`event.status !== "completed"` at line 35).

This works at runtime because the actual objects from `useSessionSocket` carry `status`, but the type contract doesn't guarantee it. If someone ever refactors to pass only the typed fields, `extractChangedFiles` would silently return an empty array.

Consider either:
- Adding `status?: string` to the sidebar's `SandboxEvent`, or
- Consolidating `SandboxEvent` into a single shared type (e.g., in `@/types/session` alongside `FileChange`)

packages/web/src/lib/files.ts
```diff
@@ -0,0 +1,69 @@
...
+    let deletions = 0;
+
+    if (normalizedTool === "edit") {
+      additions = countLines(args.newString);
+      deletions = countLines(args.oldString);
+    } else {
+      // write
+      additions = countLines(args.content);
+      deletions = 0;
+    }
```
ColeMurray, line 85: [Automated Review] **Inaccurate diff stats** — `countLines` counts *total* lines in `newString`/`oldString`, not the actual changed lines. For example, replacing `"a\nb"` → `"a\nb\nc"` reports +3/−2 when the real change is +1/−0. This inflates stats significantly for large edits where most lines are unchanged.

**Recommendation:** Use `diffLines` from the [`diff`](https://www.npmjs.com/package/diff) package (~3.2 kB gzipped, tree-shaken; built-in TypeScript types since v8). It returns `Change[]` objects with `.added`, `.removed`, and `.count` fields, giving accurate line-level addition/deletion counts:

```typescript
import { diffLines } from 'diff';

function computeEditStats(oldStr: string, newStr: string): { additions: number; deletions: number } {
  const changes = diffLines(oldStr, newStr);
  let additions = 0;
  let deletions = 0;
  for (const change of changes) {
    if (change.added) additions += change.count;
    if (change.removed) deletions += change.count;
  }
  return { additions, deletions };
}
```

For Write operations (no prior content to diff against), the current `countLines(content)` approach is fine.

packages/web/next-env.d.ts
ColeMurray: Done — restored `next-env.d.ts` from `main`. This deletion was unintentional and unrelated to the PR. Fixed in 6fcfec4.

packages/web/src/lib/files.ts
ColeMurray: Done — replaced the `countLines` heuristic with `diffLines` from the `diff` package (v8, ~3.2 kB gzipped). Edit events now use accurate line-level diffing via `computeEditStats`, with `ignoreNewlineAtEof: true` to handle tool call substrings cleanly. Write events still use `countLines` (no prior content to diff against).

Added a test case demonstrating the improvement: a 5-line edit with 1 changed line correctly reports +1/−1 instead of the old +5/−5.

Fixed in 9fbe48b.

packages/web/src/lib/files.ts
ColeMurray: Acknowledged — the type safety gap is real. The sidebar's `SandboxEvent` lacks `status`, so `extractChangedFiles` relies on runtime objects carrying it even though the type contract doesn't guarantee it.

The minimal fix is adding `status?: string` to the sidebar's `SandboxEvent`. Full consolidation of all 5 `SandboxEvent` copies into a shared type (e.g., in `@/types/session`) would be the proper long-term fix but is out of scope for this PR.

PR #235, opened 2026-02-25
feat: add branch selection to session creation

jmunckhof:
## Summary

Allows users to select a specific branch when creating a session instead of always using the repo's default branch. The full stack is updated: UI branch picker, API, control plane, sandbox cloning, sidebar display, and PR targeting.

## Changes

**Web UI**
- Branch selector dropdown on the session creation page (fetches branches via GitHub API)
- Base branch shown in sidebar session list with branch icon (hidden when `main`)
- Base branch and working branch shown separately in session detail metadata
- New `useBranches` hook and `/api/repos/[owner]/[name]/branches` API route

**Control Plane**
- New `listBranches` method on GitHub source control provider
- `branch` column added to D1 session index (migration `0011`)
- `baseBranch` field added to `SessionState`, `SessionResponse`, and shared `Session` types
- `baseBranch` exposed via WebSocket state (`getSessionState`) and REST state (`handleGetState`)
- PR creation now defaults to the session's base branch instead of the repo's GitHub default

**Data Plane (Modal)**
- `git clone` now uses `--branch` flag to clone the correct branch
- `git fetch` fetches the specific branch instead of all refs

## How it works

1. User picks a branch in the UI (defaults to repo's default branch)
2. Branch is stored as `repo_default_branch` in the DO and `branch` in D1
3. Sandbox clones that specific branch with `--branch` flag
4. When creating a PR, the base branch defaults to the user's selection
5. Sidebar shows the base branch; session detail shows both base and working branch

## Migration

- D1 migration `0011_add_branch_to_sessions.sql`: adds nullable `branch TEXT` column to `sessions` table
- Existing sessions will have `NULL` branch (no branch badge shown in sidebar)


jmunckhof:
Hey! I've been using Open-Inspect and needed the ability to work on feature branches, so I built this out

The main use case: when you have a repo with active development branches, you want the agent to clone and work on that branch, not just main. Previously, the sandbox always cloned the default branch regardless.

This touches the full stack but each layer is straightforward:

UI: branch picker dropdown (reuses the existing Combobox component)
Control plane: passes branch through to the sandbox config + stores it for PR targeting
Sandbox: git clone --branch instead of bare clone
Display: branch shown in sidebar list and session detail metadata
Happy to adjust anything based on your feedback.

CleanShot 2026-02-25 at 16 47 06@2x

packages/control-plane/src/router.ts
```diff
@@ -649,6 +653,7 @@ async function handleCreateSession(
     repoName,
     model,
     reasoningEffort,
+    branch: body.branch || defaultBranch || null,
```
ColeMurray: [Automated Review] **[Security]** The `branch` field from user input is passed through without validation. A malicious branch name could contain shell metacharacters or path traversal sequences. While `git clone --branch` is relatively safe (git validates branch names), and the Python data plane uses `create_subprocess_exec` (argv-safe, not shell), consider adding a basic branch name validation (e.g., reject strings with spaces, `..`, `~`, or null bytes) at the API boundary for defense in depth.

packages/web/src/components/session-sidebar.tsx
```diff
@@ -239,6 +246,13 @@ function SessionListItem({
         <span>{relativeTime}</span>
         <span>·</span>
         <span className="truncate">{repoInfo}</span>
+        {session.branch && session.branch !== "main" && (
```
ColeMurray: [Automated Review] **[Consistency]** The branch badge is hidden when `session.branch === "main"`, but repos with `master` as their default branch will always show the badge even when targeting the default. Consider comparing against the repo's actual default branch instead, or always showing the badge when branch is set.

packages/web/src/app/(app)/page.tsx
```diff
@@ -33,6 +41,7 @@ export default function Home() {
   const [reasoningEffort, setReasoningEffort] = useState<string | undefined>(
     getDefaultReasoningEffort(DEFAULT_MODEL)
   );
+  const [selectedBaseBranch, setSelectedBranch] = useState<string>("");
```
ColeMurray: [Automated Review] **[Naming]** `selectedBaseBranch` state paired with `setSelectedBranch` setter is inconsistent. Consider aligning: either `selectedBranch` / `setSelectedBranch` or `selectedBaseBranch` / `setSelectedBaseBranch`.

packages/control-plane/src/db/session-index.ts
```diff
@@ -17,6 +18,7 @@ interface SessionRow {
   repo_name: string;
   model: string;
   reasoning_effort: string | null;
```
ColeMurray, line 20: [Automated Review] **[Naming]** D1 column is `branch` while DO SQLite column is `base_branch` — these store the same concept with different names. Not blocking, but worth aligning in a follow-up to avoid confusion for future contributors.

packages/control-plane/src/auth/github-app.ts
```diff
@@ -560,6 +560,48 @@ export async function getInstallationRepository(
...
+  owner: string,
+  repo: string,
+  env?: InstallationTokenCacheBindings
+): Promise<{ name: string }[]> {
+  const token = await getCachedInstallationToken(config, env);
+  const branches: { name: string }[] = [];
+  let page = 1;
+
+  // Paginate through branches (100 per page, cap at 500)
+  while (branches.length < 500) {
```
ColeMurray, line 577: [Automated Review] **[Performance]** This fetches up to 500 branches across 5 paginated API calls. For repos with many branches, this could be slow on every repo selection. Consider adding caching (like the existing installation token cache pattern) or limiting the initial response. Not blocking for v1 but worth tracking.

packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -121,12 +121,15 @@ async def perform_git_sync(self) -> bool:
             clone_url = self._build_repo_url()
             image_build_mode = os.environ.get("IMAGE_BUILD_MODE") == "true"
             clone_depth = "100" if image_build_mode else "1"
+            base_branch = self.session_config.get("branch", "main")
```
ColeMurray: [Automated Review] **[DRY]** `self.session_config.get("branch", "main")` is repeated in 3 places (clone, fetch+rebase, and restore fetch+reset). Consider extracting to a property like `self.base_branch` to avoid drift if the default or key name ever changes.

ColeMurray:
I think we can probably tighten some of the optional/null around the branch usage

jmunckhof:
I’m not totally happy with the branching UI tbh. What do you think?
CleanShot 2026-02-26 at 09 37 06@2x
CleanShot 2026-02-26 at 09 36 59@2x

packages/control-plane/src/router.ts
ColeMurray: Done — added `if (body.branch && \!/^[\w.\-/]+$/.test(body.branch))` validation at the API boundary. Rejects anything with spaces, null bytes, tildes, etc. Addressed in 4ffd8cd.

packages/web/src/components/session-sidebar.tsx
ColeMurray: Done — removed the `\!== "main"` check so the badge always shows when `branch` is set. Since the D1 index doesn't store the repo's default branch, comparing against it isn't feasible here, and always showing the branch is clearer context. Addressed in 4ffd8cd.

packages/web/src/app/(app)/page.tsx
ColeMurray: Done — renamed to `selectedBranch`/`setSelectedBranch` throughout the file. Addressed in 4ffd8cd.

packages/control-plane/src/db/session-index.ts
ColeMurray: Agreed — worth aligning. Since the D1 migration is already shipped as `0011`, renaming would require a new migration. Deferring to a follow-up to keep this PR focused.

packages/control-plane/src/auth/github-app.ts
ColeMurray: Agreed — caching or search-based filtering would help for large repos. Deferring as a v2 improvement since it works fine for typical repo sizes.

packages/modal-infra/src/sandbox/entrypoint.py
ColeMurray: Done — extracted a `base_branch` property on `SandboxSupervisor`. All 3 call sites now use `self.base_branch`. Addressed in 4ffd8cd.

packages/web/src/components/session-sidebar.tsx
ColeMurray: On second thought, reverted this — 95% of sessions target main, so always showing the badge would just add UI noise. Keeping the original behavior: badge only shown for non-main branches. Reverted in b70f758.

packages/control-plane/src/db/session-index.ts
ColeMurray: Done — since the migration hasn't been applied yet, renamed the D1 column in-place from `branch` to `base_branch` in migration `0011`. Updated `SessionRow`, `SessionEntry`, the web `SessionItem` interface, and all mapping code to match. D1 and DO now both use `base_branch`/`baseBranch`. Addressed in 35f3082.

packages/web/src/components/session-sidebar.tsx
jmunckhof: > Consider comparing against the repo's actual default branch instead

This would make sense

ColeMurray (approved):

Issue #286, opened 2026-02-28
Code Server support


leejayhsu:
Hi @ColeMurray 

I would like to work on a PR to add first class support for [code server](https://github.com/coder/code-server). The general approach I was thinking was: 
- add code-server to preinstalled packages in `packages/modal-infra/src/images/base.py`
- start a modal tunnel in the sandbox create in `packages/modal-infra/src/sandbox/manager.py`
- password is generated at creation, and stored in `SessionState` (I'm leaning towards only creating password at sandbox creation, so that users don't get a new password each time the sandbox wakes up)
- also store the tunnel url in `SessionState`?
- webui shows url and password somewhere. probably in the right sidebar of the chat.

Is this a feature you'd like to see added?

ColeMurray:
@leejayhsu yes onboard with adding this!

Can we add a more detailed spec within the issue here before kicking off to confirm plan?

PR #301, opened 2026-03-04
feat: support code-server

leejayhsu:
this adds support for code-server. a code-server instance is served through a modal tunnel, and the credentials are stored in `SessionState`. this closes [286](https://github.com/ColeMurray/background-agents/issues/286).

Implementation details that differ from the original spec:
- password is rotated on sandbox resume. I decided to implement this way because I'm honestly not sure if env vars persist across modal sandbox sleep/resume cycles, and also it didn't feel annoying as a user to have a rotated password on each sandbox resume.

Screenshot 2026-03-04 at 8 56 04 AM

### DIFF
--- a/packages/control-plane/src/sandbox/lifecycle/manager.test.ts
+++ b/packages/control-plane/src/sandbox/lifecycle/manager.test.ts
@@ -50,6 +50,7 @@ function createMockSession(overrides: Partial<SessionRow> = {}): SessionRow {
     parent_session_id: null,
     spawn_source: "user" as const,
     spawn_depth: 0,
+    code_server_enabled: 0,
     created_at: Date.now() - 60000,
     updated_at: Date.now(),
     ...overrides,
@@ -161,6 +162,13 @@ function createMockStorage(
       if (sandbox) {
         sandbox.code_server_url = url;
         sandbox.code_server_password = password;
+      }
+    }),
+    clearSandboxCodeServer: vi.fn(() => {
+      calls.push("clearSandboxCodeServer");
+      if (sandbox) {
+        sandbox.code_server_url = null;
+        sandbox.code_server_password = null;
       }
     }),
   };
--- a/packages/control-plane/src/sandbox/lifecycle/manager.ts
+++ b/packages/control-plane/src/sandbox/lifecycle/manager.ts
@@ -82,6 +82,8 @@ export interface SandboxStorage {
   setLastSpawnError(error: string | null, timestamp: number | null): void;
   /** Update code-server URL and (encrypted) password on the sandbox row */
   updateSandboxCodeServer(url: string, password: string): void | Promise<void>;
+  /** Clear stale code-server URL and password (e.g. on sandbox teardown) */
+  clearSandboxCodeServer(): void;
 }
 
 /**
@@ -362,6 +364,7 @@ export class SandboxLifecycleManager {
         session.spawn_source === "agent" ? CHILD_SANDBOX_TIMEOUT_SECONDS : undefined;
 
       // Create sandbox via provider
+      const codeServerEnabled = session.code_server_enabled === 1;
       const createConfig: CreateSandboxConfig = {
         sessionId,
         sandboxId: expectedSandboxId,
@@ -376,6 +379,7 @@ export class SandboxLifecycleManager {
         repoImageSha,
         timeoutSeconds,
         branch: session.base_branch,
+        codeServerEnabled,
       };
 
       const result = await this.provider.createSandbox(createConfig);
@@ -392,11 +396,8 @@ export class SandboxLifecycleManager {
       }
 
       // Store code-server details and push to connected clients
-      if (result.codeServerUrl) {
-        await this.storeAndBroadcastCodeServer(
-          result.codeServerUrl,
-          result.codeServerPassword ?? ""
-        );
+      if (result.codeServerUrl && result.codeServerPassword) {
+        await this.storeAndBroadcastCodeServer(result.codeServerUrl, result.codeServerPassword);
       }
 
       this.storage.updateSandboxStatus("connecting");
@@ -488,6 +489,7 @@ export class SandboxLifecycleManager {
       const timeoutSeconds =
         session.spawn_source === "agent" ? CHILD_SANDBOX_TIMEOUT_SECONDS : undefined;
 
+      const codeServerEnabled = session.code_server_enabled === 1;
       const result = await this.provider.restoreFromSnapshot({
         snapshotImageId,
         sessionId: session.session_name || session.id,
@@ -501,6 +503,7 @@ export class SandboxLifecycleManager {
         userEnvVars,
         timeoutSeconds,
         branch: session.base_branch,
+        codeServerEnabled,
       });
 
       if (result.success) {
@@ -516,11 +519,8 @@ export class SandboxLifecycleManager {
         }
 
         // Store code-server details and push to connected clients
-        if (result.codeServerUrl) {
-          await this.storeAndBroadcastCodeServer(
-            result.codeServerUrl,
-            result.codeServerPassword ?? ""
-          );
+        if (result.codeServerUrl && result.codeServerPassword) {
+          await this.storeAndBroadcastCodeServer(result.codeServerUrl, result.codeServerPassword);
         }
 
         this.storage.updateSandboxStatus("connecting");
@@ -682,6 +682,7 @@ export class SandboxLifecycleManager {
         this.log.error("Heartbeat snapshot failed", { error: e instanceof Error ? e : String(e) })
       );
       this.storage.updateSandboxStatus("stale");
+      this.storage.clearSandboxCodeServer();
       this.broadcaster.broadcast({ type: "sandbox_status", status: "stale" });
 
       // Best-effort shutdown: tell sandbox to exit cleanly (connection may already be dead).
@@ -717,6 +718,7 @@ export class SandboxLifecycleManager {
         await this.callbacks.onSandboxTerminating?.();
         // Set status to stopped FIRST to block reconnection attempts
         this.storage.updateSandboxStatus("stopped");
+        this.storage.clearSandboxCodeServer();
         this.broadcaster.broadcast({ type: "sandbox_status", status: "stopped" });
 
         // Take snapshot
--- a/packages/control-plane/src/session/schema.ts
+++ b/packages/control-plane/src/session/schema.ts
@@ -25,6 +25,7 @@ CREATE TABLE IF NOT EXISTS session (
   parent_session_id TEXT,                           -- Parent session ID (NULL for top-level)
   spawn_source TEXT NOT NULL DEFAULT 'user',        -- 'user' or 'agent'
   spawn_depth INTEGER NOT NULL DEFAULT 0,           -- 0 for top-level, parent.depth + 1 for children
+  code_server_enabled INTEGER NOT NULL DEFAULT 0,   -- 0 = disabled, 1 = enabled (opt-in)
   created_at INTEGER NOT NULL,
   updated_at INTEGER NOT NULL
 );
@@ -350,6 +351,11 @@ export const MIGRATIONS: readonly SchemaMigration[] = [
       ALTER TABLE sandbox ADD COLUMN code_server_url TEXT;
       ALTER TABLE sandbox ADD COLUMN code_server_password TEXT;
     `,
+  },
+  {
+    id: 27,
+    description: "Add code_server_enabled to session",
+    run: `ALTER TABLE session ADD COLUMN code_server_enabled INTEGER NOT NULL DEFAULT 0`,
   },
 ];
 
--- a/packages/modal-infra/src/sandbox/entrypoint.py
+++ b/packages/modal-infra/src/sandbox/entrypoint.py
@@ -40,7 +40,6 @@ class SandboxSupervisor:
 
     # Configuration
     OPENCODE_PORT = 4096
-    CODE_SERVER_PORT = CODE_SERVER_PORT
     HEALTH_CHECK_TIMEOUT = 30.0
     MAX_RESTARTS = 5
     BACKOFF_BASE = 2.0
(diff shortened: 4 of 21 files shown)


packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -565,6 +608,20 @@ async def monitor_processes(self) -> None:
...
+                self.log.warn(
+                    "code_server.crash",
+                    exit_code=self.code_server_process.returncode,
+                    restart_count=code_server_restart_count,
+                )
+
+                if code_server_restart_count <= self.MAX_RESTARTS:
+                    delay = min(self.BACKOFF_BASE**code_server_restart_count, self.BACKOFF_MAX)
+                    await asyncio.sleep(delay)
+                    await self.start_code_server()
```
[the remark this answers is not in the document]
leejayhsu: resolved in https://github.com/ColeMurray/background-agents/pull/301/commits/3177a9ceb4ee1665ff499bf19a5131687944627f

packages/modal-infra/src/images/base.py
```diff
@@ -104,6 +104,11 @@
         # This ensures tools can import the plugin without needing to run bun add
         "npm install -g @opencode-ai/plugin@latest zod",
     )
+    # Install code-server for browser-based VS Code editing
+    .run_commands(
+        "curl -fsSL https://code-server.dev/install.sh | sh",
+        "code-server --version",
```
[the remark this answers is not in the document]
ColeMurray: @open-inspect[bot] can you lookup what is the latest version for this and pin it

packages/modal-infra/src/sandbox/manager.py
```diff
@@ -25,6 +27,7 @@
 log = get_logger("manager")
 
 DEFAULT_SANDBOX_TIMEOUT_SECONDS = 7200  # 2 hours
+CODE_SERVER_PORT = 8080
```
ColeMurray: [Automated Review] **[Code Quality]** `CODE_SERVER_PORT = 8080` is duplicated here and in `entrypoint.py` (`SandboxSupervisor.CODE_SERVER_PORT = 8080`). These must stay in sync. Consider extracting to a shared constants module so both files reference the same definition.

ColeMurray (commented):
[Automated Review] See inline comments.

packages/modal-infra/src/sandbox/manager.py
```diff
@@ -83,6 +88,31 @@ def _get_repo_key(self, repo_owner: str, repo_name: str) -> str:
...
+        env_vars["CODE_SERVER_PASSWORD"] = password
+        return password
+
+    @staticmethod
+    async def _resolve_code_server_tunnel(sandbox: modal.Sandbox, sandbox_id: str) -> str | None:
+        """Resolve the code-server tunnel URL from Modal, returning None on failure."""
+        try:
+            loop = asyncio.get_event_loop()
+            tunnels = await loop.run_in_executor(None, sandbox.tunnels)
+            tunnel = tunnels[CODE_SERVER_PORT]
```
ColeMurray: t

packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -313,6 +315,46 @@ def _setup_openai_oauth(self) -> None:
...
+        """Forward code-server stdout to supervisor stdout."""
+        if not self.code_server_process or not self.code_server_process.stdout:
+            return
+
+        try:
+            async for line in self.code_server_process.stdout:
+                print(f"[code-server] {line.decode().rstrip()}")
+        except Exception as e:
+            print(f"[supervisor] code-server log forwarding error: {e}")
+
```
ColeMurray, line 369: [Automated Review] Nit: This uses bare `print()` while the rest of the supervisor uses `self.log.*`. Use the structured logger for consistency so these lines appear in the same format as other sandbox logs.

packages/control-plane/src/sandbox/lifecycle/manager.test.ts
```diff
@@ -154,6 +156,13 @@ function createMockStorage(
         sandbox.last_spawn_error_at = timestamp;
       }
     }),
+    updateSandboxCodeServer: vi.fn((url: string, password: string) => {
+      calls.push(`updateSandboxCodeServer:${url}`);
+      if (sandbox) {
+        sandbox.code_server_url = url;
```
ColeMurray, line 163: [Automated Review] **[Testing]** The `updateSandboxCodeServer` mock is added but never exercised by any test. Please add at least one test case that asserts `doSpawn` (or `restoreFromSnapshot`) calls `storage.updateSandboxCodeServer` with the correct URL and password.

packages/control-plane/src/session/schema.ts
```diff
@@ -339,6 +341,14 @@ export const MIGRATIONS: readonly SchemaMigration[] = [
       ALTER TABLE session ADD COLUMN spawn_depth INTEGER NOT NULL DEFAULT 0;
     `,
   },
+  {
+    id: 26,
+    description: "Add code-server fields to sandbox",
+    run: `
+      ALTER TABLE sandbox ADD COLUMN code_server_url TEXT;
+      ALTER TABLE sandbox ADD COLUMN code_server_password TEXT;
```
ColeMurray, line 352: [Automated Review] Nit: This migration runs two `ALTER TABLE` statements in a single `run` string. If the first succeeds but the second fails, the migration is left in a partial state (SQLite doesn't support transactional DDL rollback for `ALTER TABLE`). Consider documenting this risk or running them as separate migrations.

packages/control-plane/src/sandbox/lifecycle/manager.ts
```diff
@@ -801,6 +813,19 @@ export class SandboxLifecycleManager {
     return this.wsManager.getConnectedClientCount();
   }
 
+  /**
+   * Store code-server details in the database and push to connected clients.
+   * Shared by doSpawn() and restoreFromSnapshot().
+   */
+  private storeAndBroadcastCodeServer(url: string, password: string): void {
```
ColeMurray: [Automated Review] **[Observability]** `storeAndBroadcastCodeServer` has no logging. If a user reports "editor not showing up," there's no trail to debug whether the info was received and broadcast. Consider adding a log line here.

terraform/d1/migrations/0014_add_code_server_to_sandbox.sql
```diff
@@ -0,0 +1,4 @@
+-- Code-server tunnel URL and password on the sandbox table.
+-- Both URL and password rotate on every wake/restore.
+ALTER TABLE sandbox ADD COLUMN code_server_url TEXT;
+ALTER TABLE sandbox ADD COLUMN code_server_password TEXT;
```
ColeMurray: [Automated Review] **[Security — Critical]** The `code_server_password` is stored as plaintext `TEXT` in D1. The codebase already encrypts sensitive data at rest — repo secrets use AES-256-GCM via `REPO_SECRETS_ENCRYPTION_KEY`. Storing this password unencrypted is a regression from the project's existing security posture.

The same applies to DO SQLite storage (`repository.ts`).

Please encrypt the password before persisting and decrypt on read, consistent with the repo secrets pattern.

packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -565,6 +608,25 @@ async def monitor_processes(self) -> None:
...
+                code_server_restart_count += 1
+                self.log.warn(
+                    "code_server.crash",
+                    exit_code=self.code_server_process.returncode,
+                    restart_count=code_server_restart_count,
+                )
+
+                if code_server_restart_count <= self.MAX_RESTARTS:
+                    delay = min(self.BACKOFF_BASE**code_server_restart_count, self.BACKOFF_MAX)
+                    await asyncio.sleep(delay)
```
ColeMurray, line 634: [Automated Review] **[Correctness — Critical]** If `start_code_server()` raises here (e.g., `create_subprocess_exec` fails), the exception is unhandled and will crash the entire `monitor_processes` loop — which is fatal to the sandbox.

The existing OpenCode and bridge restart paths in this method wrap their restart calls in try/except. This code-server restart should do the same:

```python
try:
    await self.start_code_server()
except Exception as e:
    self.log.warn("code_server.restart_error", exc=e)
    self.code_server_process = None
```

packages/modal-infra/src/images/base.py
```diff
@@ -104,6 +107,11 @@
         # This ensures tools can import the plugin without needing to run bun add
         "npm install -g @opencode-ai/plugin@latest zod",
     )
+    # Install code-server for browser-based VS Code editing
+    .run_commands(
+        f"curl -fsSL https://code-server.dev/install.sh | sh -s -- --version {CODE_SERVER_VERSION}",
```
ColeMurray: [Automated Review] **[Security — Critical]** `curl | sh` downloads and executes a remote script at image build time. While the version is now pinned (good), the install script itself is not integrity-verified. A compromised CDN or MITM during build would inject arbitrary code into every sandbox.

Consider downloading the release tarball directly from the GitHub releases page with SHA256 checksum verification:
```python
# Example: download binary directly with checksum
f"curl -fsSL https://github.com/coder/code-server/releases/download/v{CODE_SERVER_VERSION}/code-server_{CODE_SERVER_VERSION}_amd64.deb -o /tmp/code-server.deb",
"echo '<expected_sha256>  /tmp/code-server.deb' | sha256sum -c",
"dpkg -i /tmp/code-server.deb && rm /tmp/code-server.deb",
```

packages/modal-infra/src/sandbox/manager.py
```diff
@@ -83,6 +88,31 @@ def _get_repo_key(self, repo_owner: str, repo_name: str) -> str:
...
+    @staticmethod
+    def _generate_code_server_credentials(
+        env_vars: dict[str, str],
+    ) -> str:
+        """Generate a code-server password and inject it into env_vars.
+
+        Returns the generated password.
+        """
+        password = secrets.token_urlsafe(16)
+        env_vars["CODE_SERVER_PASSWORD"] = password
```
ColeMurray: [Automated Review] **[Design]** This static method mutates its `env_vars` argument as a side effect *and* returns a value. A reader doesn't expect `_generate_code_server_credentials` to silently inject into an unrelated dict.

Suggestion: either make the caller responsible for the env var assignment, or return a tuple. Simplest fix:
```python
@staticmethod
def _generate_code_server_password() -> str:
    return secrets.token_urlsafe(16)
```
Then at the call site: `password = self._generate_code_server_password(); env_vars["CODE_SERVER_PASSWORD"] = password`

packages/modal-infra/src/sandbox/manager.py
```diff
@@ -83,6 +88,31 @@ def _get_repo_key(self, repo_owner: str, repo_name: str) -> str:
...
+        return password
+
+    @staticmethod
+    async def _resolve_code_server_tunnel(sandbox: modal.Sandbox, sandbox_id: str) -> str | None:
+        """Resolve the code-server tunnel URL from Modal, returning None on failure."""
+        try:
+            loop = asyncio.get_event_loop()
+            tunnels = await loop.run_in_executor(None, sandbox.tunnels)
```
ColeMurray: [Automated Review] **[Deprecation]** `asyncio.get_event_loop()` is deprecated since Python 3.10 and emits a `DeprecationWarning` in 3.12+ when no running loop exists. Since this is already inside an async function, use:
```python
loop = asyncio.get_running_loop()
```

packages/modal-infra/src/sandbox/manager.py
ColeMurray: [Automated Review] **[Robustness]** `tunnels[CODE_SERVER_PORT]` will `KeyError` if the tunnel isn't ready yet (race with sandbox startup). The bare `except Exception` catches this, but silently returning `None` means the entire code-server feature is invisible to the user with no indication of why.

Consider logging the specific exception type, and potentially retrying with a short backoff since tunnel resolution can race sandbox startup.

terraform/d1/migrations/0014_add_code_server_to_sandbox.sql
ColeMurray: Addressed in 34f61f0. The code-server password is now encrypted at rest using AES-256-GCM via `REPO_SECRETS_ENCRYPTION_KEY` (the same key used for repo secrets) before being stored in the DO's SQLite database.

- **Write path**: The `updateSandboxCodeServer` adapter in the DO encrypts via `encryptToken()` before calling the repository
- **Read path**: `getSessionState()` decrypts via `decryptToken()` with a try/catch fallback for pre-encryption data (backward compat)
- **WebSocket**: The plaintext password is still broadcast over the already-authenticated WSS connection (unchanged)

Note: The D1 migration columns (`0014_add_code_server_to_sandbox.sql`) are not written to by any code in this PR — only the DO SQLite is used for storage. If these D1 columns are intended for future use, they should also use encrypted values when that code is added.

packages/modal-infra/src/sandbox/entrypoint.py
ColeMurray: Addressed in 34f61f0. The `await self.start_code_server()` call in the restart path is now wrapped in `try/except`:

```python
try:
    await self.start_code_server()
except Exception as e:
    self.log.warn("code_server.restart_failed", exc=e)
    self.code_server_process = None
```

On failure, it logs the error and sets `code_server_process = None` to stop further restart attempts — the supervisor loop continues running normally. Added 3 tests covering this: crash doesn't trigger shutdown, restart exception is caught, and max restarts gives up gracefully.

packages/control-plane/src/sandbox/lifecycle/manager.test.ts
ColeMurray: Addressed in 34f61f0. Added `tests/test_code_server.py` with 12 tests covering:

- **`_generate_code_server_credentials`** (4 tests): env var injection, return value, uniqueness, no-clobber
- **`_resolve_code_server_tunnel`** (3 tests): success, exception, missing port
- **`create_sandbox` / `restore_from_snapshot`** (2 tests): assert `code_server_url`, `code_server_password`, and `encrypted_ports` on the returned handle
- **`monitor_processes`** (3 tests): crash is non-fatal, restart exception is caught, max restarts gives up

The `updateSandboxCodeServer` mock is also updated to `async` to match the new interface signature (now returns `Promise<void>` to support encryption).

PR #303, opened 2026-03-04
fix: fetch github email using api to resolve attribution for github private emails

leejayhsu:
This fixes a bug where, if a user's GitHub email is set to private, we are unable to get the user's email. This results in a commit where the name is correct but the email is not (it would fall back on the default open inspect email).

This causes issues with git attribution, which were originally observed because Vercel preview builds would not work, because Vercel would not recognize that the PR author is part of the Vercel team.

### DIFF
--- a/packages/web/src/lib/auth.ts
+++ b/packages/web/src/lib/auth.ts
@@ -72,16 +72,23 @@ export const authOptions: NextAuthOptions = {
     }),
   ],
   callbacks: {
-    async signIn({ profile, user }) {
+    async signIn({ profile, user, account }) {
       const config = {
         allowedDomains: parseAllowlist(process.env.ALLOWED_EMAIL_DOMAINS),
         allowedUsers: parseAllowlist(process.env.ALLOWED_USERS),
       };
 
       const githubProfile = profile as { login?: string };
+
+      // Resolve the real email even when the GitHub profile email is private.
+      let email = user.email ?? undefined;
+      if (!email && account?.access_token) {
+        email = (await fetchGitHubPrimaryEmail(account.access_token)) ?? undefined;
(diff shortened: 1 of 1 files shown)


leejayhsu:
> Additional Comments (1)
> **`packages/web/src/lib/auth.ts`, line 91** The `signIn` callback uses `user.email` which is `null` for private-email users. When `ALLOWED_EMAIL_DOMAINS` is configured, `checkAccessAllowed` receives `email: undefined`, skips the domain check (see access-control.ts line 50), and returns `false` if the user is not explicitly in `ALLOWED_USERS`. This means private-email users can be incorrectly denied sign-in — even though their real email is now fetchable via `fetchGitHubPrimaryEmail`.
> 
> To fix: destructure `account` and call the helper here too:
> 
> ```ts
> async signIn({ profile, user, account }) {
>   const config = {
>     allowedDomains: parseAllowlist(process.env.ALLOWED_EMAIL_DOMAINS),
>     allowedUsers: parseAllowlist(process.env.ALLOWED_USERS),
>   };
> 
>   const githubProfile = profile as { login?: string };
>   
>   // Resolve the real email even when the GitHub profile email is private.
>   let email = user.email ?? undefined;
>   if (!email && account?.access_token) {
>     email = (await fetchGitHubPrimaryEmail(account.access_token)) ?? undefined;
>   }
> 
>   const isAllowed = checkAccessAllowed(config, {
>     githubUsername: githubProfile.login,
>     email,
>   });
>   ...
> }
> ```

addressed in https://github.com/ColeMurray/background-agents/pull/303/commits/e25b3bc8284b3f6d6b6337e36ad99d5165e49d3a

PR #334, opened 2026-03-12
Slack branch config

ffesseler:
This change adds branch preferences to the Slack bot settings. 
Users can now set a default branch for all new Slack sessions, and also define repository-specific branch overrides. When a session is started from Slack, the  bot now picks the branch in this order: repo-specific override, then global user override, then the repository’s default branch. 
The Slack App Home was updated so users can set/edit/clear the global branch, manage per-repo  overrides (including delete), and search repositories when choosing overrides. Invalid branch names are rejected with immediate inline feedback in Slack.


Screenshot : <img width="1447" height="683" alt="Screenshot 2026-03-12 at 17 19 27" src="https://github.com/user-attachments/assets/df6d2b9e-94b9-4509-97cc-694912fd56d1" />

Here is a "backward plan" I generated in case you don't accept the PR directly but still would like to implement it a bit diffrently : 
``` 
### Goal                                                                                                          
                                                                                                                   
 Add branch preference support to the Slack bot so branch selection for new sessions can be controlled by users    
 from Slack App Home.                                                                                              
                                                                                                                   
 ### Expected behavior                                                                                             
                                                                                                                   
 When starting a new session from Slack, resolve branch in this order:                                             
                                                                                                                   
 1. repo-specific branch override (if set for that user+repo)                                                      
 2. global user branch override                                                                                    
 3. repo default branch (implicit fallback)                                                                        
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────                                  
                                                                                                                   
 Technical solution (high-level)                                                                                   
                                                                                                                   
 ### 1) Extend preference model                                                                                    
                                                                                                                   
 - Add optional branch to user preferences (UserPreferences).                                                      
 - Keep existing model/reasoning behavior intact.                                                                  
                                                                                                                   
 ### 2) Add Slack modal API helper                                                                                 
                                                                                                                   
 - Add a views.open client helper (e.g. openView) in Slack client utils.                                           
 - Keep views.publish for App Home updates.                                                                        
                                                                                                                   
 ### 3) Store branch preferences in KV                                                                             
                                                                                                                   
 Use two layers of KV storage:                                                                                     
                                                                                                                   
 - Global user prefs (existing key):                                                                               
     - user_prefs:${userId}                                                                                        
     - includes model, reasoningEffort, optional branch, updatedAt                                                 
 - Per-repo override (new key family):                                                                             
     - user_repo_branch:${userId}:${repoId} -> branch string                                                       
                                                                                                                   
 Also add helper to list all repo overrides by prefix:                                                             
 - user_repo_branch:${userId}:                                                                                     
                                                                                                                   
 Clearing override = delete KV key.                                                                                
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────                                  
                                                                                                                   
 Validation requirements                                                                                           
                                                                                                                   
 Normalize branch input with trim; empty value means “unset”.                                                      
                                                                                                                   
 Reject invalid branches with one generic error message:                                                           
 - "Enter a valid Git branch name."                                                                                
                                                                                                                   
 Validation should reject at least:                                                                                
 - leading -                                                                                                       
 - leading/trailing /                                                                                              
 - trailing .                                                                                                      
 - .., //, @{                                                                                                      
 - control characters                                                                                              
 - invalid special chars / whitespace                                                                              
 - path segments that are empty, start with ., or end with .lock                                                   
                                                                                                                   
 Validate both:                                                                                                    
 1. before returning modal submission response (to show inline errors)                                             
 2. before writing to KV (defensive check)                                                                         
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────                                  
                                                                                                                   
 App Home changes                                                                                                  
                                                                                                                   
 Add a “Branch” settings area in App Home:                                                                         
                                                                                                                   
 1. Global branch section                                                                                          
     - button to open global branch modal (Set/Edit)                                                               
     - show current global branch or “repo default”                                                                
     - show “Clear branch override” button if currently set                                                        
 2. Per-repo branch section                                                                                        
     - repository selector to configure repo-specific override                                                     
     - use external_select + suggestion endpoint (not static list)                                                 
     - include precedence hint text (repo override -> global override -> repo default)                             
     - list configured repo overrides with delete buttons + confirm dialog                                         
 3. Keep model/reasoning controls and preserve branch when those change.                                           
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────                                  
                                                                                                                   
 Interaction routing behavior                                                                                      
                                                                                                                   
 In /interactions:                                                                                                 
                                                                                                                   
 1. Parse Slack payload from form body.                                                                            
 2. Handle block_suggestion for repo search options (for external_select).                                         
     - return filtered options                                                                                     
     - cap results to 100                                                                                          
 3. For view_submission branch modals:                                                                             
     - if invalid: return response_action: "errors" with block-level error                                         
     - if valid: return response_action: "clear" immediately                                                       
 4. Process expensive work asynchronously (waitUntil) for normal actions/submissions.                              
 5. Exception: actions that open modals should run inline (need fresh trigger_id).                                 
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────                                  
                                                                                                                   
 Modal flows                                                                                                       
                                                                                                                   
 ### Global branch modal                                                                                           
                                                                                                                   
 - Callback ID for global branch settings.                                                                         
 - Input field is optional.                                                                                        
 - Save updates UserPreferences.branch.                                                                            
 - Republish App Home after save.                                                                                  
                                                                                                                   
 ### Repo branch modal                                                                                             
                                                                                                                   
 - Callback ID for repo-specific settings.                                                                         
 - Include repoId (+ optionally userId) in private_metadata.                                                       
 - Input field optional; empty clears repo override.                                                               
 - Save/clear repo override KV key.                                                                                
 - Republish App Home after save.                                                                                  
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────                                  
                                                                                                                   
 Session creation integration                                                                                      
                                                                                                                   
 When creating a new session:                                                                                      
 - read user preferences                                                                                           
 - read repo-specific override                                                                                     
 - compute effective branch with precedence above                                                                  
 - send branch in create-session payload to control plane                                                          
                                                                                                                   
 No change to thread/session mapping logic besides passing branch during session creation.                         
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────                                  
                                                                                                                   
 Acceptance tests to include                                                                                       
                                                                                                                   
 - invalid global branch submission returns modal field errors                                                     
 - invalid repo branch submission returns modal field errors                                                       
 - valid modal submissions are acknowledged immediately (response_action: "clear")                                 
 - repo-specific branch is stored correctly from modal                                                             
 - repo-specific branch can be deleted from App Home                                                               
 - repo search suggestions work for large repo sets (>100 total repos)                                             
                                                                                                                   
 ────────────────────────────────────────────────────────────────────────────────   ````

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -592,6 +592,7 @@ Go to your fork's Settings → Secrets and variables → Actions, and add:
 | `CLOUDFLARE_API_TOKEN`        | Your Cloudflare API token                                                     |
 | `CLOUDFLARE_ACCOUNT_ID`       | Your Cloudflare account ID                                                    |
 | `CLOUDFLARE_WORKER_SUBDOMAIN` | Your workers.dev subdomain                                                    |
+| `DEPLOYMENT_NAME`             | Your deployment name                                                          |
 | `R2_ACCESS_KEY_ID`            | R2 access key ID                                                              |
 | `R2_SECRET_ACCESS_KEY`        | R2 secret access key                                                          |
 | `WEB_PLATFORM`                | `vercel` or `cloudflare`                                                      |
@@ -602,8 +603,8 @@ Go to your fork's Settings → Secrets and variables → Actions, and add:
 | `MODAL_TOKEN_ID`              | Modal token ID                                                                |
 | `MODAL_TOKEN_SECRET`          | Modal token secret                                                            |
 | `MODAL_WORKSPACE`             | Modal workspace name                                                          |
-| `GH_APP_CLIENT_ID`            | GitHub App client ID                                                          |
-| `GH_APP_CLIENT_SECRET`        | GitHub App client secret                                                      |
+| `GH_OAUTH_CLIENT_ID`          | GitHub App OAuth client ID                                                    |
+| `GH_OAUTH_CLIENT_SECRET`      | GitHub App OAuth client secret                                                |
 | `GH_APP_ID`                   | GitHub App ID                                                                 |
 | `GH_APP_PRIVATE_KEY`          | GitHub App private key (PKCS#8 format)                                        |
 | `GH_APP_INSTALLATION_ID`      | GitHub App installation ID                                                    |
@@ -638,6 +639,13 @@ Then upload all at once (run from your fork's directory, or use
 
 ```bash
 gh secret set -f .secrets
+```
+
+If you bulk upload from a file, set multiline secrets like `GH_APP_PRIVATE_KEY` separately so the
+PEM formatting is preserved:
+
+```bash
+gh secret set GH_APP_PRIVATE_KEY < private-key-pkcs8.pem
 ```
 
 Once configured, the GitHub Actions workflow will:
--- a/package-lock.json
+++ b/package-lock.json
@@ -2745,6 +2745,7 @@
       "version": "1.8.1",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.8.1.tgz",
       "integrity": "sha512-mehfKSMWjjNol8659Z8KxEMrdSJDDot5SXMq00dM8BN4o+CLNXQ0xH2V7EchNHV4RmbZLmmPdEaXZc5H2FXmDg==",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -3475,6 +3476,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3497,6 +3499,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3519,6 +3522,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3535,6 +3539,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3551,6 +3556,7 @@
       "cpu": [
         "arm"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3567,6 +3573,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3583,6 +3590,7 @@
       "cpu": [
         "ppc64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3599,6 +3607,7 @@
       "cpu": [
         "riscv64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3615,6 +3624,7 @@
       "cpu": [
         "s390x"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3631,6 +3641,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3647,6 +3658,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3663,6 +3675,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -3679,6 +3692,7 @@
       "cpu": [
         "arm"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3701,6 +3715,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3723,6 +3738,7 @@
       "cpu": [
         "ppc64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3745,6 +3761,7 @@
       "cpu": [
         "riscv64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3767,6 +3784,7 @@
       "cpu": [
         "s390x"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3789,6 +3807,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3811,6 +3830,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3833,6 +3853,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -3855,6 +3876,7 @@
       "cpu": [
         "wasm32"
       ],
+      "dev": true,
       "license": "Apache-2.0 AND LGPL-3.0-or-later AND MIT",
       "optional": true,
       "dependencies": {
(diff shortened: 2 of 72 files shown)

### CODE REVIEW

File: packages/slack-bot/src/index.ts [hunk 1]
```diff
@@ -298,6 +334,203 @@ function getUserPreferencesKey(userId: string): string {
...
+    return INVALID_BRANCH_ERROR;
+  }
+
+  if (
+    branch.startsWith("/") ||
+    branch.endsWith("/") ||
+    branch.endsWith(".") ||
+    branch.includes("..") ||
+    branch.includes("//") ||
+    branch.includes("@{") ||
```
ColeMurray: [Automated Review] **Critical — Branch name `@` is not rejected**

Git's `check-ref-format` explicitly forbids the single character `@` as a branch name. The current validation handles `@{` but `@` alone passes all checks and would be stored, only to fail at actual Git operations.

Suggested fix — add after the `startsWith("-")` check:
```typescript
if (branch === "@") {
  return INVALID_BRANCH_ERROR;
}
```

packages/slack-bot/src/index.ts
```diff
@@ -752,19 +1300,92 @@ app.post("/interactions", async (c) => {
...
     http_status: 200,
-    action_id: payload.actions?.[0]?.action_id,
+    interaction_type: payload.type,
+    action_id: actionId,
+    callback_id: payload.view?.callback_id,
     duration_ms: Date.now() - startTime,
   });
 
+  if (isViewSubmission) {
+    return c.json({ response_action: "clear" });
```
ColeMurray, line 1181: [Automated Review] **Critical — `view_submission` response not scoped to branch modals**

This returns `{ response_action: "clear" }` for *all* `view_submission` payloads, not just branch modals. Any future modal submission would get an unexpected `clear` response. Today there are no other modals so it's not broken, but it's a latent bug.

Suggested fix:
```typescript
if (isViewSubmission && isBranchModalCallbackId(payload.view?.callback_id)) {
  return c.json({ response_action: "clear" });
}
```

packages/slack-bot/src/index.ts
```diff
@@ -298,6 +334,203 @@ function getUserPreferencesKey(userId: string): string {
...
+  try {
+    const listed = await env.SLACK_KV.list({ prefix });
+
+    for (const key of listed.keys) {
+      const repoId = key.name.slice(prefix.length);
+      if (!repoId) {
+        continue;
+      }
+
+      const branch = await getUserRepoBranchPreference(env, userId, repoId);
```
ColeMurray: [Automated Review] **Suggestion — N+1 KV reads in `getUserRepoBranchPreferences`**

`list()` enumerates keys, then `getUserRepoBranchPreference()` calls `get()` again for each — resulting in `N+1` KV reads. The validation logic is valuable but could be inlined to avoid redundant lookups:

```typescript
for (const key of listed.keys) {
  const repoId = key.name.slice(prefix.length);
  if (!repoId) continue;
  const raw = normalizeBranchPreference((await env.SLACK_KV.get(key.name)) ?? undefined);
  if (!raw || !isValidBranchName(raw)) continue;
  preferences.set(repoId, raw);
}
```

For typical users with few overrides this is harmless, but `getUserRepoBranchPreferences` is called on every `publishAppHome()` invocation.

packages/slack-bot/src/index.ts
```diff
@@ -1184,19 +1805,79 @@ async function handleRepoSelection(
...
+      }
+    }
+
+    if (!repoId) {
+      log.warn("slack.repo_branch_pref.missing_repo", { trace_id: traceId, user_id: userId });
+      await publishAppHome(env, userId);
+      return;
+    }
+
+    await saveUserRepoBranchPreference(env, userId, repoId, branch);
```
ColeMurray, line 1689: [Automated Review] **Suggestion — `repoId` from `private_metadata` not validated against known repos**

The `repoId` is extracted from `private_metadata` and passed directly to `saveUserRepoBranchPreference` without checking whether it corresponds to an actual available repository. In normal Slack modal flow `private_metadata` is server-controlled so this is low risk, but a guard would prevent phantom KV entries for deleted/inaccessible repos.

Note: the `REPO_BRANCH_SELECTOR_ACTION_ID` handler already validates the repo — this would add the same defense on the submission side.

packages/slack-bot/src/index.ts
```diff
@@ -752,19 +1300,92 @@ app.post("/interactions", async (c) => {
...
+  }
+
+  const branchValidationError = getBranchSubmissionValidationError(payload);
+
+  if (branchValidationError) {
+    log.warn("slack.branch_pref.invalid", {
+      trace_id: traceId,
+      user_id: payload.user?.id,
+      branch:
+        normalizeBranchPreference(
```
ColeMurray: [Automated Review] **Nit — Duplicated branch value extraction for logging**

The branch value is re-extracted via a deep `?.` chain here instead of reusing the value that `getBranchSubmissionValidationError` already computed. Consider extracting the branch value once before the validation call and reusing it for both validation and logging.

packages/slack-bot/src/index.ts
```diff
@@ -34,6 +35,38 @@ import {
 
 const log = createLogger("handler");
 
+const BRANCH_MODAL_CALLBACK_ID = "branch_preference_modal";
```
ColeMurray: [Automated Review] **Suggestion — Consider extracting branch preference logic into its own module**

This PR adds ~770 lines to `index.ts`. The branch preference functions (`getBranchValidationError`, `isValidBranchName`, `normalizeBranchPreference`, KV helpers, modal builders) are self-contained and could live in a dedicated module (e.g., `src/branch-preferences.ts`). This would improve readability and make the validation logic independently unit-testable.

packages/slack-bot/src/index.ts
```diff
@@ -34,6 +35,38 @@ import {
...
+const REPO_BRANCH_MODAL_CALLBACK_ID = "repo_branch_preference_modal";
+const BRANCH_INPUT_BLOCK_ID = "branch_input";
+const BRANCH_INPUT_ACTION_ID = "branch_value";
+const REPO_BRANCH_SELECTOR_ACTION_ID = "select_repo_branch_override";
+const CLEAR_REPO_BRANCH_ACTION_ID = "clear_repo_branch_override";
+const MAX_REPO_SUGGESTION_OPTIONS = 100;
+const BRANCH_NAME_SPECIAL_CHARS_REGEX = /[\s~^:?*[\\]/;
```
ColeMurray: [Automated Review] **Nit** — The regex `/[\s~^:?*[\\]/` has an unescaped `[` which works in JS but could confuse readers. Consider escaping it: `/[\s~^:?*\[\\]/`.

packages/slack-bot/src/index.ts
```diff
@@ -34,6 +35,38 @@ import {
...
+const INVALID_BRANCH_ERROR = "Enter a valid Git branch name.";
+
+type SlackInteractionPayload = {
```
ColeMurray: [Automated Review] **Nit** — `SlackInteractionPayload` is defined inline in the handler file. Consider placing it in `./types.ts` alongside the existing `Env` type for consistency.

packages/slack-bot/src/index.ts
```diff
@@ -298,6 +334,203 @@ function getUserPreferencesKey(userId: string): string {
...
+ */
+function normalizeBranchPreference(branch: string | undefined): string | undefined {
+  const normalized = branch?.trim();
+  return normalized ? normalized : undefined;
+}
+
+/**
+ * Check whether a branch name is syntactically valid.
+ */
+function hasControlCharacters(value: string): boolean {
```
ColeMurray: [Automated Review] **Nit** — `hasControlCharacters` iterates with `Array.from(value).some(...)`. A regex `/[\x00-\x1f\x7f]/` would be more concise and avoid the intermediate array allocation.

packages/slack-bot/src/index.test.ts
```diff
@@ -0,0 +1,442 @@
...
+      options: Array<{ text: { type: string; text: string }; value: string }>;
+    };
+    expect(body.options).toEqual([
+      {
+        text: { type: "plain_text", text: "acme/repo-150" },
+        value: "acme/repo-150",
+      },
+    ]);
+  });
+});
```
ColeMurray, line 611: [Automated Review] **Suggestion — Missing test for session creation branch resolution**

The tests cover validation, modal interactions, and KV persistence well, but there's no test for the core business logic: branch resolution precedence (repo-specific override → global → default) during session creation. Consider adding a test that sets both a global and repo-specific branch, starts a session, and asserts the correct branch is passed to `createSession`.

Also a nit: the test uses `repoId: "acme/app"` (string) but mock repos use numeric `id: 1`. The KV operations work because they're string-typed, but the inconsistency could mask type bugs.

ColeMurray (changes requested):
[Automated Review]

Well-structured contribution adding branch preferences to the Slack bot. The validation logic is thorough, async patterns are correct, and test coverage is solid. Two issues should be addressed before merge — see inline comments.

ColeMurray:
> @ColeMurray Got only automated reviews here if not mistaken (some of them are duplicates BTW). Is it good for you in terms of technical and UX solution ? If it is, I'll fix the various items

yes, looks good! let's fix the identified issues and should be good to go

ColeMurray (approved):

PR #343, opened 2026-03-13
feat: MCP server support — CRUD, settings UI, sandbox integration

bencered:
## Summary

Adds full MCP (Model Context Protocol) server management: configure servers via the settings UI, persist them in D1, and inject them into agent sandboxes at runtime.

### Control Plane
- **D1 migration** (`0014_create_mcp_servers.sql`) — `mcp_servers` table with JSON `repo_scope` column
- **`McpServerStore`** — CRUD operations with repo-scope filtering (JS-side, D1 lacks JSON functions)
- **REST API routes** — `GET/POST /mcp-servers`, `GET/PUT/DELETE /mcp-servers/:id`
- **Lifecycle manager** — `loadMcpServers()` fetches and injects MCP config into sandbox creation
- **Sandbox client/provider** — `mcpServers` field threaded through create/restore requests

### Modal Sandbox
- **`_resolve_mcp_servers()`** — loads from session config or falls back to control plane API
- **`_install_mcp_packages()`** — pre-installs npm packages for stdio servers (`npx` → `npm install -g`)
- **`_build_mcp_config()`** — converts server list to OpenCode config format
- **`_generate_internal_token()`** — HMAC auth for control plane API fallback
- **Image build mode** — pre-installs MCP packages during `modal deploy`
- **`manager.py`** — passes `CONTROL_PLANE_URL` + `INTERNAL_CALLBACK_SECRET` to build sandboxes

### Web Dashboard
- **Settings UI** — accordion-style MCP server editor using shadcn components (`Input`, `Textarea`, `Label`, `Switch`, `Checkbox`, `RadioCard`, `Badge`)
- **SWR hook** (`use-mcp-servers.ts`) — list/create/update/delete with optimistic mutation
- **Next.js API proxy** — `/api/mcp-servers` and `/api/mcp-servers/[id]`
- **Settings nav** — MCP Servers tab added

### Shared
- **`McpServerConfig`** type in `@open-inspect/shared`

### Cleanup
- Removed debug timing instrumentation from sessions route

## Testing
- TypeScript typecheck passes (all workspaces)
- 92 tests pass
- Lint + Prettier clean
- No conflicts with `main`

### DIFF
--- a/packages/control-plane/src/db/mcp-servers.ts
+++ b/packages/control-plane/src/db/mcp-servers.ts
@@ -1,0 +1,321 @@
+import type { McpServerConfig } from "@open-inspect/shared";
+import { encryptToken, decryptToken } from "../auth/crypto";
+import { createLogger } from "../logger";
+
+const log = createLogger("db:mcp-servers");
+
+export class McpServerValidationError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "McpServerValidationError";
+  }
+}
+
+function generateId(): string {
+  return crypto.randomUUID().replace(/-/g, "").slice(0, 16);
+}
+
+interface McpServerRow {
+  id: string;
+  name: string;
+  type: string;
+  command: string | null;
+  url: string | null;
+  env: string;
+  repo_scope: string | null;
+  enabled: number;
+  created_at: number;
+  updated_at: number;
+}
+
+function parseRepoScopes(raw: string | null): string[] | null {
+  if (!raw) return null;
+  try {
+    const parsed = JSON.parse(raw);
+    return Array.isArray(parsed) ? parsed : [raw];
+  } catch {
+    return [raw];
+  }
+}
+
+function safeJsonParseCommand(raw: string | null): string[] | undefined {
+  if (!raw) return undefined;
+  try {
+    return JSON.parse(raw);
+  } catch {
+    return [raw];
+  }
+}
+
+function safeJsonParseEnv(raw: string): Record<string, string> {
+  try {
+    return JSON.parse(raw);
+  } catch {
+    return {};
+  }
+}
+
+function rowToConfig(row: McpServerRow, payload: Record<string, string>): McpServerConfig {
+  // The DB `env` column stores different things depending on server type:
+  //   stdio  → process environment variables  → McpServerConfig.env
+  //   remote → HTTP request headers           → McpServerConfig.headers
+  const envOrHeaders: Pick<McpServerConfig, "env" | "headers"> =
+    row.type === "remote" ? { headers: payload } : { env: payload };
+  return {
+    id: row.id,
+    name: row.name,
+    type: row.type as "stdio" | "remote",
+    command: safeJsonParseCommand(row.command),
+    url: row.url ?? undefined,
+    ...envOrHeaders,
+    repoScopes: parseRepoScopes(row.repo_scope),
+    enabled: row.enabled === 1,
+  };
+}
+
+/**
+ * Check if an error message indicates a D1 UNIQUE constraint violation.
+ *
+ * NOTE: D1 does not expose structured error codes, so we string-match against the
+ * SQLite error message. If Cloudflare ever changes the wording this check silently
+ * becomes a no-op (constraint errors fall through as 503 instead of 400). Keep an
+ * eye on this if D1 error shapes change in future Worker runtime versions.
+ */
+function isUniqueConstraintError(err: unknown): boolean {
+  const msg = err instanceof Error ? err.message : String(err);
+  return msg.toLowerCase().includes("unique constraint failed");
+}
+
+export class McpServerStore {
+  constructor(
+    private readonly db: D1Database,
+    private readonly encryptionKey?: string
+  ) {}
+
+  /** Encrypt env dict to a stored string. Falls back to plaintext if no key. */
+  private async encryptEnv(env: Record<string, string>): Promise<string> {
+    const plain = JSON.stringify(env);
+    if (!this.encryptionKey) return plain;
+    return encryptToken(plain, this.encryptionKey);
+  }
+
+  /**
+   * Decrypt env string from storage.
+   *
+   * Three cases:
+   * 1. No encryption key → treat as plaintext JSON (dev / unconfigured)
+   * 2. Key present, decrypt succeeds → normal encrypted row
+   * 3. Key present, decrypt fails → pre-encryption plaintext row (migration period).
+   *    Falls back to JSON parse; logs a warning so ops can tell the difference from
+   *    a key rotation mistake.
+   *
+   * NOTE: env values are returned in plaintext to callers. This is intentional —
+   * they are required by the sandbox at runtime. Transport to Modal is over TLS.
+   * Do NOT log the returned env object.
+   */
+  private async decryptEnv(raw: string): Promise<Record<string, string>> {
+    if (!this.encryptionKey) return safeJsonParseEnv(raw);
+    try {
+      const plain = await decryptToken(raw, this.encryptionKey);
+      return safeJsonParseEnv(plain);
+    } catch {
+      // Decryption failed — check if it looks like a pre-encryption plaintext row
+      const plaintext = safeJsonParseEnv(raw);
+      if (Object.keys(plaintext).length > 0) {
+        log.warn("MCP server env decryption failed — treating as pre-encryption plaintext row", {
+          event: "mcp_server.env_decrypt_fallback",
+        });
+        return plaintext;
+      }
+      // Empty JSON or ciphertext with wrong key — log a clear error, return empty
+      log.error("MCP server env decryption failed and raw value is not plaintext JSON", {
+        event: "mcp_server.env_decrypt_error",
+      });
+      return {};
+    }
+  }
+
+  private async decryptRow(row: McpServerRow): Promise<McpServerConfig> {
+    const env = await this.decryptEnv(row.env);
+    return rowToConfig(row, env);
+  }
+
+  async list(repoScope?: string): Promise<McpServerConfig[]> {
+    // NOTE: We load all rows then filter in-memory because repo_scope is stored as
+    // JSON-serialised array, making SQL-level filtering awkward in D1/SQLite without
+    // JSON_EACH. At expected scale (<100 MCP servers per deployment) this is fine.
+    // TODO: push down to SQL once D1 supports JSON_EACH in WHERE clauses.
+    const { results } = await this.db
+      .prepare("SELECT * FROM mcp_servers ORDER BY name")
+      .all<McpServerRow>();
+    const configs = await Promise.all(results.map((r) => this.decryptRow(r)));
+    if (repoScope === undefined) return configs;
+    const normalized = repoScope.toLowerCase();
+    return configs.filter((c) => {
+      if (!c.repoScopes) return true;
+      return c.repoScopes.some((s) => s.toLowerCase() === normalized);
+    });
+  }
+
+  async get(id: string): Promise<McpServerConfig | null> {
+    const row = await this.db
+      .prepare("SELECT * FROM mcp_servers WHERE id = ?")
+      .bind(id)
+      .first<McpServerRow>();
+    return row ? this.decryptRow(row) : null;
+  }
+
+  async create(config: Omit<McpServerConfig, "id">): Promise<McpServerConfig> {
+    const id = generateId();
+    const now = Date.now();
+
+    if (config.type === "stdio" && (!config.command || config.command.length === 0)) {
+      throw new McpServerValidationError("stdio MCP servers require a command");
+    }
+    if (config.type === "remote" && !config.url) {
+      throw new McpServerValidationError("remote MCP servers require a URL");
+    }
+
+    // For remote servers, the DB `env` column stores HTTP headers (McpServerConfig.headers).
+    // For stdio servers, it stores process environment variables (McpServerConfig.env).
+    const encryptedEnv = await this.encryptEnv(
+      config.type === "remote" ? (config.headers ?? {}) : (config.env ?? {})
+    );
+
+    try {
+      await this.db
+        .prepare(
+          `INSERT INTO mcp_servers (id, name, type, command, url, env, repo_scope, enabled, created_at, updated_at)
+           VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
+        )
+        .bind(
+          id,
+          config.name,
+          config.type,
+          config.command ? JSON.stringify(config.command) : null,
+          config.url ?? null,
+          encryptedEnv,
+          config.repoScopes?.length
+            ? JSON.stringify(config.repoScopes.map((r) => r.toLowerCase()))
+            : null,
+          config.enabled ? 1 : 0,
+          now,
+          now
+        )
+        .run();
+    } catch (err) {
+      if (isUniqueConstraintError(err)) {
+        throw new McpServerValidationError(`An MCP server named '${config.name}' already exists`);
+      }
+      throw err;
+    }
+
+    const created = await this.get(id);
+    if (!created) {
+      throw new Error(`MCP server '${id}' not found after insert — this should not happen`);
+    }
+    return created;
+  }
+
+  async update(
+    id: string,
+    patch: Partial<
+      Pick<
+        McpServerConfig,
+        "name" | "type" | "command" | "url" | "env" | "headers" | "repoScopes" | "enabled"
+      >
+    >
+  ): Promise<McpServerConfig | null> {
+    const existing = await this.get(id);
+    if (!existing) return null;
+
+    // Explicit allowlist — id, created_at, updated_at cannot be overwritten
+    const merged: Omit<McpServerConfig, "id"> = {
+      name: patch.name ?? existing.name,
+      type: patch.type ?? existing.type,
+      command: patch.command !== undefined ? patch.command : existing.command,
+      url: patch.url !== undefined ? patch.url : existing.url,
+      env: patch.env !== undefined ? patch.env : existing.env,
+      headers: patch.headers !== undefined ? patch.headers : existing.headers,
+      repoScopes: patch.repoScopes !== undefined ? patch.repoScopes : existing.repoScopes,
+      enabled: patch.enabled !== undefined ? patch.enabled : existing.enabled,
+    };
+    // Validate the merged result — catches cases where type is changed without
+    // updating the corresponding command/url field (same rules as create()).
+    if (merged.type === "stdio" && (!merged.command || merged.command.length === 0)) {
+      throw new McpServerValidationError("stdio MCP servers require a command");
+    }
+    if (merged.type === "remote" && !merged.url) {
+      throw new McpServerValidationError("remote MCP servers require a URL");
+    }
+
+    const now = Date.now();
+    // For remote servers, the DB `env` column stores HTTP headers; for stdio, process env.
+    const encryptedEnv = await this.encryptEnv(
+      merged.type === "remote" ? (merged.headers ?? {}) : (merged.env ?? {})
+    );
+
+    try {
+      await this.db
+        .prepare(
+          `UPDATE mcp_servers SET name = ?, type = ?, command = ?, url = ?, env = ?, repo_scope = ?, enabled = ?, updated_at = ?
+           WHERE id = ?`
+        )
+        .bind(
+          merged.name,
+          merged.type,
+          merged.command ? JSON.stringify(merged.command) : null,
+          merged.url ?? null,
+          encryptedEnv,
+          merged.repoScopes?.length
+            ? JSON.stringify(merged.repoScopes.map((r) => r.toLowerCase()))
+            : null,
+          merged.enabled ? 1 : 0,
+          now,
+          id
+        )
+        .run();
+    } catch (err) {
+      if (isUniqueConstraintError(err)) {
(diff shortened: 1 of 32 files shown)


ColeMurray:
can you separate out the MCP specific changes in this PR

bencered:
Removed the extraneous changes and refactored the MCP UI to use shadcn/ui components.

package-lock.json
```diff
@@ -1590,14 +1590,14 @@
       }
```
ColeMurray, line 1590: most of these look like version downgrades

packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -366,25 +366,147 @@ async def _forward_code_server_logs(self) -> None:
...
+                continue
+            # npx @scope/package  →  install @scope/package
+            # npx -y @scope/package  →  install @scope/package
+            parts = [c for c in cmd if isinstance(c, str)]
+            if not parts or parts[0] != "npx":
+                continue
+            # Skip flags like -y, --yes, -p, --package
+            pkg_parts = [p for p in parts[1:] if not p.startswith("-")]
+            if pkg_parts:
+                packages.append(pkg_parts[0])
```
ColeMurray: **[Security]** The package name extracted here is passed directly to `subprocess.run(["npm", "install", "-g", *packages])` without validation. While using a list (not a shell string) avoids shell injection, there's no check that the extracted string is a valid npm package name.

A malicious or misconfigured MCP server config could inject unexpected package names. Consider validating against a pattern like `^(@[\w.-]+\/)?[\w.-]+(@[\w.-]+)?$` before passing to npm:

```python
import re
NPM_PKG_RE = re.compile(r'^(@[\w.-]+/)?[\w.-]+(@[\w.-]+)?$')
# ...
if pkg_parts and NPM_PKG_RE.match(pkg_parts[0]):
    packages.append(pkg_parts[0])
```

packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -366,25 +366,147 @@ async def _forward_code_server_logs(self) -> None:
...
+
+        # Fallback: fetch from control plane API if not passed in session config
+        cp_url = self.control_plane_url
+        secret = os.environ.get("INTERNAL_CALLBACK_SECRET", "")
+        if not cp_url or not secret:
+            return []
+
+        try:
+            token = self._generate_internal_token(secret)
+            resp = httpx.get(
```
ColeMurray: **[Correctness]** `_resolve_mcp_servers` uses synchronous `httpx.get()`, but it's called from `async def start_opencode()` (line 502). This will block the event loop during the HTTP request (up to the 5s timeout).

For the image build path (called from sync `main()`), sync is fine. But for the `start_opencode` path, consider either:
1. Using `httpx.AsyncClient` and making this method `async`
2. Running the sync call in a thread: `await asyncio.to_thread(self._resolve_mcp_servers)`

packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -401,6 +523,12 @@ async def start_opencode(self) -> None:
...
 
+        # Write config to .opencode directory
+        opencode_dir.mkdir(parents=True, exist_ok=True)
+        config_file = opencode_dir / "opencode.json"
+        config_file.write_text(json.dumps(opencode_config, indent=2))
+        self.log.info("opencode.config_written", path=str(config_file))
+
         env = {
             **os.environ,
             "OPENCODE_CONFIG_CONTENT": json.dumps(opencode_config),
```
ColeMurray, line 546: **[Correctness]** The MCP config is now written to both `.opencode/opencode.json` (line 529) *and* passed via `OPENCODE_CONFIG_CONTENT` env var here. If OpenCode reads both sources, there could be conflicts or double-loading. Which takes precedence? Consider removing one to avoid ambiguity.

packages/control-plane/src/db/mcp-servers.ts
```diff
@@ -0,0 +1,160 @@
...
+      repoScopes = Array.isArray(parsed) ? parsed : [row.repo_scope];
+    } catch {
+      repoScopes = [row.repo_scope];
+    }
+  }
+  return {
+    id: row.id,
+    name: row.name,
+    type: row.type as "stdio" | "remote",
+    command: row.command ? JSON.parse(row.command) : undefined,
```
ColeMurray: **[Robustness]** `JSON.parse(row.command)` and `JSON.parse(row.env)` (line 36) are called without try/catch. If corrupted data exists in D1, this will throw and crash the entire list/get response.

The `repo_scope` parsing on line 24 already has a defensive try/catch — the same pattern should be applied here:

```typescript
command: row.command ? (() => { try { return JSON.parse(row.command); } catch { return [row.command]; } })() : undefined,
```

Or extract a `safeJsonParse` helper.

packages/web/src/components/settings/mcp-servers-settings.tsx
```diff
@@ -0,0 +1,530 @@
...
+  repoScopes: [],
+  scopeMode: "global",
+  enabled: true,
+};
+
+function configToForm(config: McpServerConfig): FormState {
+  return {
+    name: config.name,
+    type: config.type,
+    command: config.command?.join(" ") ?? "",
```
ColeMurray: **[Correctness]** `configToForm` joins the command array with spaces, but `parseCommand` handles quoted arguments. This means the round-trip is lossy: if a command was saved as `["node", "path with spaces"]`, it displays as `node path with spaces` and re-parses as `["node", "path", "with", "spaces"]`.

Consider quoting tokens that contain whitespace when converting back to the display string:

```typescript
command: config.command?.map(t => /\s/.test(t) ? `"${t}"` : t).join(" ") ?? "",
```

packages/modal-infra/src/sandbox/manager.py
```diff
@@ -274,6 +275,11 @@ async def create_build_sandbox(
         if user_env_vars:
             env_vars.update(user_env_vars)
 
+        # Pass CONTROL_PLANE_URL and INTERNAL_CALLBACK_SECRET so the build
+        # sandbox can fetch MCP servers and pre-install their packages.
+        control_plane_url = os.environ.get("CONTROL_PLANE_URL", "")
+        internal_secret = os.environ.get("INTERNAL_CALLBACK_SECRET", "")
```
ColeMurray: **[Security]** `INTERNAL_CALLBACK_SECRET` is a long-lived secret used for control-plane HMAC auth. Passing it into image build sandboxes means it could be captured in the filesystem snapshot if it's written to a file, log, or `/proc` entry during the build.

Consider generating a short-lived, scoped token for the build sandbox instead of passing the master secret — or at minimum, ensure the entrypoint clears it from the environment after use.

packages/web/src/components/settings/mcp-servers-settings.tsx
```diff
@@ -0,0 +1,530 @@
...
+      ) : (
+        <div>
+          <Label className="mb-1">Command</Label>
+          <Input
+            value={form.command}
+            onChange={(e) => setForm({ ...form, command: e.target.value })}
+            placeholder="npx -y @playwright/mcp"
+          />
+          <p className="text-xs text-muted-foreground mt-1">
+            Space-separated command and arguments
```
ColeMurray: **[DX]** The help text says "Space-separated command and arguments" but `parseCommand` (line 78) actually supports quoted strings (`"..."` and `'...'`). Consider updating this to: "Space-separated command and arguments. Use quotes for arguments with spaces."

packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -366,25 +366,147 @@ async def _forward_code_server_logs(self) -> None:
                 self.log.info("code_server.stdout", line=line.decode().rstrip())
         except Exception as e:
             self.log.warn("code_server.log_forward_error", exc=e)
+    @staticmethod
```
ColeMurray: Nit: Missing blank line before `@staticmethod` — PEP 8 expects two blank lines between method definitions in a class body.

ColeMurray (commented):
## Review Summary

Solid end-to-end implementation of MCP server management. The architecture is well-designed: D1-backed CRUD store, control-plane REST API, settings UI with SWR, and sandbox-side injection with a resilient fallback pattern. The layering follows existing codebase conventions nicely (consistent with repo secrets).

### Positive Feedback
- **Clean separation of concerns**: `McpServerStore` / routes / SWR hook / UI component layering is well-structured.
- **Defensive fallback pattern**: The sandbox's `_resolve_mcp_servers()` falling back from session config to API fetch is a smart resilience choice.
- **Prior review items addressed**: `list()` correctly filters in JS, `create()` returns via `this.get(id)`, `parseCommand` handles quotes, and `handleDelete` has `window.confirm`.

See inline comments for specific issues and suggestions.

packages/control-plane/src/db/mcp-servers.ts
```diff
@@ -0,0 +1,160 @@
...
+    enabled: row.enabled === 1,
+  };
+}
+
+export class McpServerStore {
+  constructor(private readonly db: D1Database) {}
+
+  async list(repoScope?: string): Promise<McpServerConfig[]> {
+    const { results } = await this.db
+      .prepare("SELECT * FROM mcp_servers ORDER BY name")
```
ColeMurray, line 149: not a huge deal but this is going to load every MCP server and seems like a code-smell

packages/control-plane/src/db/mcp-servers.ts
```diff
@@ -0,0 +1,160 @@
+import type { McpServerConfig } from "@open-inspect/shared";
+
+function generateId(): string {
+  return crypto.randomUUID().replace(/-/g, "").slice(0, 16);
+}
+
+interface McpServerRow {
```
ColeMurray, line 18: based on the underlying openCode spec, we may want to add headers here

packages/modal-infra/src/sandbox/types.py
```diff
@@ -117,3 +117,4 @@ class SessionConfig(BaseModel):
     opencode_session_id: str | None = None
     provider: str = "anthropic"
     model: str = "claude-sonnet-4-6"
+    mcp_servers: list[dict] | None = None
```
ColeMurray: i believe we know the type shape for this based on the typescript definition. we should prefer to use a more defined type here and other parts of the python code

packages/control-plane/src/routes/mcp-servers.ts
```diff
@@ -0,0 +1,154 @@
...
+  request: Request,
+  env: Env,
+  match: RegExpMatchArray,
+  _ctx: RequestContext
+): Promise<Response> {
+  const id = match.groups?.id;
+  if (!id) return error("Missing server ID", 400);
+  if (!env.DB) return error("Database not configured", 500);
+
+  let body: Partial<McpServerConfig>;
```
ColeMurray, line 116: **[Pattern]** The update route uses `PATCH`, but every existing route in the codebase uses `PUT` for updates (secrets, model-preferences, integration-settings, automations). No existing route uses `PATCH`. Should be `PUT` for consistency.

packages/control-plane/src/routes/mcp-servers.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * MCP server configuration routes.
+ */
+
+import type { McpServerConfig } from "@open-inspect/shared";
+import { McpServerStore } from "../db/mcp-servers";
+import type { Env } from "../types";
+import { createLogger } from "../logger";
+import { type Route, type RequestContext, parsePattern, json, error } from "./shared";
```
ColeMurray, line 9: **[Pattern]** Two issues with the "DB not configured" handling across these route handlers:

1. `handleListMcpServers` returns `json([])` when `!env.DB`, silently hiding misconfiguration. Every other route returns an explicit error.
2. The other handlers return `error("Database not configured", 500)` — but the established pattern across the codebase is **`503`**, not `500` (see `repo-images.ts`, `secrets.ts`, `integration-settings.ts`, `automations.ts`).

Both should be `return error("Database not configured", 503);`

packages/control-plane/src/routes/mcp-servers.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  if (!body.name || typeof body.name !== "string") {
+    return error("name is required", 400);
+  }
+  if (body.type !== "stdio" && body.type !== "remote") {
+    return error("type must be 'stdio' or 'remote'", 400);
+  }
+
+  try {
+    const store = new McpServerStore(env.DB);
```
ColeMurray: **[Pattern]** Structured logging is incomplete. Existing routes use `event: "noun.verb"` format with `request_id` and `trace_id` from the `ctx` parameter:

```typescript
logger.info("MCP server created", {
  event: "mcp_server.created",
  request_id: ctx.request_id,
  trace_id: ctx.trace_id,
  id: server.id,
  name: server.name,
});
```

Currently `ctx` is unused (`_ctx`) across all handlers, which breaks log correlation with upstream request tracing.

packages/control-plane/src/db/mcp-servers.ts
```diff
@@ -0,0 +1,160 @@
...
+  async list(repoScope?: string): Promise<McpServerConfig[]> {
+    const { results } = await this.db
+      .prepare("SELECT * FROM mcp_servers ORDER BY name")
+      .all<McpServerRow>();
+    const configs = results.map(rowToConfig);
+    if (repoScope === undefined) return configs;
+    const normalized = repoScope.toLowerCase();
+    return configs.filter((c) => {
+      if (!c.repoScopes) return true; // global servers match all repos
+      return c.repoScopes.some((s) => s.toLowerCase() === normalized);
```
ColeMurray, line 156: **[Pattern]** Existing stores define a custom `ValidationError` subclass (e.g., `SecretsValidationError`, `ModelPreferencesValidationError`) so route handlers can distinguish validation failures (→ 400) from infrastructure failures (→ 503). Using generic `Error` here means the route's catch block can't differentiate — a D1 connection failure would also be caught as a 400.

packages/control-plane/src/db/mcp-servers.ts
```diff
@@ -0,0 +1,160 @@
...
+    });
+  }
+
+  async get(id: string): Promise<McpServerConfig | null> {
+    const row = await this.db
+      .prepare("SELECT * FROM mcp_servers WHERE id = ?")
+      .bind(id)
+      .first<McpServerRow>();
+    return row ? rowToConfig(row) : null;
```
ColeMurray: **[Safety]** The update method uses unconstrained spread:

```typescript
const merged = { ...existing, ...patch };
```

This allows callers to overwrite `id` and other fields that should be immutable. Existing stores (e.g., `AutomationStore.update`) use an explicit allowlist pattern to gate which fields are updateable. This prevents route handlers from accidentally forwarding request body fields like `id` or `created_at`.

ColeMurray (commented):
## Design Review — Integration with Existing Patterns

I focused on how this PR integrates with the existing system conventions. Several of the code-level issues (JSON.parse safety, INTERNAL_CALLBACK_SECRET exposure, sync httpx, etc.) have already been flagged — this review covers the remaining pattern and architectural concerns.

### Unrelated Changes Bundled In

Similarly, the **dependency downgrades** (wrangler 4.73→4.67, workerd, rollup, etc.) are unexplained and deserve their own PR with rationale — downgrading infra deps is risky.

### MCP `env` Values Stored in Plaintext

MCP server `env` fields will likely contain API keys and tokens. These are stored as **plaintext JSON** in D1, unlike repo secrets which use AES-256-GCM encryption via `REPO_SECRETS_ENCRYPTION_KEY`. This is a significant gap — MCP env vars serve the same purpose as repo secrets (injecting credentials into sandboxes) and should get the same at-rest encryption treatment, or alternatively, MCP servers should reference repo secrets by key rather than duplicating a separate unencrypted secret store.

### Redundant Python Fallback Path

The `_resolve_mcp_servers()` fallback that fetches from the control plane API is architecturally concerning beyond just the `INTERNAL_CALLBACK_SECRET` exposure (already flagged). The lifecycle manager already loads MCP servers from D1 and injects them into `session_config` before spawning — so the fallback should never be needed on the primary path. It adds:
- A second code path to maintain and test
- HTTP dependency from sandbox → control plane
- HMAC token generation reimplemented in Python (duplicating `auth/internal.py`)

If the primary injection path is working correctly, this fallback is dead code that expands the attack surface. Consider removing it and relying solely on the `session_config` injection.

### Lifecycle Manager DI Pattern Broken

Adding `db?: D1Database` directly to `SandboxLifecycleConfig` breaks the existing dependency injection pattern. The lifecycle manager was previously pure — it operated through abstract interfaces (`SandboxProvider`, `SandboxStorage`, `RepoImageLookup`). Now it directly depends on `D1Database` and instantiates `McpServerStore` internally.

The pattern-consistent approach would be to either add a `loadMcpServers` method to the `SandboxStorage` interface (paralleling `getUserEnvVars()`), or inject an `McpServerLookup` interface similar to `repoImageLookup`.

### No Tests

The codebase has 363 unit tests and 75 integration tests. This PR adds zero tests for the new McpServerStore, route handlers, Python MCP logic, or web proxy routes. Given the security sensitivity (credential handling, package installation), test coverage here is important.

packages/control-plane/src/db/mcp-servers.ts
```diff
@@ -0,0 +1,160 @@
...
+  env: string;
+  repo_scope: string | null;
+  enabled: number;
+  created_at: number;
+  updated_at: number;
+}
+
+function rowToConfig(row: McpServerRow): McpServerConfig {
+  let repoScopes: string[] | null = null;
+  if (row.repo_scope) {
```
ColeMurray: **[DRY]** The `repo_scope` JSON parse + fallback logic is duplicated between `rowToConfig()` (here) and `getForSession()` (line 103). Consider extracting to a single `parseRepoScopes(raw: string | null): string[] | null` helper.

terraform/d1/migrations/0014_create_mcp_servers.sql
```diff
@@ -0,0 +1,13 @@
+-- MCP server configurations
+CREATE TABLE IF NOT EXISTS mcp_servers (
+  id         TEXT PRIMARY KEY,
+  name       TEXT NOT NULL,
```
ColeMurray, line 4: **[Correctness]** There is no uniqueness constraint on `name`. Since the Python `_build_mcp_config` uses `name` as the dict key, duplicate names will silently cause the last one to win. Consider adding `UNIQUE(name)` to prevent this at the storage layer.

packages/control-plane/src/sandbox/lifecycle/manager.ts
```diff
@@ -139,6 +141,8 @@ export interface SandboxLifecycleConfig {
   model: string;
   /** Session ID for log correlation. Optional — logs will omit sessionId if not provided. */
   sessionId?: string;
+  /** D1 database for loading MCP servers. */
```
ColeMurray: **[Architecture]** Adding `db?: D1Database` directly here breaks the existing DI pattern — see top-level review comment. The lifecycle manager previously had no direct infrastructure dependencies; everything flowed through abstract interfaces. Consider an `McpServerLookup` interface (paralleling `repoImageLookup`) or adding this to `SandboxStorage`.

packages/web/src/hooks/use-mcp-servers.ts
```diff
@@ -0,0 +1,63 @@
...
+    body: JSON.stringify(config),
+  });
+  if (!response.ok) {
+    const error = await response.json().catch(() => ({}));
+    throw new Error(error.error || "Failed to create MCP server");
+  }
+  const server = await response.json();
+  await globalMutate(MCP_SERVERS_KEY);
+  return server;
+}
```
ColeMurray, line 32: **[Perf]** The mutation functions (`createMcpServer`, `updateMcpServer`, `deleteMcpServer`) each call `globalMutate(MCP_SERVERS_KEY)` internally. But the component's `save()` function also calls `mutate()` after invoking them. This triggers two SWR revalidation fetches per mutation. Either remove `globalMutate` from the standalone functions (let the caller decide when to revalidate) or remove the `mutate()` call in the component.

packages/web/src/components/settings/mcp-servers-settings.tsx
```diff
@@ -0,0 +1,530 @@
...
+          checked={form.enabled}
+          onCheckedChange={(checked) => setForm({ ...form, enabled: checked })}
+        />
+        <Label htmlFor={`mcp-enabled-${radioPrefix}`}>Enabled</Label>
+      </div>
+    </>
+  );
+}
+
+export function McpServersSettings() {
```
ColeMurray, line 300: **[Pattern]** Other settings components (`secrets-settings.tsx`, `models-settings.tsx`) use a `useEffect` to sync SWR data into local form state when it first loads or when SWR revalidates in the background. This component only sets form state via explicit user actions (`startEdit`). If SWR data refreshes in the background (e.g., another tab updates a server), the open edit form will show stale data with no indication.

packages/control-plane/src/source-control/providers/github-provider.ts
```diff
@@ -204,93 +206,122 @@ export class GitHubSourceControlProvider implements SourceControlProvider {
...
-      const repo = await getInstallationRepository(
-        this.appConfig,
-        config.owner,
-        config.name,
-        this.kvCache ? { REPOS_CACHE: this.kvCache } : undefined
-      );
-      if (!repo) {
-        return null;
+    // Try each installation until one has access
+    for (const appConfig of this.allAppConfigs) {
```
ColeMurray, line 217: we're picking up some of the multi-app changes here

PR #344, opened 2026-03-13
fix(web): replace instant-logout avatar with UserMenu dropdown

bencered:
The profile avatar in the sidebar triggers sign-out on click with no confirmation. This replaces it with a dropdown menu.

### Changes
- Extract `UserMenu` component from inline avatar button
- Fixed positioning so dropdown escapes sidebar `overflow: hidden`
- Click-outside-to-dismiss
- Dropdown opens below avatar (not above)
- Shows user name + sign out option

Screenshot 2026-03-13 at 15 31 33


ColeMurray:
@open-inspect[bot] can you fix the failing lint & format action

PR #346, opened 2026-03-14
feat: multi-installation GitHub App support + images search

bencered:
### Multi-installation GitHub App support
- `getAllGitHubAppConfigs()` parses comma-separated `GITHUB_APP_INSTALLATION_ID` env var
- `listRepositories()` merges and deduplicates repos across all installations
- `checkRepositoryAccess()` falls back through installations; throws when ALL fail (vs silent null)
- Empty installation IDs filtered out (trailing commas, whitespace)

### Images settings search
- Added search input to filter repositories on the images settings page
- `aria-label` for accessibility

### Tests
- 37 tests covering parsing edge cases, multi-installation fallback, error surfacing, deduplication

### Config
- `terraform.tfvars` updated with comma-separated installation IDs (gitignored)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,4 @@ terraform/environments/production/plan.out
 terraform/environments/production/backend_override.tf
 terraform/environments/production/terraform.tfstate*
 terraform/environments/production/.terraform*
+scripts/deploy-cf.sh
--- a/packages/control-plane/src/auth/github-app.test.ts
+++ b/packages/control-plane/src/auth/github-app.test.ts
@@ -2,6 +2,7 @@ import { afterEach, describe, it, expect, vi } from "vitest";
 import {
   isGitHubAppConfigured,
   getGitHubAppConfig,
+  getAllGitHubAppConfigs,
   getCachedInstallationToken,
   INSTALLATION_TOKEN_CACHE_MAX_AGE_MS,
   INSTALLATION_TOKEN_MIN_REMAINING_MS,
@@ -114,6 +115,102 @@ describe("github-app utilities", () => {
           GITHUB_APP_PRIVATE_KEY: "key",
         })
       ).toBeNull();
+    });
+  });
+
+  describe("getGitHubAppConfig with comma-separated IDs", () => {
+    const baseEnv = {
+      GITHUB_APP_ID: "123",
+      GITHUB_APP_PRIVATE_KEY: "key",
+    };
+
+    it("returns the first installation ID when comma-separated", () => {
+      const config = getGitHubAppConfig({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: "111,222",
+      });
+      expect(config?.installationId).toBe("111");
+    });
+
+    it("trims whitespace from the first ID", () => {
+      const config = getGitHubAppConfig({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: " 111 , 222 ",
+      });
+      expect(config?.installationId).toBe("111");
+    });
+
+    it("returns null when first ID is empty (trailing comma)", () => {
+      const config = getGitHubAppConfig({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: ",222",
+      });
+      expect(config).toBeNull();
+    });
+
+    it("returns null when all IDs are whitespace", () => {
+      const config = getGitHubAppConfig({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: " , , ",
+      });
+      expect(config).toBeNull();
+    });
+  });
+
+  describe("getAllGitHubAppConfigs", () => {
+    const baseEnv = {
+      GITHUB_APP_ID: "123",
+      GITHUB_APP_PRIVATE_KEY: "key",
+    };
+
+    it("returns one config per installation ID", () => {
+      const configs = getAllGitHubAppConfigs({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: "111,222",
+      });
+      expect(configs).toHaveLength(2);
+      expect(configs[0].installationId).toBe("111");
+      expect(configs[1].installationId).toBe("222");
+    });
+
+    it("trims whitespace from IDs", () => {
+      const configs = getAllGitHubAppConfigs({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: " 111 , 222 ",
+      });
+      expect(configs[0].installationId).toBe("111");
+      expect(configs[1].installationId).toBe("222");
+    });
+
+    it("filters out empty IDs from trailing commas", () => {
+      const configs = getAllGitHubAppConfigs({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: "111,,222,",
+      });
+      expect(configs).toHaveLength(2);
+      expect(configs[0].installationId).toBe("111");
+      expect(configs[1].installationId).toBe("222");
+    });
+
+    it("returns empty array for whitespace-only IDs", () => {
+      const configs = getAllGitHubAppConfigs({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: " , , ",
+      });
+      expect(configs).toHaveLength(0);
+    });
+
+    it("returns empty array when not configured", () => {
+      expect(getAllGitHubAppConfigs({})).toEqual([]);
+    });
+
+    it("handles single ID without commas", () => {
+      const configs = getAllGitHubAppConfigs({
+        ...baseEnv,
+        GITHUB_APP_INSTALLATION_ID: "111",
+      });
+      expect(configs).toHaveLength(1);
+      expect(configs[0].installationId).toBe("111");
     });
   });
 
--- a/packages/control-plane/src/auth/github-app.ts
+++ b/packages/control-plane/src/auth/github-app.ts
@@ -615,6 +615,8 @@ export function isGitHubAppConfigured(env: {
 
 /**
  * Get GitHub App config from environment.
+ * Returns config for the first installation ID. Use getAllGitHubAppConfigs
+ * when you need to iterate over multiple installations.
  */
 export function getGitHubAppConfig(env: {
   GITHUB_APP_ID?: string;
@@ -625,9 +627,37 @@ export function getGitHubAppConfig(env: {
     return null;
   }
 
+  const firstInstallationId = env.GITHUB_APP_INSTALLATION_ID!.split(",")[0].trim();
+  if (!firstInstallationId) {
+    return null;
+  }
   return {
     appId: env.GITHUB_APP_ID!,
     privateKey: env.GITHUB_APP_PRIVATE_KEY!,
-    installationId: env.GITHUB_APP_INSTALLATION_ID!,
+    installationId: firstInstallationId,
   };
+}
+
+/**
+ * Get GitHub App configs for ALL installation IDs (comma-separated).
+ * Use this when listing repos or checking access across multiple installations.
+ */
+export function getAllGitHubAppConfigs(env: {
+  GITHUB_APP_ID?: string;
+  GITHUB_APP_PRIVATE_KEY?: string;
+  GITHUB_APP_INSTALLATION_ID?: string;
+}): GitHubAppConfig[] {
+  if (!isGitHubAppConfigured(env)) {
+    return [];
+  }
+
+  return env
+    .GITHUB_APP_INSTALLATION_ID!.split(",")
+    .map((id) => id.trim())
+    .filter((id) => id.length > 0)
+    .map((id) => ({
+      appId: env.GITHUB_APP_ID!,
+      privateKey: env.GITHUB_APP_PRIVATE_KEY!,
+      installationId: id,
+    }));
 }
--- a/packages/control-plane/src/routes/shared.ts
+++ b/packages/control-plane/src/routes/shared.ts
@@ -5,7 +5,7 @@
 import type { CorrelationContext } from "../logger";
 import type { RequestMetrics } from "../db/instrumented-d1";
 import type { Env } from "../types";
-import { getGitHubAppConfig } from "../auth/github-app";
+import { getGitHubAppConfig, getAllGitHubAppConfigs } from "../auth/github-app";
 import {
   createSourceControlProvider,
   resolveScmProviderFromEnv,
@@ -67,11 +67,13 @@ export function error(message: string, status = 400): Response {
  */
 export function createRouteSourceControlProvider(env: Env): SourceControlProvider {
   const appConfig = getGitHubAppConfig(env);
+  const allAppConfigs = getAllGitHubAppConfigs(env);
   const provider = resolveScmProviderFromEnv(env.SCM_PROVIDER);
   return createSourceControlProvider({
     provider,
     github: {
       appConfig: appConfig ?? undefined,
+      allAppConfigs,
       kvCache: env.REPOS_CACHE,
     },
   });
--- a/packages/control-plane/src/session/durable-object.ts
+++ b/packages/control-plane/src/session/durable-object.ts
@@ -11,7 +11,7 @@ import { DurableObject } from "cloudflare:workers";
 import { initSchema } from "./schema";
 import { buildSessionInternalUrl, SessionInternalPaths } from "./contracts";
 import { generateId, hashToken, timingSafeEqual, encryptToken, decryptToken } from "../auth/crypto";
-import { getGitHubAppConfig } from "../auth/github-app";
+import { getGitHubAppConfig, getAllGitHubAppConfigs } from "../auth/github-app";
 import { createModalClient } from "../sandbox/client";
 import { createModalProvider } from "../sandbox/providers/modal-provider";
 import { createLogger, parseLogLevel } from "../logger";
@@ -512,12 +512,14 @@ export class SessionDO extends DurableObject<Env> {
    */
   private createSourceControlProvider(): SourceControlProvider {
     const appConfig = getGitHubAppConfig(this.env);
+    const allAppConfigs = getAllGitHubAppConfigs(this.env);
     const provider = resolveScmProviderFromEnv(this.env.SCM_PROVIDER);
 
     return createSourceControlProviderImpl({
       provider,
       github: {
         appConfig: appConfig ?? undefined,
+        allAppConfigs,
         kvCache: this.env.REPOS_CACHE,
       },
     });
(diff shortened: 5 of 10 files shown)


bencered (commented):
## PR Review: feat/multi-installation-and-search

**Tests:** ✅ All 37 pass. Implementation is well-structured overall. A few issues worth addressing before merge.

---

### 🔴 High

**`getGitHubAppConfig` returns an empty `installationId` for whitespace-only input**

When `GITHUB_APP_INSTALLATION_ID` is a truthy-but-invalid string like `" , , "`, `isGitHubAppConfigured` returns `true` (the string is non-empty), but `getGitHubAppConfig` extracts an empty string as the first installation ID:

```ts
// GITHUB_APP_INSTALLATION_ID = " , , "
const firstInstallationId = env.GITHUB_APP_INSTALLATION_ID!.split(',')[0].trim(); // → ""
return { ..., installationId: '' }; // ← passed downstream as valid config
```

This means callers like `listBranches` and `generatePushAuth` (which use `appConfig` directly) receive a config with an empty `installationId` and make silently broken API calls. `getAllGitHubAppConfigs` correctly guards against this with `.filter(id => id.length > 0)`, but `getGitHubAppConfig` does not.

**Fix:**
```ts
const firstInstallationId = env.GITHUB_APP_INSTALLATION_ID!.split(',')[0].trim();
if (!firstInstallationId) return null; // ← add this guard
```

A test for `getGitHubAppConfig({ ..., GITHUB_APP_INSTALLATION_ID: ' , , ' })` returning `null` would also prevent regression.

---

### 🟡 Medium

**Deduplication key in `listRepositories` is case-sensitive**

```ts
const key = `${repo.owner}/${repo.name}`;
```

GitHub repo names are case-insensitive. If two installations return the same repo with different casing (e.g. `Org/Repo` vs `org/repo`), both will be included. The dedup key should use `.toLowerCase()`:

```ts
const key = `${repo.owner}/${repo.name}`.toLowerCase();
```

---

**Serial token fetches — consider `Promise.allSettled` for `listRepositories`**

The current implementation fetches installations sequentially, adding N serial round-trips. Since the calls are independent, a parallel approach would reduce latency:

```ts
const cacheOpts = this.kvCache ? { REPOS_CACHE: this.kvCache } : undefined;
const results = await Promise.allSettled(
  this.allAppConfigs.map(cfg => listInstallationRepositories(cfg, cacheOpts))
);
for (const r of results) {
  if (r.status === 'rejected') { lastError = r.reason; continue; }
  successCount++;
  for (const repo of r.value.repos) { /* dedupe as before */ }
}
```

Lower priority if N is expected to be small (2–3), but worth noting for future growth.

---

**`listBranches` / `generatePushAuth` use only `this.appConfig` (first installation)**

These methods fall back exclusively to the primary `appConfig`. If a repo is only accessible via a non-primary installation ID, `listBranches` and `generatePushAuth` will fail. If this is an intentional scope limitation, it should be documented. If not, it needs fallback logic consistent with `checkRepositoryAccess`.

---

**Error swallowing in `checkRepositoryAccess` for mixed null/error results**

If installation A returns `null` (definitive not-found) and installation B throws, `anyDefinitiveNotFound` is `true` and the method returns `null`, silently swallowing B's error. Depending on the error type (e.g. GitHub 500), this could hide real problems. The `console.warn` captures it in logs — but consider whether a structured log field is more appropriate for observability.

---

### 🟢 Low

**`isGitHubAppConfigured` truthiness check doesn't validate the ID format**

`!!env.GITHUB_APP_INSTALLATION_ID` is satisfied by any non-empty string including `" , , "`. As noted above this creates a hole in `getGitHubAppConfig`. A tighter check:

```ts
return !!(
  env.GITHUB_APP_ID &&
  env.GITHUB_APP_PRIVATE_KEY &&
  env.GITHUB_APP_INSTALLATION_ID?.split(',').some(id => id.trim())
);
```

---

**Missing test: `getGitHubAppConfig` with whitespace-only IDs**

There is a test asserting `getAllGitHubAppConfigs` returns `[]` for `" , , "`, but no corresponding test for `getGitHubAppConfig` with the same input. The existing suite only covers the happy path (trimming, selecting first ID). Add:

```ts
it('returns null for whitespace-only IDs', () => {
  expect(getGitHubAppConfig({ ...baseEnv, GITHUB_APP_INSTALLATION_ID: ' , , ' })).toBeNull();
});
```

---

**Images search UX: no debounce**

The search input updates state on every keystroke, which triggers a `useMemo` recompute. With a large repo list this could be sluggish. `useDeferredValue` or a simple ~150 ms debounce on `setSearch` would keep it smooth. Not a bug — just a polish note.

---

## Summary

| Severity | Issue |
|---|---|
| 🔴 High | Empty `installationId` passthrough in `getGitHubAppConfig` |
| 🟡 Medium | Case-sensitive dedup key in `listRepositories` |
| 🟡 Medium | Serial installation fetches (performance) |
| 🟡 Medium | `listBranches`/`generatePushAuth` not multi-install-aware |
| 🟡 Medium | Silent error swallow in `checkRepositoryAccess` mixed results |
| 🟢 Low | `isGitHubAppConfigured` doesn't validate ID format |
| 🟢 Low | Missing regression test for whitespace-only ID in `getGitHubAppConfig` |
| 🟢 Low | No debounce on images search input |

The multi-installation architecture is clean and the test coverage is solid. The main blocker is the empty-installationId bug in `getGitHubAppConfig`; everything else can be addressed in follow-up or as judgment calls.

bencered (commented):
## PR #346 v2 Review — Multi-Installation GitHub App Support

All 40 tests pass ✅. The parallelisation approach and overall architecture are solid. A few issues remain before merge.

---

### 🔴 Token-scope bug in  when  throws

**File:**  — 

The inner  swallows **both** the `getInstallationRepository` error *and* any error thrown by the subsequent `getCachedInstallationToken` call:

```ts
for (const appConfig of this.allAppConfigs) {
  try {
    const repo = await getInstallationRepository(appConfig, ...);
    if (repo) {
      const token = await getCachedInstallationToken(appConfig); // ← if this throws...
      return { authType: "app", token };
    }
  } catch {
    // ...we silently fall through to the NEXT installation or the primary fallback
  }
}
```

**Scenario:** Two installations. The target repo lives in installation B. `getInstallationRepository(B)` succeeds (repo found), but `getCachedInstallationToken(B)` throws (e.g. RSA key wrong, rate-limited). The catch block silently falls through to the primary-installation fallback and returns a token for installation A — which does **not** have write access to the repo. The git push will fail with an opaque 403, hiding the real cause.

**Fix:** Split the try/catch so only the repository lookup is guarded:

```ts
for (const appConfig of this.allAppConfigs) {
  let repo: InstallationRepository | null = null;
  try {
    repo = await getInstallationRepository(appConfig, repoOwner, repoName, ...);
  } catch {
    continue; // repo lookup failed, try next installation
  }
  if (repo) {
    // Token errors should propagate — we know this is the right installation
    const token = await getCachedInstallationToken(appConfig);
    return { authType: "app", token };
  }
}
```

---

### 🟡 Silent fallback when repo is not found in any installation ()

When `repoOwner`/`repoName` are provided but no installation claims the repo (all return null or throw), the code silently falls back to `allAppConfigs[0]`'s token without any warning. The caller receives a token that is almost certainly scoped to the wrong installation, and the subsequent git push will fail with a confusing error.

**Suggestion:** At minimum, add a `console.warn` before the fallback path:

```ts
// After the for-loop exits without returning
console.warn(
  `generatePushAuth: repo ${repoOwner}/${repoName} not found in any installation; falling back to primary`
);
```

Alternatively, throw here rather than silently proceeding, since a token known to lack access provides no benefit.

---

### 🟡  fails fast on non-404/403 errors — asymmetric with 

```ts
// listBranches
const status = extractHttpStatus(error);
if (status === 404 || status === 403) {
  lastError = error;
  continue;          // skip to next installation
}
throw SourceControlProviderError.fromFetchError(...); // ← throws immediately for 500, timeout, etc.
```

By contrast, `listRepositories` uses `Promise.allSettled` and tries all installations regardless of error type. A transient 500 or network timeout in the **first** installation means the second installation is never tried, even though it might have access to the repo.

Consider also retrying on `status === undefined` (network errors) and 5xx codes, consistent with how `listRepositories` behaves.

---

### 🟡 Missing test coverage for new  paths

The new repo-lookup flow in `generatePushAuth` has zero test coverage. Missing cases:

1. **Happy path**: `repoOwner`/`repoName` provided, repo found in second installation → token for second installation is returned (not primary).
2. **Repo not in any installation**: all `getInstallationRepository` calls return null → falls back to primary token.
3. **Bug case from 🔴 above**: `getInstallationRepository` succeeds for configB, but `getCachedInstallationToken(configB)` throws — current code silently returns wrong token; after fix, should propagate the error.
4. ** multi-installation**: no tests for the fallback-on-404/403 iteration path.

Without test (1), it's impossible to tell from CI whether the repo-lookup feature actually routes to the correct installation.

---

### 🟢 Minor: redundant  logging inside  map

```ts
this.allAppConfigs.map((appConfig) =>
  listInstallationRepositories(appConfig, ...).catch((error) => {
    console.warn(`Failed to list repos ...`);
    throw error; // re-throw so allSettled sees it as rejected
  })
)
```

The `.catch()` re-throws anyway, so its only effect is the `console.warn`. That's fine, but it makes the code look more complex than it is. Consider either logging in the `for (const result of results)` loop (where you already have access to `result.reason`) or leaving a comment explaining the intent.

---

### Summary

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | 🔴 |  |  swallows token errors; wrong installation's token silently returned |
| 2 | 🟡 |  | Silent fallback when repo not found in any installation |
| 3 | 🟡 |  | Non-404/403 errors abort iteration; inconsistent with  |
| 4 | 🟡 |  | No tests for  repo-lookup paths or  multi-installation |
| 5 | 🟢 |  | Redundant  inside map |

The parallelisation via `Promise.allSettled` is sound and the dedup logic is correct. The `getAllGitHubAppConfigs` / constructor fallback chain is clean. Items 2–5 are polish; item 1 is a correctness bug that should be fixed before merge.

bencered (commented):
## PR #346 v2 Review — Multi-Installation GitHub App Support

All 40 tests pass ✅. The parallelisation approach and overall architecture are solid. A few issues remain before merge.

---

### 🔴 Token-scope bug in `generatePushAuth` when `getCachedInstallationToken` throws

**File:** `github-provider.ts` — `generatePushAuth`

The inner `catch {}` swallows **both** the `getInstallationRepository` error *and* any error thrown by the subsequent `getCachedInstallationToken` call:

```ts
for (const appConfig of this.allAppConfigs) {
  try {
    const repo = await getInstallationRepository(appConfig, ...);
    if (repo) {
      const token = await getCachedInstallationToken(appConfig); // ← if this throws...
      return { authType: "app", token };
    }
  } catch {
    // ...we silently fall through to the NEXT installation or the primary fallback
  }
}
```

**Scenario:** Two installations. The target repo lives in installation B. `getInstallationRepository(B)` succeeds (repo found), but `getCachedInstallationToken(B)` throws (e.g. RSA key wrong, rate-limited). The catch block silently falls through to the primary-installation fallback and returns a token for installation A — which does **not** have write access to the repo. The git push will fail with an opaque 403, hiding the real cause.

**Fix:** Split the try/catch so only the repository lookup is guarded:

```ts
for (const appConfig of this.allAppConfigs) {
  let repo: InstallationRepository | null = null;
  try {
    repo = await getInstallationRepository(appConfig, repoOwner, repoName, ...);
  } catch {
    continue; // repo lookup failed, try next installation
  }
  if (repo) {
    // Token errors should propagate — we know this is the right installation
    const token = await getCachedInstallationToken(appConfig);
    return { authType: "app", token };
  }
}
```

---

### 🟡 Silent fallback when repo is not found in any installation (`generatePushAuth`)

When `repoOwner`/`repoName` are provided but no installation claims the repo (all return null or throw), the code silently falls back to `allAppConfigs[0]`'s token without any warning. The caller receives a token that is almost certainly scoped to the wrong installation, and the subsequent git push will fail with a confusing error.

**Suggestion:** At minimum, add a `console.warn` before the fallback path:

```ts
// After the for-loop exits without returning
console.warn(
  `generatePushAuth: repo ${repoOwner}/${repoName} not found in any installation; falling back to primary`
);
```

Alternatively, throw here rather than silently proceeding, since a token known to lack access provides no benefit.

---

### 🟡 `listBranches` fails fast on non-404/403 errors — asymmetric with `listRepositories`

```ts
// listBranches
const status = extractHttpStatus(error);
if (status === 404 || status === 403) {
  lastError = error;
  continue;          // skip to next installation
}
throw SourceControlProviderError.fromFetchError(...); // ← throws immediately for 500, timeout, etc.
```

By contrast, `listRepositories` uses `Promise.allSettled` and tries all installations regardless of error type. A transient 500 or network timeout in the **first** installation means the second installation is never tried, even though it might have access to the repo.

Consider also continuing on `status === undefined` (network errors) and 5xx codes, consistent with how `listRepositories` behaves.

---

### 🟡 Missing test coverage for new `generatePushAuth` paths

The new repo-lookup flow in `generatePushAuth` has zero test coverage. Missing cases:

1. **Happy path**: `repoOwner`/`repoName` provided, repo found in second installation → token for second installation is returned (not primary).
2. **Repo not in any installation**: all `getInstallationRepository` calls return null → falls back to primary token.
3. **Bug case from 🔴 above**: `getInstallationRepository` succeeds for configB, but `getCachedInstallationToken(configB)` throws — current code silently returns wrong token; after fix, should propagate the error.
4. **`listBranches` multi-installation**: no tests for the fallback-on-404/403 iteration path.

Without test (1), it's impossible to tell from CI whether the repo-lookup feature actually routes to the correct installation.

---

### 🟢 Minor: redundant `.catch()` logging inside `listRepositories` map

```ts
this.allAppConfigs.map((appConfig) =>
  listInstallationRepositories(appConfig, ...).catch((error) => {
    console.warn(`Failed to list repos ...`);
    throw error; // re-throw so allSettled sees it as rejected
  })
)
```

The `.catch()` re-throws anyway, so its only effect is the `console.warn`. That's fine, but it makes the code look more complex than it is. Consider either logging in the `for (const result of results)` loop (where you already have access to `result.reason`) or leaving a comment explaining the intent.

---

### Summary

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | 🔴 | `generatePushAuth` | `catch {}` swallows token errors; wrong installation's token silently returned |
| 2 | 🟡 | `generatePushAuth` | Silent fallback when repo not found in any installation |
| 3 | 🟡 | `listBranches` | Non-404/403 errors abort iteration; inconsistent with `listRepositories` |
| 4 | 🟡 | `github-provider.test.ts` | No tests for `generatePushAuth` repo-lookup paths or `listBranches` multi-installation |
| 5 | 🟢 | `listRepositories` | Redundant `.catch()` inside map |

The parallelisation via `Promise.allSettled` is sound and the dedup logic is correct. The `getAllGitHubAppConfigs` / constructor fallback chain is clean. Items 2–5 are polish; item 1 is a correctness bug that should be fixed before merge.

ColeMurray:
### Multi-installation gap analysis

Reviewed the full codebase to understand the blast radius. The PR adds multi-installation parsing and routing on the TypeScript control plane, but several other layers still assume a single installation ID. Here's the complete picture.

---

#### 1. `pull-request-service.ts:105` — `generatePushAuth()` called without repo context

The only production call site for `generatePushAuth` passes no arguments, so it always falls back to `allAppConfigs[0]`. `session.repo_owner` / `session.repo_name` are available and used 21 lines later (lines 127–128). Pushes to repos on a secondary installation will get a token scoped to the wrong org → 403.

```ts
// Current (line 105)
pushAuth = await this.deps.sourceControlProvider.generatePushAuth();

// Fix
pushAuth = await this.deps.sourceControlProvider.generatePushAuth(session.repo_owner, session.repo_name);
```

---

#### 2. Modal/Python — token generation uses single `GITHUB_APP_INSTALLATION_ID`

The Python side reads `GITHUB_APP_INSTALLATION_ID` via `os.environ.get()` and passes it verbatim to `generate_installation_token()`, which interpolates it directly into the GitHub API URL:

```python
url = f"https://api.github.com/app/installations/{installation_id}/access_tokens"
```

**If the env var contains comma-separated IDs (e.g. `"111,222"`), this URL becomes malformed → 404 on every token request.** This breaks sandbox spawn, restore, and image builds — not just for secondary installations, but for ALL repos, since the primary installation's token can't be generated either.

Affected call sites (all identical pattern):
- `web_api.py:130` — sandbox spawn
- `web_api.py:511` — sandbox restore
- `functions.py:72` — legacy spawn path
- `scheduler/image_builder.py:129` — repo image builds + cron

The token failure is caught silently (`except Exception`), so sandboxes start without a clone token. Public repos would still clone; **private repos fail silently**.

**Fix options:**
- **(a)** Have the control plane resolve the correct `installation_id` per-repo and pass it to Modal as part of the create/restore request. Modal then uses that specific ID.
- **(b)** Parse comma-separated IDs on the Python side too (mirroring `getAllGitHubAppConfigs`), with the control plane telling Modal which one to use.
- **(c)** Pre-generate the token on the control plane and pass it to Modal directly. Less clean since tokens expire in 1 hour and would need refresh.

---

#### 3. GitHub bot — `handlers.ts:127`

```ts
installationId: env.GITHUB_APP_INSTALLATION_ID,
```

The github-bot reads the env var directly (not via `getGitHubAppConfig()`) and passes it as-is to `generateInstallationToken()`. With comma-separated IDs, this produces a malformed GitHub API URL → the bot can't generate tokens for any webhook handling (PR review assignments, @mention commands).

---

#### 4. Terraform — `variables.tf:87`

```hcl
variable "github_app_installation_id" { type = string }
```

The single string variable propagates unchanged to all three services (control-plane worker, github-bot worker, Modal secret). The TypeScript control plane now correctly parses comma-separated values, but Modal and github-bot do not. There's no separate variable or per-service splitting.

---

#### 5. Image builder cron — no per-repo installation routing

`scheduler/image_builder.py` rebuilds repo images on a cron schedule. It calls `_generate_clone_token()` which uses the single `GITHUB_APP_INSTALLATION_ID`. Even if the comma-separated parsing were fixed, there's no mechanism to route different repos to different installations during image builds — all repos must be cloneable with the first installation's token.

---

### Summary

| Layer | Issue | Severity |
|---|---|---|
| `pull-request-service.ts` | `generatePushAuth()` called without repo context | Bug — PR pushes to secondary-installation repos fail |
| Python/Modal (all token sites) | Comma-separated ID passed verbatim to GitHub API URL | **Breaking** — malformed URL, no tokens generated for any repo |
| `github-bot/handlers.ts` | Reads env var directly, not via parser | **Breaking** — bot webhook handling fails entirely |
| Terraform | Single string var, no per-service splitting | Config gap |
| Image builder cron | No per-repo installation routing | Feature gap |

The TypeScript control plane changes are well-structured. The core issue is that setting `GITHUB_APP_INSTALLATION_ID` to comma-separated values will break the Python/Modal and github-bot layers before multi-installation routing can even be exercised. These layers need corresponding changes to parse (or receive) the correct installation ID per-repo.

PR #357, opened 2026-03-15
chore: sandbox cleanup — auth consolidation, base_branch fix, Go models

bencered:
## Summary

Sandbox infrastructure cleanup and OpenCode Go model support.

### Modal Sandbox (`entrypoint.py`)
- **Auth consolidation**: `_setup_openai_oauth` now builds a unified `auth.json` supporting OpenAI OAuth, OpenCode Zen (pay-as-you-go), and OpenCode Go (subscription) providers
- **`base_branch` fix**: `or "main"` fallback instead of `dict.get("branch", "main")` — empty string no longer bypasses the default
- **`session_config.loaded`** log line for debugging session config keys
- **Logging**: renamed `message` → `error_message` in `supervisor.fatal` to avoid structlog field collision

### Modal Image (`base.py`)
- `copy=True` on `add_local_dir` — bakes sandbox code into image layer so `run_commands` can follow
- Sandbox version stamp (`.version` file)
- Cache buster → `v43`, writes to `/tmp/.cache-buster` instead of `/dev/null`

### Shared Models (`models.ts`)
- Added OpenCode Go models: `opencode-go/kimi-k2.5`, `opencode-go/minimax-m2.5`, `opencode-go/glm-5`
- New "OpenCode Go" category in `MODEL_OPTIONS`

## Note
`OPENCODE_ZEN_API_KEY` and `OPENCODE_GO_API_KEY` env vars must be added to the Modal secret spec separately for Zen/Go auth to activate in sandboxes. Without them, `auth.json` simply omits those providers.


packages/modal-infra/src/sandbox/entrypoint.py
```diff
@@ -291,25 +293,42 @@ def _install_tools(self, workdir: Path) -> None:
...
                 "expires": 0,
             }
-
             account_id = os.environ.get("OPENAI_OAUTH_ACCOUNT_ID")
             if account_id:
-                openai_entry["accountId"] = account_id
+                entry["accountId"] = account_id
+            auth_data["openai"] = entry
+
+        zen_key = os.environ.get("OPENCODE_ZEN_API_KEY")
```
ColeMurray, line 318: this should already be addressed via the RepoSecrets functionality within the system, where the control-plane will inject the env vars into the sandbox and OpenCode will pick them up.

the OAI ones exist as a special case due to the need of refreshing the subscription.

https://github.com/anomalyco/opencode/blob/1b86c27fb8d5b96e09cab48ad33ce08f381652cb/packages/opencode/src/provider/provider.ts#L952-L962

set with env: ["OPENCODE_API_KEY"]

PR #361, opened 2026-03-16
feat: add Daytona sandbox provider as self-hosted Modal alternative

### LINKED ISSUE #359: feat: Daytona sandbox provider (self-hosted alternative to Modal)
Currently the only sandbox backend is Modal, which means you need a Modal account and are locked into their pricing/infrastructure. For self-hosted or air-gapped deployments this is a non-starter.

This adds `packages/daytona-infra` — a FastAPI service that implements the same HTTP API surface as `modal-infra` but uses the [Daytona SDK](https://github.com/daytonaio/sdk) for container management and S3/MinIO for snapshot storage. The control plane picks the backend at runtime based on which env vars are set (`DAYTONA_API_URL` + `DAYTONA_API_SECRET` vs `MODAL_API_SECRET` + `MODAL_WORKSPACE`), so existing Modal deployments are unaffected.

## What's included

**`packages/daytona-infra`** (Python, FastAPI)
- Sandbox lifecycle: create, warm, snapshot (tarball → S3), restore from snapshot
- Repo image builds with async callback to control plane
- GitHub App token generation (works with both github.com and GHES)
- OpenCode supervisor entrypoint + WebSocket bridge (ported from modal-infra)
- Dockerfiles for both the API server and the sandbox base image

**`packages/control-plane`**
- `DaytonaClient` — HTTP client matching the `ModalClient` interface
- `DaytonaSandboxProvider` — implements `SandboxProvider` with proper error classification for circuit breaker
- `SessionDO` and repo-image routes select the provider based on env config
- `DAYTONA_API_URL` / `DAYTONA_API_SECRET` added to `Env` type

**Terraform**
- `daytona_api_url` and `daytona_api_secret` variables
- Wired into the control plane worker bindings

## Motivation

We needed this for a deployment where Modal wasn't an option (internal infrastructure, private network). The Daytona SDK gives us Docker-level control over sandbox containers while keeping the same API contract the control plane already expects. S3/MinIO replaces Modal's snapshot mechanism.

Happy to iterate on any of this. The API surface is intentionally identical to Modal's so nothing else in the stack needs to change.

raghavpillai:
## Summary

Closes #359

- Adds `packages/daytona-infra` — a FastAPI backend that implements the same sandbox HTTP API as `modal-infra` using the Daytona SDK for container management and S3/MinIO for snapshots
- Adds `DaytonaClient` and `DaytonaSandboxProvider` to the control plane, matching the existing `ModalClient` / `ModalSandboxProvider` interfaces
- `SessionDO` and repo-image routes pick the provider at runtime: set `DAYTONA_API_URL` + `DAYTONA_API_SECRET` for Daytona, otherwise falls back to Modal
- Terraform variables for `daytona_api_url` and `daytona_api_secret`, wired into control plane worker bindings

### `packages/daytona-infra`

- **Sandbox lifecycle**: create, warm, snapshot (workspace tarball to S3), restore from snapshot
- **Repo image builds**: async build with callback to control plane on completion
- **Auth**: GitHub App token generation, HMAC-signed internal tokens
- **Supervisor**: OpenCode server + WebSocket bridge process management with crash recovery (ported from `modal-infra`)
- **Dockerfiles**: `Dockerfile` (API server) and `Dockerfile.sandbox` (dev environment base image with Node 22, Python 3.12, OpenCode, Playwright)

### Control plane

- `DaytonaClient` (`sandbox/daytona-client.ts`) — HTTP client with HMAC auth, request logging, error classification
- `DaytonaSandboxProvider` (`sandbox/providers/daytona-provider.ts`) — `SandboxProvider` implementation with transient/permanent error classification for circuit breaker
- Provider selection in `SessionDO.createLifecycleManager()` and `repo-images.ts`

No changes to existing Modal behavior — if `DAYTONA_API_URL` is unset, everything works exactly as before.

### Review feedback addressed

- Renamed `modal_object_id` to `provider_object_id` in web API responses and updated DaytonaClient to read the new key
- Implemented actual S3 delete in `/api/delete-provider-image`
- GHES proxy: removed HMAC auth requirement (OAuth paths need to forward GitHub tokens), forwards Authorization header, uses `GHES_CA_BUNDLE` env var for TLS verification, path allowlist (`login/oauth/`, `api/v3/`)
- Error responses now return HTTP 500 via JSONResponse instead of HTTP 200 with `success: false`
- Replaced `subprocess.run` with `asyncio.create_subprocess_exec` in `manager.py`
- DaytonaClient now throws `DaytonaApiError` on `success: false` responses instead of returning error objects


packages/daytona-infra/src/auth/github_app.py
```diff
@@ -0,0 +1,115 @@
...
+    Returns:
+        API base URL without trailing slash.
+    """
+    host = (hostname or os.environ.get("GITHUB_HOSTNAME", "github.com")).lower().rstrip("/")
+    if host == "github.com":
+        return "https://api.github.com"
+    return f"https://{host}/api/v3"
+
+
+def generate_jwt(app_id: str, private_key: str) -> str:
```
ColeMurray: we should likely introduce a shared sandbox-infra package to avoid duplicating this in the N many providers

packages/daytona-infra/src/registry/models.py
```diff
@@ -0,0 +1,62 @@
...
+class SnapshotStatus(StrEnum):
+    """Status of a snapshot."""
+
+    BUILDING = "building"
+    READY = "ready"
+    FAILED = "failed"
+    EXPIRED = "expired"
+
+
+class Repository(BaseModel):
```
ColeMurray: ditto with this, it appears duplicated from the modal infra

ColeMurray (changes requested):
rebase with sandbox-runtime and remove duplicates

ColeMurray:
this is now available, resolving

PR #362, opened 2026-03-16
feat: GitHub Enterprise Server (GHES) support

### LINKED ISSUE #360: feat: GitHub Enterprise Server (GHES) support
All GitHub URLs in the codebase are hardcoded to github.com. If you're running a GHES instance, nothing works — OAuth, App auth, API calls, git clone URLs, noreply emails, avatar URLs, PR creation links, all of it.

This replaces every hardcoded `github.com` / `api.github.com` reference with a configurable hostname. Set `GITHUB_HOSTNAME=github.example.com` and the rest follows: API base becomes `https://github.example.com/api/v3`, noreply domain becomes `users.noreply.github.example.com`, etc.

## Scope

The change touches basically everything that talks to GitHub:

- **control-plane**: OAuth token exchange, GitHub App installation tokens, source control provider (PR creation, branch listing, repo access checks, git push URLs), participant avatar URLs
- **github-bot**: installation token generation, permission checks, reaction API calls
- **modal-infra / daytona-infra**: git clone/push URLs, noreply email fallback, image builder `ls-remote`
- **web**: NextAuth GitHub provider config — split browser-side (user's browser hits GHES directly for OAuth) vs server-side (Worker routes through tunnel proxy for token exchange + API)
- **terraform**: `github_hostname`, `ghes_tunnel_url`, and `web_app_custom_url` variables threaded through to workers

## GHES in a private VPC

GHES instances are usually not on the public internet, which means Cloudflare Workers can't reach them directly. The `GHES_TUNNEL_URL` env var points to a reverse proxy (we used a Cloudflare Tunnel, but any tunnel that forwards to the GHES instance works). The web app sends the user's browser directly to GHES for OAuth authorize, then routes the server-side token exchange through the tunnel.

The `daytona-infra` package also includes a `/ghes-proxy/{path}` endpoint that forwards requests to GHES for the same reason — the infra API server runs inside the private network and can reach GHES directly.

## Design

A single `resolveGitHubUrls(hostname?)` function derives all URL variants from the hostname. Every call site that previously hardcoded `github.com` or `api.github.com` now calls this (or receives the resolved URLs via config). Defaults to github.com when `GITHUB_HOSTNAME` is unset, so existing deployments are unaffected.

Depends on #359 (Daytona provider) for the `daytona-infra` pieces, but the control-plane/web/github-bot/modal-infra changes work independently.

raghavpillai:
## Summary

Closes #360.

Replaces all hardcoded `github.com` / `api.github.com` URLs with a hostname derived from `GITHUB_HOSTNAME`. Set it to your GHES hostname and everything else adjusts — API base, OAuth endpoints, git remote URLs, noreply emails, avatar URLs.

This PR is standalone and based on `main` — no dependency on the Daytona provider PR.

### Changes

**`packages/control-plane`**
- New `github-urls.ts` — `resolveGitHubUrls(hostname?)` returns `{ apiBase, webBase, gitHost, noreplyDomain, isEnterprise }` from a single hostname
- `github-app.ts`: `apiBase` threaded through installation token generation and repo listing
- `github.ts`: OAuth token exchange, user info, email fetch, noreply generation all use resolved URLs
- `github-provider.ts`: PR creation, branch listing, repo access, git push spec, label/reviewer endpoints
- `participant-service.ts`: avatar URL uses configured hostname
- `shared.ts` (`createRouteSourceControlProvider`): resolves URLs and applies `GHES_TUNNEL_URL` override for private VPC setups
- `types.ts`: `GITHUB_HOSTNAME` and `GHES_TUNNEL_URL` added to `Env`

**`packages/github-bot`**
- `github-auth.ts`: `resolveApiBase()` exported for installation tokens and permission checks
- `handlers.ts`: imports `resolveApiBase` from `github-auth`, reaction API URLs use resolved base, `apiBase` flows through caller gating
- `types.ts`: `GITHUB_HOSTNAME` added to `Env`

**`packages/linear-bot`**
- `types.ts`: `GITHUB_HOSTNAME` added to `Env`
- `webhook-handler.ts`: uses `env.GITHUB_HOSTNAME` for resolved URLs

**`packages/modal-infra`**
- `github_app.py`: `resolve_api_base()` for token generation
- `bridge.py`: noreply email fallback uses `GITHUB_HOSTNAME`
- `manager.py`: `VCS_HOST` reads `GITHUB_HOSTNAME` instead of hardcoding `github.com`
- `image_builder.py`: `git ls-remote` URLs use resolved host

**`packages/web`**
- `auth.ts`: NextAuth GitHub provider splits browser-side URLs (user hits GHES directly for OAuth) from server-side URLs (Worker goes through `GHES_TUNNEL_URL` for token exchange and API calls)

**Terraform**
- `github_hostname`, `ghes_tunnel_url`, `web_app_custom_url` variables
- Wired into control plane, github-bot, and linear-bot worker bindings
- `locals.tf`: `web_app_url` respects `web_app_custom_url` override

### Private VPC pattern

GHES is usually not publicly accessible. `GHES_TUNNEL_URL` points to a reverse proxy (Cloudflare Tunnel, bore, etc.) that the control plane uses for server-side GitHub API calls. The user's browser still goes directly to GHES for OAuth authorize (since the user is on the private network).


packages/control-plane/src/routes/shared.ts
```diff
@@ -68,11 +69,17 @@ export function error(message: string, status = 400): Response {
...
   const appConfig = getGitHubAppConfig(env);
   const provider = resolveScmProviderFromEnv(env.SCM_PROVIDER);
+  // When GHES is behind a private VPC, override apiBase to route through the tunnel proxy
+  const baseUrls = resolveGitHubUrls(env.GITHUB_HOSTNAME);
+  const urls = env.GHES_TUNNEL_URL
+    ? { ...baseUrls, apiBase: `${env.GHES_TUNNEL_URL.replace(/\/+$/, "")}/api/v3` }
+    : baseUrls;
   return createSourceControlProvider({
     provider,
     github: {
```
ColeMurray, line 79: [Automated Review] **Suggestion: Consolidate tunnel URL rewriting logic.**

This `GHES_TUNNEL_URL` → `/api/v3` rewriting pattern is duplicated here and in `getGitHubAppConfig()` (`github-app.ts`). Both independently resolve the same tunnel-aware API base.

This is fragile — if the tunnel URL format changes, both sites need updating. Since `getGitHubAppConfig()` already bakes `apiBase` into the config (including tunnel logic), this second derivation here is redundant for the `appConfig` path.

Consider either:
1. Having `resolveGitHubUrls()` accept an optional `tunnelUrl` parameter so there's a single derivation site, or
2. Extracting a small helper like `resolveGitHubApiBase(hostname?, tunnelUrl?)` that both sites call.

packages/control-plane/src/auth/github-app.ts
```diff
@@ -620,14 +632,23 @@ export function getGitHubAppConfig(env: {
...
+  GHES_TUNNEL_URL?: string;
 }): GitHubAppConfig | null {
   if (!isGitHubAppConfigured(env)) {
     return null;
   }
 
+  const urls = resolveGitHubUrls(env.GITHUB_HOSTNAME);
+  // When GHES is behind a private VPC, use the tunnel proxy for API calls
+  const apiBase = env.GHES_TUNNEL_URL
+    ? `${env.GHES_TUNNEL_URL.replace(/\/+$/, "")}/api/v3`
```
ColeMurray, line 645: [Automated Review] **Suggestion: Guard tunnel rewrite with `isEnterprise`.**

If `GHES_TUNNEL_URL` is set but `GITHUB_HOSTNAME` is absent (or `"github.com"`), this code still appends `/api/v3` to the tunnel URL. That would be incorrect for github.com, where the API lives at `api.github.com`, not `github.com/api/v3`.

This is likely a misconfiguration scenario, but a defensive guard would prevent subtle breakage:
```typescript
const apiBase = urls.isEnterprise && env.GHES_TUNNEL_URL
  ? `${env.GHES_TUNNEL_URL.replace(/\/+$/, "")}/api/v3`
  : urls.apiBase;
```

Also see comment on `shared.ts` — this same rewriting logic is duplicated there.

packages/control-plane/src/github-urls.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return {
+    hostname: host,
+    apiBase: isEnterprise ? `https://${host}/api/v3` : "https://api.github.com",
+    webBase: `https://${host}`,
+    gitHost: host,
+    noreplyDomain: isEnterprise ? `users.noreply.${host}` : "users.noreply.github.com",
+    isEnterprise,
+  };
+}
```
ColeMurray, line 56: [Automated Review] **Suggestion: Add unit tests for `resolveGitHubUrls()`.**

This function is the foundation of the entire GHES feature, but it has no dedicated unit tests. Given the branching logic (enterprise vs. github.com, noreply domain format, API base derivation), a small test file would be valuable. Suggested test cases:

- `undefined` / `""` / `"github.com"` → standard github.com URLs
- `"github.example.com"` → GHES URLs with `/api/v3`, `isEnterprise: true`
- `"GITHUB.EXAMPLE.COM/"` → case normalization + trailing slash stripping
- Verify all returned fields (`apiBase`, `webBase`, `gitHost`, `noreplyDomain`, `isEnterprise`)

packages/github-bot/src/github-auth.ts
```diff
@@ -2,6 +2,14 @@ export interface GitHubAppConfig {
   appId: string;
   privateKey: string;
   installationId: string;
+  /** GitHub REST API base URL. Defaults to https://api.github.com. */
+  apiBase?: string;
+}
+
+/** Resolve the GitHub API base URL from a hostname (supports GHES). */
+export function resolveApiBase(hostname?: string): string {
+  const host = (hostname || "github.com").toLowerCase().replace(/\/+$/, "");
```
ColeMurray, line 11: [Automated Review] **Nit: Use `resolveApiBase()` as fallback in private functions.**

`getInstallationToken()` (line 79) and `checkSenderPermission()` (line 119) use `apiBase ?? "https://api.github.com"` as their fallback instead of calling the `resolveApiBase()` function defined right here. If a future caller omits `apiBase`, the fallback silently targets github.com regardless of `GITHUB_HOSTNAME`.

Suggested:
```typescript
const base = apiBase ?? resolveApiBase();
```

Note: `resolveApiBase()` doesn't read the env in the github-bot (unlike the Python version), so this would still default to github.com — but it centralizes the logic.

packages/control-plane/src/session/participant-service.ts
```diff
@@ -59,10 +59,14 @@ export interface ParticipantServiceDeps {
...
 export function getAvatarUrl(
   login: string | null | undefined,
-  provider: SourceControlProviderName = "github"
+  provider: SourceControlProviderName = "github",
+  githubHostname?: string
 ): string | undefined {
   if (!login) return undefined;
-  if (provider === "github") return `https://github.com/${login}.png`;
+  if (provider === "github") {
+    const host = (githubHostname || "github.com").toLowerCase().replace(/\/+$/, "");
```
ColeMurray, line 67: [Automated Review] **Nit: Reuse `resolveGitHubUrls()` instead of inline normalization.**

The hostname normalization logic here (`(githubHostname || "github.com").toLowerCase().replace(...)`) duplicates what `resolveGitHubUrls()` already does. Using:
```typescript
const { hostname } = resolveGitHubUrls(githubHostname);
return `https://${hostname}/${login}.png`;
```
would be more consistent with the rest of this PR's approach.

packages/modal-infra/src/sandbox/bridge.py
```diff
@@ -33,9 +33,15 @@
 
 configure_logging()
 
+
 # Fallback git identity when prompt author has no SCM name/email configured.
 # Matches the co-author trailer used in generateCommitMessage (shared/git.ts).
-FALLBACK_GIT_USER = GitUser(name="OpenInspect", email="open-inspect@noreply.github.com")
+def _fallback_noreply_domain() -> str:
+    host = os.environ.get("GITHUB_HOSTNAME", "github.com").lower().rstrip("/")
+    return f"noreply.{host}" if host != "github.com" else "noreply.github.com"
```
ColeMurray, line 41: [Automated Review] **Nit: Clarify the noreply domain distinction.**

The GHES noreply domain here uses `noreply.{host}` while the TypeScript side uses `users.noreply.{host}`. This is intentional (bot fallback identity vs. user attribution), but a brief inline comment like `# No 'users.' prefix — this is the bot's fallback identity, not a user noreply` would help future readers understand the distinction.

ColeMurray (commented):
## [Automated Review] PR #362 — feat: GitHub Enterprise Server (GHES) support

**Files changed**: 27 | +335 / -63

### Summary

Replaces all hardcoded `github.com` / `api.github.com` URLs across the stack with a hostname-configurable abstraction backed by `GITHUB_HOSTNAME`. A new `resolveGitHubUrls()` in the control plane derives API base, web base, git host, and noreply domain from a single hostname. Per-package equivalents handle github-bot and Python modal-infra. Terraform wires the new `github_hostname` and `ghes_tunnel_url` variables to all workers. The web auth layer splits browser-side (user-reachable GHES) from server-side (tunnel-proxied) endpoints. Well-structured change that methodically touches all necessary layers.

**Note**: The diff includes unrelated changes merged from main (command palette #367, next.js bump #366). This review focuses only on the GHES commits.

### Positive Feedback

1. **Clean central abstraction** — `resolveGitHubUrls()` returns a well-typed `GitHubUrls` interface that makes it impossible to miss a dimension (API, web, git host, noreply). Single-hostname-in, all-URLs-out is elegant.
2. **Smart browser/server split in web auth** — Separation of browser-side URLs (user hits GHES for OAuth redirect) from server-side URLs (tunnel-proxied for token exchange) correctly handles private VPC topology.
3. **Thorough Terraform wiring** — All worker types plus the Cloudflare web worker get the new variables. The `web_app_custom_url` override is a nice touch.

See inline comments for specific feedback.

ColeMurray:
## [Automated Review] Architectural Assessment

The inline comments addressed localized code quality. This follow-up evaluates whether the PR's overall design is the right approach.

---

### 1. `resolveGitHubUrls()` is in the wrong package

This is the central issue. The function lives in `packages/control-plane/src/github-urls.ts`, but 5 packages need the same logic:

| Package | What it does instead |
|---------|---------------------|
| `control-plane` | Uses `resolveGitHubUrls()` (the canonical one) |
| `github-bot` | Rolls its own `resolveApiBase()` — same logic, different interface |
| `linear-bot` | Inlines `(env.GITHUB_HOSTNAME \|\| "github.com").toLowerCase().replace(...)` |
| `web` | Inlines the same hostname normalization + ternary for API base |
| `modal-infra` (Python) | Has its own `resolve_api_base()` (Python can't share TS — expected) |

The repo already has `@open-inspect/shared` specifically for cross-package types and utilities — every TS package depends on it. `resolveGitHubUrls()` belongs there. Instead, the PR creates **4 independent TypeScript implementations** of "hostname → GitHub URLs."

### 2. Hostname normalization is copy-pasted 7 times

The exact pattern `(hostname || "github.com").toLowerCase().replace(/\/+$/, "")` (or the Python equivalent) appears in **7 locations**. Four of those are TypeScript packages sharing a common dependency graph designed for exactly this kind of reuse:

- `control-plane/src/github-urls.ts`
- `control-plane/src/session/participant-service.ts`
- `github-bot/src/github-auth.ts`
- `linear-bot/src/webhook-handler.ts`
- `modal-infra/src/auth/github_app.py`
- `modal-infra/src/sandbox/bridge.py`
- `modal-infra/src/scheduler/image_builder.py`

### 3. Tunnel URL rewriting logic is triplicated

The `GHES_TUNNEL_URL` → `tunnelUrl + "/api/v3"` rewrite appears in 3 places:

- `github-app.ts:644-645`
- `routes/shared.ts:74-75`
- `web/src/lib/auth.ts:41-42`

If the path convention ever changes, three files across two packages need updating. This should be a single function call — `resolveGitHubUrls()` could accept an optional `tunnelUrl` parameter to centralize this.

### 4. No hostname input validation

None of the 7 normalization sites validate that the input is actually a hostname. If someone configures `GITHUB_HOSTNAME=https://github.example.com` (with protocol — a plausible mistake), the result is `https://https://github.example.com/api/v3`. The function should strip a protocol prefix or throw a clear error.

### 5. `GITHUB_API_BASE` constant is dead code

The PR marks it `@deprecated` but no consumer references it in the PR branch. Dead exports should be removed, not annotated.

### 6. Missing tests

`resolveGitHubUrls()` is the single most important function in this PR — every API call, OAuth flow, git push, and avatar URL depends on it — and it has zero tests. The only test changes add `apiBase: "https://api.github.com"` to existing assertions. No test verifies the GHES path actually works.

### 7. `getAvatarUrl()` ignores the PR's own abstraction

`participant-service.ts:67` inlines hostname normalization instead of calling `resolveGitHubUrls()`, which is defined in the same package. If the author's own code doesn't use the abstraction, the abstraction's coverage is incomplete.

---

### Recommended changes

1. **Move `resolveGitHubUrls()` to `@open-inspect/shared`**, with an optional `tunnelUrl` parameter to handle tunnel rewriting in one place
2. **Delete `resolveApiBase()` from github-bot** — import from shared
3. **Replace inline hostname normalization** in linear-bot, web, and participant-service with the shared function
4. **Add hostname validation** (strip or reject protocol prefixes)
5. **Add unit tests** for the core function (GHES vs github.com, edge cases, tunnel URL)
6. **Remove the dead `GITHUB_API_BASE` export**

This would consolidate 4 independent TS implementations into 1 shared function (Python duplication is expected and acceptable).

### Verdict

The feature objective is sound, the browser/server URL split for web auth is well-thought-out, and the Terraform wiring is thorough. But the core abstraction lives in the wrong package, and the resulting duplication across 4 TS packages undermines the value of having `@open-inspect/shared`. Request changes before merge.

ColeMurray (changes requested):

ColeMurray:
closing due to inactivity

PR #373, opened 2026-03-18
feat: improve syntax highlighting with rehype-highlight

leejayhsu:
This improves the syntax highlighting for code blocks and prose (single backtick text). 

In the future it might be nice for the highlighting theme to be a user setting.

Screenshot 2026-03-18 at 12 48 46 PM

### DIFF
--- a/packages/web/public/hljs-themes/atom-one-dark.css
+++ b/packages/web/public/hljs-themes/atom-one-dark.css
@@ -1,10 +1,10 @@
 pre code.hljs {
   display: block;
   overflow-x: auto;
-  padding: 1em
+  padding: 1em;
 }
 code.hljs {
-  padding: 3px 5px
+  padding: 3px 5px;
 }
 /*
 
@@ -27,34 +27,34 @@ hue-6-2: #e6c07b
 */
 .hljs {
   color: #abb2bf;
-  background: #282c34
+  background: #282c34;
 }
 .hljs-comment,
 .hljs-quote {
   color: #5c6370;
-  font-style: italic
+  font-style: italic;
 }
 .hljs-doctag,
 .hljs-keyword,
 .hljs-formula {
-  color: #c678dd
+  color: #c678dd;
 }
 .hljs-section,
 .hljs-name,
 .hljs-selector-tag,
 .hljs-deletion,
 .hljs-subst {
-  color: #e06c75
+  color: #e06c75;
 }
 .hljs-literal {
-  color: #56b6c2
+  color: #56b6c2;
 }
 .hljs-string,
 .hljs-regexp,
 .hljs-addition,
 .hljs-attribute,
 .hljs-meta .hljs-string {
-  color: #98c379
+  color: #98c379;
 }
 .hljs-attr,
 .hljs-variable,
@@ -64,7 +64,7 @@ hue-6-2: #e6c07b
 .hljs-selector-attr,
 .hljs-selector-pseudo,
 .hljs-number {
-  color: #d19a66
+  color: #d19a66;
 }
 .hljs-symbol,
 .hljs-bullet,
@@ -72,19 +72,19 @@ hue-6-2: #e6c07b
 .hljs-meta,
 .hljs-selector-id,
 .hljs-title {
-  color: #61aeee
+  color: #61aeee;
 }
 .hljs-built_in,
 .hljs-title.class_,
 .hljs-class .hljs-title {
-  color: #e6c07b
+  color: #e6c07b;
 }
 .hljs-emphasis {
-  font-style: italic
+  font-style: italic;
 }
 .hljs-strong {
-  font-weight: bold
+  font-weight: bold;
 }
 .hljs-link {
-  text-decoration: underline
+  text-decoration: underline;
 }
--- a/packages/web/src/app/globals.css
+++ b/packages/web/src/app/globals.css
@@ -8,6 +8,12 @@
   content: none;
 }
 
+/* Let <pre>'s bg-card be authoritative — avoid color frame from hljs theme background */
+pre code.hljs {
+  background: transparent;
+  padding: 0;
+}
+
 /*
  * Design System Color Tokens
  *
--- a/packages/web/src/components/syntax-highlight-theme.tsx
+++ b/packages/web/src/components/syntax-highlight-theme.tsx
@@ -25,9 +25,7 @@ export function SyntaxHighlightTheme() {
     const href = HLJS_THEMES[resolvedTheme ?? "light"] ?? HLJS_THEMES.light;
 
     // Check if this theme is already loaded
-    const existing = document.querySelector(
-      'link[data-hljs-theme]',
-    ) as HTMLLinkElement | null;
+    const existing = document.querySelector("link[data-hljs-theme]") as HTMLLinkElement | null;
     if (existing?.getAttribute("href") === href) return;
 
     const link = document.createElement("link");
@@ -40,9 +38,12 @@ export function SyntaxHighlightTheme() {
     document.head.appendChild(link);
 
     // Fallback removal if onload doesn't fire (e.g. cached)
-    if (existing) {
-      setTimeout(() => existing.remove(), 100);
-    }
+    const timer = existing ? setTimeout(() => existing.remove(), 100) : undefined;
+
+    return () => {
+      clearTimeout(timer);
+      link.remove();
+    };
   }, [resolvedTheme]);
 
   return null;
--- a/packages/web/public/hljs-themes/atom-one-light.css
+++ b/packages/web/public/hljs-themes/atom-one-light.css
@@ -1,10 +1,10 @@
 pre code.hljs {
   display: block;
   overflow-x: auto;
-  padding: 1em
+  padding: 1em;
 }
 code.hljs {
-  padding: 3px 5px
+  padding: 3px 5px;
 }
 /*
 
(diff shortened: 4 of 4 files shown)


packages/web/src/components/safe-markdown.tsx
```diff
@@ -1,10 +1,13 @@
 "use client";
 
 import ReactMarkdown from "react-markdown";
+import rehypeHighlight from "rehype-highlight";
 import rehypeSanitize, { defaultSchema } from "rehype-sanitize";
 import remarkGfm from "remark-gfm";
 import type { ComponentPropsWithoutRef } from "react";
 
+import "highlight.js/styles/atom-one-dark.css";
```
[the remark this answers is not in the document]
leejayhsu: i feel like this suggestion would be better suited when there is proper light/dark theme support. and also maybe user selectable theme would address this. But these are things that I would prefer to tackle in a separate PR.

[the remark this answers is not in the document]
ColeMurray: @leejayhsu the existing UI does respond to the users system preferences for light/dark mode. it would be ideal to have this mirror that behavior

image
image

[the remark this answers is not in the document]
leejayhsu: oops i didn't realize there was already dark/light mode. ok i will address this 👍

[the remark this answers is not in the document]
leejayhsu: I took a stab at this in https://github.com/ColeMurray/background-agents/pull/373/commits/f0135ed6828fa1feb993e73503ad442eb4ae58ef

I think vendorizing the theme css might seem a bit weird but my intention was to make this easier to build a user selectable theme picker later (user preferred light theme and user preferred dark theme).

does this implementation seem reasonable?

packages/web/public/hljs-themes/atom-one-dark.css
```diff
@@ -0,0 +1,90 @@
...
+hue-4:   #98c379
+hue-5:   #e06c75
+hue-5-2: #be5046
+hue-6:   #d19a66
+hue-6-2: #e6c07b
+
+*/
+.hljs {
+  color: #abb2bf;
+  background: #282c34
```
ColeMurray: [Automated Review] **Suggestion [Visual]:** The `.hljs` background here (`#282c34`) conflicts with the `<pre>` element's `bg-card` background + `p-3` padding in `safe-markdown.tsx`. Since `code.hljs` is `display: block`, it fills the content box but not the padding box — so the `pre`'s 12px padding area shows `bg-card` (~`#1d1d1d` in dark mode) creating a visible color frame around the differently-colored code block. Padding also stacks to 28px (12px pre + 16px code).

Simplest fix — add to `globals.css`:
```css
pre code.hljs { background: transparent; }
```
This lets the `pre`'s `bg-card` be authoritative for the background.

packages/web/src/components/syntax-highlight-theme.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+/**
+ * Dynamically loads the appropriate highlight.js theme stylesheet based on
+ * the current color scheme. Cleanly swaps stylesheets when the theme changes.
+ */
+export function SyntaxHighlightTheme() {
+  const { resolvedTheme } = useTheme();
+
+  useEffect(() => {
```
ColeMurray, line 24: [Automated Review] **Suggestion [Correctness]:** This effect manipulates the DOM but doesn't return a cleanup function. On a rapid theme triple-toggle (dark→light→dark), the third invocation early-returns because `querySelector` finds the original dark `<link>` and the href matches. But the second invocation's `onload` callback then removes that dark link — leaving only the light stylesheet active while `resolvedTheme` is dark, with no re-run since the dependency hasn't changed.

Extremely unlikely in practice, but adding cleanup makes this robust:
```tsx
return () => {
  clearTimeout(timer);
  link.remove();
};
```

packages/web/src/components/safe-markdown.tsx
```diff
@@ -1,10 +1,13 @@
 "use client";
 
 import ReactMarkdown from "react-markdown";
+import rehypeHighlight from "rehype-highlight";
```
ColeMurray, line 4: [Automated Review] **Suggestion [Performance]:** `rehype-highlight` bundles lowlight's common languages (~37 grammars). Not blocking, but if bundle size is a concern, the `languages` option can restrict to only the languages that actually appear in practice. Worth measuring with `next build --analyze`.

packages/web/src/components/syntax-highlight-theme.tsx
```diff
@@ -0,0 +1,49 @@
...
+const HLJS_THEMES: Record<string, string> = {
+  light: "/hljs-themes/atom-one-light.css",
+  dark: "/hljs-themes/atom-one-dark.css",
+};
+
+/**
+ * Dynamically loads the appropriate highlight.js theme stylesheet based on
+ * the current color scheme. Cleanly swaps stylesheets when the theme changes.
+ */
+export function SyntaxHighlightTheme() {
```
ColeMurray, line 21: [Automated Review] **Suggestion [Maintainability]:** This component is rendered per `SafeMarkdown` instance. Currently there's only one usage, but if `SafeMarkdown` gets reused (e.g., in a message list), N instances will independently run the same DOM-manipulation effect. The `existing` check prevents duplicate stylesheets so it's not broken — just worth considering lifting to a layout or provider level if reuse grows.

packages/web/public/hljs-themes/atom-one-dark.css
```diff
@@ -0,0 +1,90 @@
...
+}
+.hljs-emphasis {
+  font-style: italic
+}
+.hljs-strong {
+  font-weight: bold
+}
+.hljs-link {
+  text-decoration: underline
+}
```
ColeMurray: [Automated Review] **Nit:** Missing trailing newline at end of file (same for `atom-one-light.css`).

packages/web/src/components/safe-markdown.tsx
```diff
@@ -63,9 +66,10 @@ interface SafeMarkdownProps {
 export function SafeMarkdown({ content, className = "" }: SafeMarkdownProps) {
   return (
     <div className={`prose prose-sm dark:prose-invert max-w-none break-words ${className}`}>
+      <SyntaxHighlightTheme />
       <ReactMarkdown
         remarkPlugins={[remarkGfm]}
-        rehypePlugins={[[rehypeSanitize, sanitizeSchema]]}
+        rehypePlugins={[rehypeHighlight, [rehypeSanitize, sanitizeSchema]]}
```
ColeMurray, line 72: [Automated Review] **Positive:** Good plugin ordering — `rehypeHighlight` before `rehypeSanitize` ensures all generated HTML is sanitized before reaching the DOM. The existing allowlist for `className` on `code`/`span` is exactly what rehype-highlight emits, so no new XSS surface is introduced.

packages/web/src/components/syntax-highlight-theme.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useTheme } from "next-themes";
+import { useEffect } from "react";
+
+/**
+ * Maps color scheme to highlight.js theme CSS files served from /hljs-themes/.
+ *
+ * To support user-configurable themes, extend this mapping or accept
+ * theme names as props/from user settings.
+ */
+const HLJS_THEMES: Record<string, string> = {
```
ColeMurray, line 12: [Automated Review] **Positive:** Nice design — dynamic stylesheet loading from `public/` with `resolvedTheme` avoids bundling both themes and sets up a clean path toward user-configurable themes.

packages/web/src/app/globals.css
```diff
@@ -2,6 +2,12 @@
 @tailwind components;
 @tailwind utilities;
 
+/* Remove backtick decorations from prose inline code */
+.prose :where(code):not(:where([class~="not-prose"], [class~="not-prose"] *))::before,
```
ColeMurray, line 6: [Automated Review] **Positive:** Clean approach — uses Tailwind Typography's own `:where()` specificity pattern to override the backtick decorations without needing `!important`.

ColeMurray (commented):
[Automated Review] Overall solid PR — correct plugin ordering, clean theme-switching design, and no new XSS surface. See inline comments for suggestions and nitpicks.

packages/web/public/hljs-themes/atom-one-dark.css
ColeMurray: Done — added `pre code.hljs { background: transparent; padding: 0; }` to `globals.css`. Lets `<pre>`'s `bg-card` control the background and avoids the doubled padding. Addressed in 58a7323.

packages/web/src/components/syntax-highlight-theme.tsx
ColeMurray: Done — added cleanup that clears the fallback timer and removes the link element. Prevents the stale-callback race on rapid toggling. Addressed in 58a7323.

packages/web/src/components/safe-markdown.tsx
ColeMurray: Acknowledged — the common language bundle is reasonable for now since agent output spans many languages. Can revisit with the `languages` option and `next build --analyze` if bundle size becomes a concern.

packages/web/src/components/syntax-highlight-theme.tsx
ColeMurray: Noted — `SafeMarkdown` is rendered per-message in the event timeline, so there will be multiple instances. The `existing` check deduplicates the stylesheet so it's functionally harmless. Fine to leave as-is for now.

packages/web/public/hljs-themes/atom-one-dark.css
ColeMurray: Fixed — added trailing newlines and ran prettier on both theme files. Addressed in 58a7323.

ColeMurray (approved):

PR #377, opened 2026-03-19
fix(web): allowlist control plane query params

ColeMurray:
## Summary
- add a shared web helper that rebuilds forwarded control-plane query strings from an explicit allowlist of `status`, `limit`, and `offset`
- update the sessions, automations, and automation runs API routes to use the allowlisted query builder instead of passing through the full client query string
- add unit coverage for allowed param forwarding, dropped unknown params, and empty allowed values

## Testing
- `npm test -w @open-inspect/web -- control-plane-query.test.ts`

---
*Created with [Open-Inspect](https://open-inspect-prod.vercel.app/session/25834e07f9b89de9a9fbd5952a18e875)*


ColeMurray:
@open-inspect[bot] please re-review

PR #395, opened 2026-03-20
fix: bootstrap sandbox runtime for Modal deploy

Hirshey:
## Summary
- bootstrap `packages/modal-infra/deploy.py` so `modal deploy deploy.py` works from a clean monorepo checkout and imports the full `src` package during deploy
- resolve `sandbox_runtime` in `src/images/base.py` from the importable package in deployed Modal functions, with a monorepo fallback for local development
- update Terraform's Modal deploy setup and local docs to install/use both sibling Python packages explicitly, and add regression tests for the deploy/runtime path

## Root cause
PR #369 extracted `packages/sandbox-runtime`, but the Modal deploy/image path still depended on `sandbox_runtime` already being importable without explicitly bootstrapping that sibling package in a clean environment. In production this surfaced as sandbox startup failures like:

```py
ModuleNotFoundError: No module named 'sandbox_runtime'
```

That left sessions stuck at `Sandbox: connecting` because the sandbox process died before the first bridge connection.

## What changed
- `packages/modal-infra/deploy.py`
  - prepends the local monorepo package paths for `modal-infra` and `sandbox-runtime`
  - validates the expected runtime files exist up front
  - imports `src` rather than only `src.app` so all Modal functions/endpoints register during deploy
- `packages/modal-infra/src/images/base.py`
  - resolves the runtime directory from the importable `sandbox_runtime` package in deployed Modal functions
  - falls back to the sibling monorepo path for local dev/image build contexts
  - fails early if expected runtime files are missing
- `.github/workflows/terraform.yml`
  - installs `packages/sandbox-runtime` and `packages/modal-infra` as editable packages before Modal deploy
- docs/tests
  - document the sibling package requirement for direct Modal work
  - add deploy/runtime smoke coverage and update Modal sandbox test mocks to match the current async Modal API

### Local checks
- `npm install`
- `npm run lint`
- `npm run typecheck`
- `npm test`
- `pytest packages/modal-infra/tests -q`
- `pytest packages/sandbox-runtime/tests -q`

### Manual deploy validation on fork
- deployed from this branch with `modal deploy deploy.py`
- created a new session on my forked deployment and confirmed it moved past `connecting`, reached `sandbox.ready`, and completed a prompt successfully
- confirmed Modal logs showed `bridge.connect` success instead of `ModuleNotFoundError`
- triggered a fresh repo-image build after the fix and confirmed it completed successfully

## Notes
- Python `mypy` still reports pre-existing errors in untouched files (`entrypoint.py`, `manager.py`, `web_api.py`, `functions.py`); this branch removes the new `base.py` type issue it introduced but does not attempt to clean up unrelated baseline typing debt.

### DIFF
--- a/packages/modal-infra/src/images/base.py
+++ b/packages/modal-infra/src/images/base.py
@@ -29,7 +29,9 @@ def _resolve_sandbox_runtime_dir() -> Path:
         if module_file is None:
             raise RuntimeError("sandbox_runtime package is missing a __file__ path")
         runtime_dir = Path(module_file).resolve().parent
-    except ModuleNotFoundError:
+    except ModuleNotFoundError as exc:
+        if exc.name != "sandbox_runtime":
+            raise
         runtime_dir = (
             Path(__file__).resolve().parents[3] / "sandbox-runtime" / "src" / "sandbox_runtime"
         )
--- a/packages/modal-infra/tests/test_deploy.py
+++ b/packages/modal-infra/tests/test_deploy.py
@@ -15,7 +15,12 @@ def _load_deploy_module():
     assert spec.loader is not None
 
     module = module_from_spec(spec)
-    spec.loader.exec_module(module)
+    sys.modules["modal_infra_deploy_test"] = module
+    try:
+        spec.loader.exec_module(module)
+    except Exception:
+        sys.modules.pop("modal_infra_deploy_test", None)
+        raise
     return module
 
 


.github/workflows/terraform.yml
```diff
@@ -285,8 +285,9 @@ jobs:
 
       - name: Install Modal CLI and dependencies
         run: |
+          python -m pip install --upgrade pip
           pip install -e packages/sandbox-runtime
-          pip install modal fastapi pydantic httpx PyJWT
+          pip install -e packages/modal-infra
```
ColeMurray, line 290: this change makes sense and can be followed-up in another PR, but doesn't seem related to the core issue described

packages/modal-infra/src/images/base.py
```diff
@@ -10,14 +10,43 @@
 - Sandbox entrypoint and bridge code
 """
 
+from importlib import import_module
 from pathlib import Path
 
 import modal
 
-import sandbox_runtime
```
ColeMurray, line 17: it's not clear why we are removing this, as we definitely need to import the sandbox_runtime. it should be importable and error otherwise

packages/modal-infra/tests/test_build_sandbox.py
```diff
@@ -10,7 +11,7 @@
 def _fake_sandbox_create(captured):
     """Return a fake Sandbox.create that captures kwargs."""
 
-    def fake_create(*args, **kwargs):
+    async def fake_create_aio(*args, **kwargs):
```
ColeMurray, line 14: these should be fixed, but not related to this PR

packages/modal-infra/deploy.py
```diff
@@ -2,20 +2,69 @@
...
+def bootstrap_local_deploy_paths() -> Path:
+    """
+    Bootstrap local monorepo package paths for deployment.
+
+    Modal deploy runs from `packages/modal-infra/`, but the extracted
+    `sandbox_runtime` package lives in a sibling package and is not published to
+    PyPI. Ensure both local packages are importable before importing `src`.
+    """
+    _prepend_sys_path(MODAL_INFRA_ROOT)
+    _prepend_sys_path(SANDBOX_RUNTIME_SRC)
```
ColeMurray, line 35: i don't think we want to do this, as the package should otherwise be installed prior to running this, rather than modifying the sys path such that it can be imported

Hirshey:
Closing in favor of #399 . I reproduced the same sandbox_runtime failure from a clean environment and validated a workaround, but #399 fixes it at the proper dependency/setup layer, which is the better solution for this repo.

PR #403, opened 2026-03-24
fix: correct Terraform worker deployment and CI setup docs

pnwatin:
## Summary
- fix Cloudflare worker Terraform ordering so cron triggers wait for the worker deployment to exist
- remove the control-plane migration settings that caused `terraform apply` to fail when following the Getting Started deployment flow
- update `docs/GETTING_STARTED.md` with the correct GitHub Actions secret names, add `DEPLOYMENT_NAME`, and document safe upload of `GH_APP_PRIVATE_KEY`

## Why
Following `docs/GETTING_STARTED.md` exposed two deployment issues in fork-based setups. First, Terraform could fail during control-plane deployment because the configured migration fields did not match the expected initial deployment path. Second, the CI/CD docs listed secret names that did not match the workflow, which led to incorrect worker URLs and missing GitHub App configuration in deployed environments.

### DIFF
--- a/terraform/environments/production/workers-control-plane.tf
+++ b/terraform/environments/production/workers-control-plane.tf
@@ -85,9 +85,9 @@ module "control_plane_worker" {
 
   compatibility_date  = "2024-09-23"
   compatibility_flags = ["nodejs_compat"]
-  migration_tag       = "v1"
-  migration_old_tag   = null
-  new_sqlite_classes  = []
+  migration_tag       = var.control_plane_migration_tag
+  migration_old_tag   = var.control_plane_migration_old_tag
+  new_sqlite_classes  = var.control_plane_new_sqlite_classes
 
   cron_triggers = ["* * * * *"]
 
--- a/terraform/environments/production/terraform.tfvars.example
+++ b/terraform/environments/production/terraform.tfvars.example
@@ -155,6 +155,12 @@ project_root = "../../../"
 enable_durable_object_bindings = false
 enable_service_bindings = false
 
+# Optional: override these only when an existing deployment needs to re-enter
+# Phase 1 for a later Durable Object migration. Fresh deployments can omit them.
+# control_plane_migration_tag = "v1"
+# control_plane_migration_old_tag = null
+# control_plane_new_sqlite_classes = []
+
 # =============================================================================
 # Access Control
 # =============================================================================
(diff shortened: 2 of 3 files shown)


terraform/environments/production/workers-control-plane.tf
```diff
@@ -85,9 +85,9 @@ module "control_plane_worker" {
 
   compatibility_date  = "2024-09-23"
   compatibility_flags = ["nodejs_compat"]
-  migration_tag       = "v2"
-  migration_old_tag   = "v1"
-  new_sqlite_classes  = ["SchedulerDO"]
+  migration_tag       = "v1"
+  migration_old_tag   = null
+  new_sqlite_classes  = []
```
ColeMurray: [Automated Review] **Suggestion (non-blocking)**: These fresh-deploy defaults fix the immediate first-deploy problem, but they create a latent issue for existing deployments. If an existing deployment at migration tag v2 needs to re-enter Phase 1 (e.g., to add a new DO class), these values would attempt to regress from v2→v1 and re-declare existing classes as new.

Recommend lifting these three settings into Terraform variables with fresh-deploy defaults. Existing deployments override in their gitignored `terraform.tfvars`. No workflow changes needed — the CI defaults `enable_durable_object_bindings` to `true`, so the migration block evaluates to `null` during normal runs.

```hcl
# variables.tf
variable "control_plane_migration_tag" {
  default = "v1"
}
variable "control_plane_migration_old_tag" {
  default = null
}
variable "control_plane_new_sqlite_classes" {
  type    = list(string)
  default = []
}

# workers-control-plane.tf
migration_tag       = var.control_plane_migration_tag
migration_old_tag   = var.control_plane_migration_old_tag
new_sqlite_classes  = var.control_plane_new_sqlite_classes
```

docs/GETTING_STARTED.md
```diff
@@ -592,6 +592,7 @@ Go to your fork's Settings → Secrets and variables → Actions, and add:
 | `CLOUDFLARE_API_TOKEN`        | Your Cloudflare API token                                                     |
 | `CLOUDFLARE_ACCOUNT_ID`       | Your Cloudflare account ID                                                    |
 | `CLOUDFLARE_WORKER_SUBDOMAIN` | Your workers.dev subdomain                                                    |
+| `DEPLOYMENT_NAME`             | Your deployment name                              |
```
ColeMurray: [Automated Review] **Nit**: This row has shorter padding than the others, breaking source-level column alignment.

PR #410, opened 2026-03-25
feat(slack): add file attachment support

duboff:
## Summary

- Users can attach files (images, documents) to Slack messages and they are forwarded to the coding agent
- Supports up to 3 files per message, up to 5MB each
- Images (mimetype starts with `image/`) are typed as `image`, other files as `file`
- Files too large or that fail to download are skipped gracefully
- Forwarded/quoted messages (Slack `attachments`) are extracted and included in prompt context
- Adds `downloadSlackFile` utility to `slack-client.ts` for authenticated Slack file downloads

## Changes

- `SlackFile` and `FileAttachment` interfaces for type safety
- `downloadAndEncodeSlackFiles()` — downloads files via Slack API, base64 encodes, classifies by mimetype
- `formatAttachments()` — extracts text from quoted/forwarded Slack messages for prompt context
- `sendPrompt` / `startSessionAndSendPrompt` / `handleIncomingMessage` — thread `fileAttachments` through to the control plane
- `handleAppMention` / `handleDirectMessage` — download files after receiving event, pass to shared handler
- `handleRepoSelection` — retrieve stored `fileAttachments` from KV when user picks a repo after clarification

## Test plan

- [ ] Send a message with an image attachment — verify it reaches the agent
- [ ] Send a message with a document (PDF, text) — verify it reaches the agent
- [ ] Send >3 files in one message — verify only first 3 are processed
- [ ] Send a file >5MB — verify it is skipped with no error
- [ ] Forward/quote a message to the bot — verify quoted text appears in prompt context
- [ ] Attach files in a DM — verify they are forwarded

## Summary by CodeRabbit

* **New Features**
  * File attachments: Attach up to 3 files per message (5 MB each); non-downloadable or oversized files are ignored and logged.
  * Attachments included in bot interactions: Uploaded files sent with your messages are now passed into bot prompts.
  * Quoted message context: Forwarded/quoted message text is automatically included as context in bot replies.
  * Enhanced user profiles: User email addresses are now captured and available.

### DIFF
--- a/packages/slack-bot/src/index.ts
+++ b/packages/slack-bot/src/index.ts
@@ -24,6 +24,7 @@ import {
   getThreadMessages,
   publishView,
   openView,
+  downloadSlackFile,
 } from "./utils/slack-client";
 import { resolveUserNames } from "./utils/resolve-users";
 import { createClassifier } from "./classifier";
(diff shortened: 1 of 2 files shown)


ColeMurray (changes requested):
took a review through this. from my understanding of the change, we're downloading the attachments in the slackbot and then base64 encoding them into the prompt. 

While it would work, this seems a bit odd to me and likely has quality performance implications. We're not making use of the LLM's vision capabilities for images, and passing base64 for text/doc files when it would likely be better to either pass the text raw, or let the agent handle the file download.

onboard with supporting attachments, but we likely want to take a different approach here.

ColeMurray:
resolving due to inactivity

PR #436, opened 2026-03-31
feat(slack): respond to thread replies without @mention

LarsBuur:
## Summary
- When a user replies in a channel thread where the bot already has an active session, process the message as a follow-up without requiring an @mention
- The first message still needs `@mention` to start a session, but subsequent thread replies are handled automatically

## Changes
- `packages/slack-bot/src/index.ts`: handle `message` events in threads with existing sessions via `lookupThreadSession`

## Use case
Currently users must `@mention` the bot on every single message in a thread, even in a 1:1 conversation. This makes the interaction feel unnatural. With this change, once a session is started via `@mention`, follow-up replies in the same thread are processed automatically.

## Test plan
- [ ] `@mention` the bot in a channel to start a session
- [ ] Reply in the thread without `@mention`
- [ ] Verify the bot processes the reply and responds
- [ ] Verify messages in threads without an active session are still ignored

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured app-mention events are handled early to prevent unintended subsequent processing.
  * Added support for threading: messages in thread conversations are now detected, tied to existing thread sessions, cleaned of duplicate mention content, and forwarded with thread context preserved.


ColeMurray (changes requested):
see comment

ColeMurray:
closing due to inactivity

PR #452, opened 2026-04-06
fix: hydrate PR artifacts on subscribe and keep session branch live

Quantumlyy:
This fixes the session view around PR creation, reconnects, and refreshes.

Before this change, a PR could appear from the live websocket event but disappear after reload because the client did not rebuild its artifact state from the subscribe payload.
The active session branch could also stay stale until a refresh.

This change makes the websocket subscribe payload carry persisted artifacts, updates live artifact events to use the same full artifact shape, and broadcasts branch updates when PR creation changes the session branch.
On the web side, the session hook now treats `subscribed.artifacts` as the source of truth for reconnects, replaces stale local artifact state with that snapshot, and still avoids duplicate artifacts when a live `artifact_created` event arrives for something already hydrated.

I also trimmed the web follow-up so the socket hook is mostly transport and state wiring.
The only metadata-key fallback now lives where it is actually needed in the sidebar PR metadata display.

Checks run:
- `npm run test:integration -w @open-inspect/control-plane`

## Summary by CodeRabbit

* **New Features**
  * Subscriptions now include full session artifacts (rich metadata and numeric timestamps) and a new session-branch message updates clients in real time.
* **Bug Fixes**
  * Branch names are validated/sanitized during PR creation; invalid names return HTTP 400. Session-branch broadcasts are emitted in the correct order and artifact broadcasts use a unified artifact shape with duplicate-safe handling.
* **Tests**
  * Expanded unit and integration tests covering artifacts, branch broadcasts, ordering, sanitization, and related edge cases.


packages/web/src/hooks/use-session-socket.ts
```diff
@@ -109,18 +110,16 @@ function toUiSandboxEvent(event: SharedSandboxEvent): SandboxEvent {
...
   return {
     id: artifact.id,
     type: artifact.type as Artifact["type"],
     url: artifact.url,
-    createdAt: Date.now(),
-    metadata: artifact.prNumber ? { prNumber: artifact.prNumber } : undefined,
+    createdAt: artifact.createdAt,
+    metadata:
+      artifact.metadata && typeof artifact.metadata === "object"
+        ? (artifact.metadata as Artifact["metadata"])
```
ColeMurray: [Automated Review] **Critical — Unsafe `as` cast breaks metadata key mapping.**

Before this PR, the server sent `{ prNumber: prResult.id }` as a top-level field, and the old `toUiArtifact` explicitly mapped it:
```ts
metadata: artifact.prNumber ? { prNumber: artifact.prNumber } : undefined
```

This PR replaces that with a raw cast of the DB metadata shape (`{ number, state, head, base }`) directly to the web `Artifact["metadata"]` type, which expects `{ prNumber, prState, head, base }`. The keys `number`/`state` do not match `prNumber`/`prState`. At runtime, any code reading `artifact.metadata?.prNumber` gets `undefined`.

Today `metadata-section.tsx` has the fallback (`number ?? prNumber`), and `action-bar.tsx` only reads `url` — so there is no runtime bug *yet*. But the previous explicit mapping was correct and this PR silently removes it.

**Suggested fix** (pick one):
1. Map keys explicitly here: `prNumber: (metadata as any).number, prState: (metadata as any).state`
2. Update the web `Artifact` type to use `number`/`state` as canonical keys
3. Have the server broadcast with `prNumber`/`prState` keys to match the existing web type (simplest — avoids the backward-compat dance in `metadata-section.tsx` entirely)

packages/web/src/components/action-bar.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+        sessionId="session-1"
+        sessionStatus="active"
+        artifacts={[
+          {
+            id: "artifact-pr-1",
+            type: "pr",
+            url: "https://github.com/acme/web-app/pull/42",
+            metadata: {
+              prNumber: 42,
+              prState: "open",
```
ColeMurray, line 28: [Automated Review] **Suggestion — Test metadata does not match the wire format.**

This test passes `{ prNumber: 42, prState: "open" }`, but after this PR the server sends `{ number: 42, state: "open" }`. The test only passes because `ActionBar` reads `url`, not metadata keys.

Compare with `use-session-socket.test.tsx` which correctly uses `{ number, state }`. The test data here should match what the server actually produces, otherwise it gives false confidence if someone later adds metadata reads to `ActionBar`.

packages/control-plane/src/session/pull-request-service.ts
```diff
@@ -143,21 +139,33 @@ export class SessionPullRequestService {
...
+      const pushResult = await this.deps.pushBranchToRemote(sanitizedHeadBranch, pushSpec);
       if (!pushResult.success) {
         return { kind: "error", status: 500, error: pushResult.error };
       }
 
-      this.deps.repository.updateSessionBranch(session.id, headBranch);
+      if (session.branch_name !== sanitizedHeadBranch) {
+        this.deps.repository.updateSessionBranch(session.id, sanitizedHeadBranch);
+      }
+      this.deps.broadcastSessionBranch(sanitizedHeadBranch);
```
ColeMurray, line 170: [Automated Review] **Suggestion — Consider adding a comment explaining why the broadcast is unconditional.**

The DB write is correctly guarded (`if (session.branch_name !== sanitizedHeadBranch)`), but the broadcast always fires — even when the branch hasn't changed. The test "skips branch writes when the sanitized branch is unchanged but still broadcasts" confirms this is intentional, but a brief comment here explaining *why* (eventual consistency for clients that missed a prior message? idempotent for reconnects?) would help future readers.

packages/control-plane/src/session/pull-request-service.test.ts
```diff
@@ -230,8 +274,80 @@ describe("SessionPullRequestService", () => {
...
+        number: 42,
+        state: "open",
+        head: "open-inspect/session-name-1",
+        base: "main",
+      },
+      createdAt: expect.any(Number),
+    });
+  });
+
+  it("syncs the branch after push and before PR creation", async () => {
```
ColeMurray, line 291: [Automated Review] **Nit** — `resolveHeadBranchForPrSpy.mockRestore()` would be safer in an `afterEach` block or via `vi.restoreAllMocks()` to avoid leaking the spy if the test throws before reaching this line.

ColeMurray (dismissed):
## [Automated Review]

**PR**: #452 — fix: hydrate PR artifacts on subscribe and keep session branch live
**Author**: @Quantumlyy (external contributor)
**Files changed**: 10 | +654 / -42

### Summary

Fixes a real bug where PR artifacts appeared from live WebSocket events but vanished on reconnect/refresh. Adds `session_branch` broadcast so client-side branch stays current. The approach is sound, but there is one blocking issue around metadata key naming that needs resolution before merging.

### Positive Feedback

1. **Thorough test coverage.** 306 lines of hook tests with `invocationCallOrder` ordering assertions and integration-level SQLite insertion tests. Well above average for an external contribution.
2. **The `sanitizeBranchName` guard in the service is genuinely necessary.** It covers a real gap in `resolveHeadBranchForPr`'s fallback path (`branch-resolution.ts:78-81`) which returns unsanitized values. Good defensive coding.
3. **Correct reconnect semantics.** Replacing `setArtifacts([])` with `setArtifacts(data.artifacts.map(toUiArtifact))` solves both hydration and stale-data-on-reconnect in one line.

### Verdict

**Request Changes.** One blocking issue: the metadata key mismatch (`number`/`state` vs `prNumber`/`prState`) introduced by removing the explicit key mapping in `toUiArtifact`. The `metadata-section.tsx` fallback patches one call site, but the root cause — an unsafe `as` cast bridging mismatched schemas — should be resolved before merging.

ColeMurray (approved):

PR #485, opened 2026-04-14
fix: improve Cloudflare bootstrap deploy flow

sidpalas:
This PR fixes (or clarifies the docs to avoid user error) for some issues I ran into when performing a fresh setup.

### Changes:
- add fresh-install migration guidance to docs/GETTING_STARTED.md
- add enable_control_plane_cron in the production Terraform env
- add enable_cron_triggers to the worker module and gate cron resource creation on it
- add tfvars notes for first-time Slack enablement and bootstrap flags

### Note:
- keeps the existing checked-in control-plane migration values, but adds comments explaining what fresh installs should use

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for fresh control‑plane deployments and clarified Phase 1/Phase 2 instructions, including activation steps and a note to enable scheduled execution in Phase 2.
  * Clarified Modal CLI setup/usage and expanded Slack deployment guidance for retrying applies when workers aren’t yet present.

* **New Features**
  * Added an option to control creation and activation of scheduled (cron) execution for the control‑plane, allowing it to be disabled during initial bootstrap and enabled afterward.


sidpalas:
Ah, I had an older copy of the repo I was working from and https://github.com/ColeMurray/background-agents/pull/403 already addressed the migration tagging issues.

I think there is still a chicken and egg problem with the control plane and cron trigger, but users can get around that by omitting the trigger array on first install.

PR #537, opened 2026-04-22
feat: add fetchUser GraphQL query to Linear client

ColeMurray:
## Summary

- Adds `fetchUser(client, userId)` to `linear-client.ts` — queries Linear's `user(id:)` GraphQL endpoint to resolve name and email
- Follows the same pattern as `fetchIssueDetails`: try/catch with error logging, returns `null` on failure
- Includes 4 tests: success with email, null email, user not found, API error

**Pre-step 7** of the unified user model migration ([pre-steps doc](https://github.com/ColeMurray/background-agents/blob/main/docs/internal/2026-04-21-unified-user-model-pre-steps.md#pre-step-7-add-a-fetchuser-graphql-query-to-the-linear-client)). The main migration will call `fetchUser(client, webhook.appUserId)` before session creation to resolve identity for linking — having the query function already in place keeps that PR focused on wiring.

## Test plan

- [x] `npm run typecheck -w @open-inspect/linear-bot` passes
- [x] `npm test -w @open-inspect/linear-bot` — all 94 tests pass (90 existing + 4 new)

## Summary by CodeRabbit

* **New Features**
  * Added user retrieval from Linear with normalized {id, name, email|null} responses and safe error handling that returns null on failure.

* **Tests**
  * Added thorough tests for user retrieval covering success, null fields, missing user, network errors, and GraphQL errors; tests isolate and restore global fetch between cases.

### DIFF
--- a/packages/linear-bot/src/utils/linear-client.test.ts
+++ b/packages/linear-bot/src/utils/linear-client.test.ts
@@ -72,4 +72,14 @@ describe("fetchUser", () => {
     const result = await fetchUser(client, "user-1");
     expect(result).toBeNull();
   });
+
+  it("returns null on GraphQL errors payload", async () => {
+    mockFetchResponse({
+      data: null,
+      errors: [{ message: "Not authorized" }],
+    });
+
+    const result = await fetchUser(client, "user-1");
+    expect(result).toBeNull();
+  });
 });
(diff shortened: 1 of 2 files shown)


ColeMurray:
Good catch — `linearGraphQL` only checked HTTP status, so GraphQL-level errors (auth, schema, etc. returning `200 OK` with `{ errors: [...] }`) silently fell through for all callers, not just `fetchUser`. Fixed in `linearGraphQL` itself so every caller benefits: it now throws when the response contains an `errors` array. Added a test for the `{ errors: [...] }` case.

Addressed in e304bf28.

PR #546, opened 2026-04-22
fix: use derived worker URLs in Terraform outputs

Zeyuzhao:
## Summary
- derive the control plane and linear bot URLs in Terraform locals
- reuse those derived URLs in outputs and verification commands

## Testing
- Not run

## Summary by CodeRabbit

* **Chores**
  * Updated production infrastructure configuration to source control plane and Linear bot URLs from local configuration values instead of worker module outputs.
  * Verification commands now reference updated URL configurations.

### DIFF
--- a/terraform/environments/production/locals.tf
+++ b/terraform/environments/production/locals.tf
@@ -7,6 +7,7 @@ locals {
   control_plane_host = "open-inspect-control-plane-${local.name_suffix}.${var.cloudflare_worker_subdomain}.workers.dev"
   control_plane_url  = "https://${local.control_plane_host}"
   ws_url             = "wss://${local.control_plane_host}"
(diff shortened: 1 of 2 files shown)


MartinRoberts-Fountain:
In my opinion, it's best to grab the outputs from the source of truth instead of using the local.

ColeMurray:
@Zeyuzhao thanks for the contribution! can you clarify the intent of the change? the modules should already directly provide the exact URL. unclear on why we'd want to use the local instead

Zeyuzhao:
Sorry, I think this was an accidental pull request! Closing this PR.

Zeyuzhao:
For context, I was fixing a bug where the links were missing my username:

https://open-inspect-control-plane-cloudagent.zeyuzhao.workers.dev was displayed as https://open-inspect-control-plane-cloudagent.workers.dev

and the agent provided an edit that refactored the code in this way. I split the large commit into separate PRs, and accidentally submitted this PR upstream. 

I believe the issue has been fixed since then.

PR #559, opened 2026-04-25
fix: Set the repository image branch to the repository's default branch and add a delete button to delete built images

kartikye:
## Summary
  - resolve repo-image builds against the repository's actual default branch instead of assuming `main`
  - add a delete path for stored repo images and expose it in the settings UI
  - add unit and integration coverage for the new repo-image store and route behavior

  ## Why
  Repo image handling was too rigid for repositories whose default branch is not `main`, and there was no direct way to
  clear stored images from the UI. This change makes build triggering branch-aware and adds an explicit cleanup action
  for stored images.

  ## Impact
  Users can now:
  - trigger repo-image builds against the correct default branch for a repo
  - delete stored repo images from the pre-built images settings screen without affecting active builds

Happy to remove the delete button but if needed.

## Summary by CodeRabbit

* **New Features**
  * Delete pre-built repository images from the settings UI (per-repo delete button).
  * API endpoint to delete stored repo images; deletion reports how many images were removed.
  * Build triggering now uses each repository's configured default branch.

* **Bug Fixes**
  * Images in "building" status are not returned or deleted.


packages/control-plane/src/routes/repo-images.ts
```diff
@@ -266,7 +270,7 @@ async function handleTriggerBuild(
       {
         repoOwner: owner,
         repoName: name,
-        defaultBranch: "main",
+        defaultBranch: resolved.defaultBranch,
```
ColeMurray, line 273: [Automated Review] The manual trigger now registers and builds images for `resolved.defaultBranch`, but the Modal scheduler still computes `remote_sha` with `_git_ls_remote_sha(repo_owner, repo_name, "main", clone_token)` before deciding whether to call this endpoint. For repositories whose default branch is not `main`, scheduled rebuilds will either be skipped when `main` is absent or repeatedly triggered against the wrong SHA when both branches exist. Can we propagate the resolved/default branch into the scheduler comparison before triggering builds?

ColeMurray (changes requested):

ColeMurray:
closing due to stale

PR #566, opened 2026-04-27
fix: use derived worker URLs in terraform outputs and activate venv for Modal deploy

IlyaFilippov:
## Overview

The `cloudflare-worker` module's `worker_url` output omits the account subdomain, producing incorrect URLs like `open-inspect-control-plane-cannaflow.workers.dev` instead of the correct `open-inspect-control-plane-cannaflow.aged-cloud-e34a.workers.dev`. This PR replaces all references to `module.*.worker_url` in outputs with the correct locals that already include the subdomain.

This is the same bug reported in upstream PRs ColeMurray/background-agents#541 and ColeMurray/background-agents#546, which were closed without merging. Those PRs took the same approach (use locals instead of module outputs). The repo owner suggested the module should provide the correct URL directly, but the root cause — the module not having access to the account subdomain — was never fixed.

We chose to fix this at the outputs level rather than modifying the `cloudflare-worker` module because all functional URLs (env vars passed to workers, Modal, web app) already use `local.control_plane_url` and are correct. The module's `worker_url` output is only used in display outputs and verification commands — nowhere functional. Adding a `worker_subdomain` variable to the module would require changes to all 4 worker module calls for no functional benefit.

Additionally, the Modal deploy script now activates the Python venv if present, ensuring `sandbox_runtime` and other local dependencies are available when running `modal deploy` via Terraform.

## Changes

- **`outputs.tf`**: Replace `module.control_plane_worker.worker_url` and `module.linear_bot_worker[0].worker_url` references with `local.control_plane_url` and `local.linear_bot_url`
- **`locals.tf`**: Add `linear_bot_url` local (mirrors existing `control_plane_url` pattern)
- **`variables.tf`**: Change `r2_media_location` default from `ENAM` to `WEUR`
- **`deploy.sh`**: Activate `.venv` before running `modal deploy` if a venv exists

## Side effects and further ideas

- The upstream `cloudflare-worker` module still has a broken `worker_url` output — could be fixed by passing `worker_subdomain` into the module, but that's a larger change best done upstream
- No functional behavior changes — only terraform output display and Modal deploy reliability

## Checklist

- [ ] Tests (terraform plan verified locally)
- [x] Linear task in title — N/A
- [ ] DB migration — N/A

## Summary by CodeRabbit

* **Chores**
  * Refactored infrastructure configuration to consolidate service URL definitions into centralized configuration values.
  * Updated default media storage location from Eastern North America to Western Europe region for deployments without explicit location preferences.


IlyaFilippov:
Sorry, opened in wrong repository.
Will be happy to contribute in future weeks once I'm familiar with the entire codebase, gonna deploy it in all of my organizations.

PR #572, opened 2026-04-28
Add GitHub Automations Feature

MartinRoberts-Fountain:
image

Unlocks the GitHub automation

## Summary by CodeRabbit

* **New Features**
  * GitHub trigger integration (PRs, issues, comments, check suites) with selectable event types, normalization, and forwarding to the control plane.
  * Human-readable GitHub event context blocks for automations.

* **Enhancements**
  * UI: event-type selector, enabled GitHub trigger card, and GitHub-specific instruction placeholder.
  * New automation condition types: branch, label, path patterns, actor, check conclusion.
  * Session archiving UI and archiveSession helper; sidebar archive flow.

* **Tests**
  * Added tests for GitHub normalization, identity parsing, tokens, and queue behavior.


MartinRoberts-Fountain:
Pushed a commit with the suggestions implemented.

packages/modal-infra/src/registry/models.py
```diff
@@ -0,0 +1,62 @@
+"""Data models for repository and snapshot registry."""
```
ColeMurray: this can likely be deleted and seems unrelated to the PR

packages/shared/src/triggers/github/normalizer.ts
```diff
@@ -0,0 +1,289 @@
+/**
```
ColeMurray, line 1: it would be desirable to improve the typing here as there is a lot of Record<string,unknown> happening.
i believe we can use `@octokit/webhooks-types` for this where applicable

packages/shared/src/triggers/github/context.ts
```diff
@@ -0,0 +1,251 @@
+/**
```
ColeMurray, line 1: in the github bot and some of the other bots, we wrap the context with a <user_context> tag to help protect against prompt injection. hardly fool-proof, but probably worth while adding here for protection + consistency

packages/shared/src/triggers/github/normalizer.ts
```diff
@@ -0,0 +1,289 @@
...
+  if (!comment) return null;
+
+  const commentId = comment.id;
+  if (typeof commentId !== "number" || !Number.isFinite(commentId)) return null;
+
+  const issueNumber = issue?.number;
+  if (typeof issueNumber !== "number" || !Number.isFinite(issueNumber)) return null;
+
+  const triggerKey = `issue_comment:${commentId}`;
+  const concurrencyKey = `issue:${issueNumber}`;
```
ColeMurray: note, this uses issue as a concurrency key. if there is multiple comments that come in, they'll all compete.
not sure if intended

ColeMurray (dismissed):
comments + file removal

ColeMurray (approved):

ColeMurray:
thanks for the contribution!

PR #584, opened 2026-05-02
refactor: extract encryptTokenPair utility

ColeMurray:
## Summary

- Extracts a shared `encryptTokenPair(accessToken, refreshToken, encryptionKey)` utility in `auth/crypto.ts` that encrypts an access/refresh token pair, returning `null` for unprovided tokens and throwing on encryption failure
- Replaces 3 inline try/catch encryption blocks in `router.ts` (`handleCreateSession` and `handleSessionWsToken`) with calls to the shared utility
- Callers decide error policy: `handleCreateSession` catches and returns 500, `handleSessionWsToken` catches and continues with null tokens

## Test plan

- [x] 5 new unit tests for `encryptTokenPair` covering: both undefined, access-only, refresh-only, both provided (with round-trip decryption verification), and invalid key (throws)
- [x] All 1039 existing control-plane tests pass
- [x] Typecheck passes
- [x] Lint passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified encryption of access and refresh tokens into a single token-pair flow and tightened error handling so encryption failures now surface as errors instead of being partially ignored.

* **Tests**
  * Added tests covering token-pair encryption, missing-token cases, successful encrypt/decrypt, and failure/error scenarios.

### DIFF
--- a/packages/control-plane/src/auth/crypto.test.ts
+++ b/packages/control-plane/src/auth/crypto.test.ts
@@ -1,6 +1,13 @@
 import { describe, it, expect } from "vitest";
 import { timingSafeEqual } from "@open-inspect/shared";
-import { encryptToken, decryptToken, generateEncryptionKey, generateId, hashToken } from "./crypto";
+import {
+  encryptToken,
+  encryptTokenPair,
+  decryptToken,
+  generateEncryptionKey,
+  generateId,
+  hashToken,
+} from "./crypto";
 
 describe("crypto", () => {
   describe("generateEncryptionKey", () => {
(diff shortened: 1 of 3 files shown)


ColeMurray:
Good catch on the behavior change. We reviewed this and are comfortable with it — `TOKEN_ENCRYPTION_KEY` has been present in all deployments for a long time, and an encryption failure at this point would indicate a misconfiguration or a broken key rather than a transient issue worth degrading through. If `encryptToken` fails for the refresh token, it would almost certainly fail for the access token too (same key, same algorithm), so the graceful degradation path was effectively dead code.

Keeping both fatal simplifies the contract: `encryptTokenPair` either succeeds or throws, and callers get a single error policy decision rather than per-token semantics.

ColeMurray:
Agreed — updated in c127b10a. `handleSessionWsToken` now returns 500 on encryption failure instead of silently continuing with null tokens, matching `handleCreateSession`'s behavior.

The "overwrite valid tokens" concern doesn't actually apply here (the DO's ws-token handler checks `clientSentAnyToken` and skips the update when both are null), but returning nulls and silently degrading is still the wrong behavior when the encryption key is present and broken.

Issue #557, opened 2026-04-24
Session filters by user


Bnowako:
Problem:
When multiple people colaborate on open-inspect it's easy to loose your sessions. 

Solution:
Allow to filter sessions by user. 

By default show current user sessions, and allow to toggle 'all' / filter by user.

Bnowako:
@ColeMurray I think it would be nicer if we would show in my sessions, all sessions the user participated in, but this info is in DO. Wdyt about adding a table which would be an index of user-session that we can easily retrieve there?

If you sent 1 message to a session it will be shown as "My sessions"

Krishnachaitanyakc:
hey @ColeMurray, @Bnowako. I am interested in working on this, I did notice that both prior PRs (#599, #600) went creator only and closed silently. want to confirm the design before opening another one. @Bnowako's proposal on this issue is participation based and I think that's the right call.

**Proposed design**

1. New D1 table `session_participants(user_id, session_id, scm_login, first_seen_at)`. Composite PK `(user_id, session_id)`, index on `(user_id, first_seen_at DESC)`. Backfilled from existing `sessions` rows where `user_id IS NOT NULL`.

2. "Participation" means any of: created the session, explicitly added via `POST /participants`, or sent ≥1 prompt (WS or API). The prompt case is the load bearing one per @Bnowako's wording. Writes are idempotent (`INSERT OR IGNORE`) from `initialize.ts` and the prompt enqueue paths in `message-queue.ts`. All keyed on canonical `users.id`, resolved via `UserStore.getIdentity` for both bot `provider:id` prefixes and raw web ids. Read only, never auto creates.

3. API: `GET /sessions?mineScmUserId=<id>&mineProvider=<github|slack|linear>` plus optional `?scmLogin=` fallback for pre migration rows where `sessions.user_id IS NULL`. Control plane resolves `(provider, providerUserId)` to canonical `users.id` and JOINs `session_participants`. Unresolved identity returns an empty page. 400s on empty values or asymmetric params.

4. BFF identity flow (the bit #600 got wrong):

   ```
   sign-in:  session.user.id = <github numeric>
   create:   resolveOrCreateUser(github, <num>) = <canonical UUID>
             → session_participants(user_id=<UUID>, session_id=X)
   list:     BFF /api/sessions?mine=true injects mineScmUserId=<num>&mineProvider=github
             → getIdentity(github, <num>) = <same UUID>
             → JOIN matches
   ```

[...]

PR #599, opened 2026-05-07
feat: surface creator identity in session listings

ColeMurray:
## Summary
- LEFT JOIN `sessions` to `users` in `SessionIndexStore.list` to expose `creatorDisplayName` and `creatorAvatarUrl` per row.
- Extend the shared `Session` type so the wire format declares creator fields.
- Backwards compatible: existing clients ignore the new optional fields.

## Why
First of five stacked PRs implementing #557 (Session filters by user). This PR is the data foundation — surfacing creator identity per row — so subsequent PRs can wire up the per-row creator avatar in the sidebar's "all" view without further schema work. No UX change here.

Spec / design docs live in `docs/internal/2026-05-06-session-filters-by-user-{research,design-decisions,spec,implementation-plan}.md`.

## Test plan
- [x] Unit tests: `SessionIndexStore.list` populates creator fields from joined users row, falls through to nulls when `user_id IS NULL` or unlinked
- [x] Integration tests (real D1): same coverage end-to-end against the actual migrations
- [x] All existing `SessionIndexStore` tests continue to pass (35 unit + 30 integration)
- [x] Workspace `typecheck` clean

## Notes
- `get()` does not JOIN — only `list()` enriches. Consumers that need creator info on a single row should switch to `list()` or fetch the user via `UserStore.getUserById`.
- One extra column in the SELECT projection, indexed JOIN on `users.id` PK. No measurable perf impact at D1 scale.
- Rollback: revert. No schema change, no migration.

## Summary by CodeRabbit

* **New Features**
  * Sessions now display creator identity information, including display name and avatar URL. Creator details are populated from linked user profiles and appear as null when no user is associated.

### DIFF
--- a/packages/control-plane/src/db/session-index.test.ts
+++ b/packages/control-plane/src/db/session-index.test.ts
@@ -24,13 +24,23 @@ type SessionRow = {
   pr_count: number;
   created_at: number;
   updated_at: number;
+  creator_display_name?: string | null;
+  creator_avatar_url?: string | null;
+};
+
+type UserRow = {
+  id: string;
+  display_name: string | null;
+  avatar_url: string | null;
 };
 
 const QUERY_PATTERNS = {
   INSERT_SESSION: /^INSERT OR IGNORE INTO sessions/,
   SELECT_BY_ID: /^SELECT \* FROM sessions WHERE id = \?$/,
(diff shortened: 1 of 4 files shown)


MartinRoberts-Fountain:
@ColeMurray is this feature still something you intend on adding in the future? We're looking at doing something like this ourselves if it's no longer on the roadmap for this project.

PR #622, opened 2026-05-11
fix(control-plane): dedupe presence list by participantId

donnfelker:
Fixes a bug where a prompt engineer would show up twice (or more) based upon how many tabs they had open. This change ensures that they only show up once in the list of prompt engineers. 

## Summary
- The presence list returned one entry per WebSocket client, so a single user with multiple connections (e.g. two browser tabs) appeared as multiple participants - rendering duplicate avatars in the UI and producing React duplicate-key warnings.
- Dedupe `getPresenceList()` output by `participantId`: any active socket marks the participant active, and the most recent `lastSeen` across sockets wins.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate presence entries for the same participant when connected from multiple tabs/devices; presence now shows a single entry per participant.
  * Presence now uses the most recent activity timestamp and treats any active connection as active for the participant, preventing premature removals.
* **Tests**
  * Added integration and unit tests validating presence deduplication and correct presence_update vs presence_leave behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ColeMurray/background-agents/pull/622)

### DIFF
--- a/packages/control-plane/src/session/presence-service.ts
+++ b/packages/control-plane/src/session/presence-service.ts
@@ -35,16 +35,30 @@ export class PresenceService {
 
   /**
    * Get list of present participants.
+   *
+   * A single participant can hold multiple WebSocket connections (e.g. two
+   * browser tabs), so we dedupe by participantId: any active socket marks the
+   * participant active, and we take the most recent lastSeen across sockets.
    */
   getPresenceList(): ParticipantPresence[] {
-    return Array.from(this.deps.getAuthenticatedClients()).map((c) => ({
-      participantId: c.participantId,
-      userId: c.userId,
-      name: c.name,
-      avatar: c.avatar,
-      status: c.status,
-      lastSeen: c.lastSeen,
-    }));
+    const byId = new Map<string, ParticipantPresence>();
+    for (const c of this.deps.getAuthenticatedClients()) {
+      const existing = byId.get(c.participantId);
+      if (!existing) {
+        byId.set(c.participantId, {
+          participantId: c.participantId,
+          userId: c.userId,
+          name: c.name,
+          avatar: c.avatar,
+          status: c.status,
+          lastSeen: c.lastSeen,
+        });
+        continue;
+      }
+      if (c.status === "active") existing.status = "active";
+      if (c.lastSeen > existing.lastSeen) existing.lastSeen = c.lastSeen;
+    }
+    return Array.from(byId.values());
   }
 
   /**
--- a/packages/control-plane/src/session/durable-object.ts
+++ b/packages/control-plane/src/session/durable-object.ts
@@ -1018,7 +1018,17 @@ export class SessionDO extends DurableObject<Env> {
       } else {
         const client = this.wsManager.removeClient(ws);
         if (client) {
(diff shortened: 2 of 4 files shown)


packages/control-plane/src/session/presence-service.ts
```diff
@@ -35,16 +35,30 @@ export class PresenceService {
...
-    return Array.from(this.deps.getAuthenticatedClients()).map((c) => ({
-      participantId: c.participantId,
-      userId: c.userId,
-      name: c.name,
-      avatar: c.avatar,
-      status: c.status,
-      lastSeen: c.lastSeen,
-    }));
+    const byId = new Map<string, ParticipantPresence>();
+    for (const c of this.deps.getAuthenticatedClients()) {
```
ColeMurray, line 45: [Automated Review] **Symmetric bug: `presence_leave` removes the participant even when other tabs are still connected.**

This PR fixes the duplicate-on-connect side of multi-tab presence, but the mirror-image bug on disconnect remains. In `packages/control-plane/src/session/durable-object.ts` around line 1019-1022:

```ts
const client = this.wsManager.removeClient(ws);
if (client) {
  this.broadcast({ type: "presence_leave", userId: client.userId });
}
```

The web client filters all entries by `userId` on `presence_leave` (`packages/web/src/hooks/use-session-socket.ts:360-362`):

```ts
case "presence_leave":
  setParticipants((prev) => prev.filter((p) => p.userId !== data.userId));
  break;
```

So if a user has two tabs open and closes one, every other viewer's UI removes the participant entirely — even though the user is still connected on the other tab. The user visually disappears until the next `presence_update`, partially undoing the fix this PR is making.

Suggested fix in this PR (small and directly related):

```ts
const client = this.wsManager.removeClient(ws);
if (client) {
  const stillPresent = Array.from(this.wsManager.getAuthenticatedClients())
    .some((c) => c.participantId === client.participantId);
  if (stillPresent) {
    this.presenceService.broadcastPresence(); // refresh — still here, possibly different status
  } else {
    this.broadcast({ type: "presence_leave", userId: client.userId });
  }
}
```

Alternatively, change `presence_leave` to carry `participantId` rather than `userId` and have the client filter on that — but that's a wire-protocol change requiring matching client updates.

ColeMurray (changes requested):
[Automated Review] Requesting changes — see inline comment for the symmetric `presence_leave` bug not addressed by this PR. The dedupe fix here is correct, but its mirror image on disconnect partially undoes the user-visible improvement.

packages/control-plane/src/session/presence-service.ts
donnfelker: Addressed in [`2cb1f94`](https://github.com/ColeMurray/background-agents/commit/2cb1f94ef050df987c9f737a1f016581354d2870).

Gated `presence_leave` on the last-socket case: after `removeClient`, check whether any other authenticated socket still belongs to the same `participantId`. If yes → `presenceService.broadcastPresence()` (deduped refresh); if no → `presence_leave` as before. Also added two integration tests in `websocket-client.test.ts` covering both branches.

ColeMurray (approved):

PR #628, opened 2026-05-12
feat(web): allow per-repo override of auto-review-on-open in GitHub integration settings

donnfelker:
## Summary
- Adds an "Auto-review new PRs" control to the per-repo override row in the GitHub integration settings UI.
- Lets users either inherit the global `autoReviewOnOpen` default or override it for a specific repo with an on/off toggle.
- Pure UI change wired through the existing `RepoOverrideRow` save flow.

**Example: Auto review PR's is off globally , but enabled on this repo**
CleanShot 2026-05-12 at 11 51 44@2x

## Test plan
- Manual: open Settings → Integrations → GitHub, add a repo override, switch the new control between "Use global default" and "Override for this repo", flip the Switch, save, and confirm the override persists.

## Summary by CodeRabbit

* **New Features**
  * Added per-repository override capability for "Auto-review new PRs" in GitHub integration settings
  * Users can now choose to override global settings and customize auto-review behavior for individual repositories

[Review Change Stack](https://app.coderabbit.ai/change-stack/ColeMurray/background-agents/pull/628)

### DIFF
--- a/packages/web/src/components/settings/integrations/github-integration-settings.tsx
+++ b/packages/web/src/components/settings/integrations/github-integration-settings.tsx
@@ -608,6 +608,12 @@ function RepoOverrideRow({
   const [commentActionInstructions, setCommentActionInstructions] = useState(
     entry.settings.commentActionInstructions ?? ""
   );
+  const [autoReviewMode, setAutoReviewMode] = useState<"global" | "override">(
+    entry.settings.autoReviewOnOpen !== undefined ? "override" : "global"
+  );
+  const [autoReviewOnOpen, setAutoReviewOnOpen] = useState(
+    entry.settings.autoReviewOnOpen ?? false
+  );
   const [newUsername, setNewUsername] = useState("");
   const [saving, setSaving] = useState(false);
   const [dirty, setDirty] = useState(false);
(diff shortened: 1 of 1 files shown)


packages/web/src/components/settings/integrations/github-integration-settings.tsx
```diff
@@ -608,6 +608,12 @@ function RepoOverrideRow({
   const [commentActionInstructions, setCommentActionInstructions] = useState(
     entry.settings.commentActionInstructions ?? ""
   );
+  const [autoReviewMode, setAutoReviewMode] = useState<"global" | "override">(
+    entry.settings.autoReviewOnOpen !== undefined ? "override" : "global"
+  );
+  const [autoReviewOnOpen, setAutoReviewOnOpen] = useState(
+    entry.settings.autoReviewOnOpen ?? false
```
ColeMurray, line 615: Small UX concern: this defaults the override switch to `false`, but the global init at line 125 in this same file defaults to `true` (`settings?.defaults?.autoReviewOnOpen ?? true`). So a user who clicks "Override for this repo" and hits Save without touching the switch silently flips auto-review from **on → off** for that repo while the global stays on.

Could we plumb the global default down into `RepoOverrideRow` and seed this state from it? That way switching to "override" preserves the current effective behavior until the user deliberately changes it. At minimum, mirror the `?? true` polarity here so the override doesn't bias toward a silent downgrade.

The text overrides (`codeReviewInstructions ?? ""`, etc.) have the same shape, but a boolean is sharper — there's no "field is empty" visual cue to tip the user off.

ColeMurray: will follow up in separate PR

ColeMurray (approved):

PR #629, opened 2026-05-12
fix(modal): Support non-main envs in Modal

kadams54:
## Summary

Modal has a hierarchy of workspaces -> environments. The default environment is `main`. If you stick with that, the environment doesn't have any impact on Modal's URLs; however, if you use a non-default environment, e.g., `prod`, that environment name is added into the URL. These changes support generating Modal's URL slug properly to support both `main` and non-default environments.

## Backwards compatibility

- The new `modal_environment` Terraform variable defaults to `main` so projects that don't set it will not be impacted.

## Docs updated

- `docs/GETTING_STARTED.md` — updated `terraform.tfvars` block
- `terraform/environments/production/terraform.tfvars.example` — updated example to include the new variable

## Summary by CodeRabbit

* **Documentation**
  * Updated getting started guide with expanded Modal prerequisites, now requiring both Workspace and Environment names with guidance on locating them.

* **New Features**
  * Terraform configuration now includes environment variable support for Modal, enabling flexible workspace configuration across different deployment environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ColeMurray/background-agents/pull/629)

### DIFF
--- a/terraform/environments/production/modal.tf
+++ b/terraform/environments/production/modal.tf
@@ -28,7 +28,7 @@ module "modal_app" {
   modal_token_secret = var.modal_token_secret
 
   app_name      = "open-inspect"
-  workspace     = var.modal_workspace
+  workspace     = local.modal_workspace_slug
   deploy_path   = "${var.project_root}/packages/modal-infra"
   deploy_module = "deploy"
   source_hash   = data.external.modal_source_hash[0].result.hash
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -150,7 +150,8 @@ Create an R2 API Token:
 1. Go to [Modal Settings](https://modal.com/settings)
 2. **Create a new API token**: Settings -> API Tokens -> New Token
 3. Note the **Token ID** and **Token Secret**
-4. Note your **Workspace name** (visible in your Modal dashboard URL)
+4. Note your **Workspace** and **Environment names** (visible in your Modal dashboard URL,
+   https://modal.com/apps/<modal_workspace>/<modal_environment>)
 
 ### Daytona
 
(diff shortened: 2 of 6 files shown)


terraform/environments/production/modal.tf
```diff
@@ -28,7 +28,7 @@ module "modal_app" {
   modal_token_secret = var.modal_token_secret
 
   app_name      = "open-inspect"
-  workspace     = var.modal_workspace
+  workspace     = local.modal_workspace_slug
```
ColeMurray, line 31: [Automated Review]
The slug is plumbed into URL construction here, but `modal deploy` / `modal secret create` in `terraform/modules/modal-app/scripts/{deploy,create-secrets}.sh` are invoked without `--env` (and no `MODAL_ENVIRONMENT` is exported), so they deploy to the workspace's *default* environment regardless of `var.modal_environment`.

Concretely, with `modal_environment = "production"`:
- This module's outputs and the worker's `MODAL_WORKSPACE` → `myworkspace-production--…modal.run`
- But `modal deploy` lands the app at the workspace default (typically `main`) → `myworkspace--…modal.run`
- Result: every endpoint 404s.

Suggested fix:
1. Add a `modal_environment` input to the `modal-app` module.
2. Pass it as `MODAL_ENVIRONMENT` in the `environment` blocks of `null_resource.modal_secrets` and `null_resource.modal_deploy` in `modules/modal-app/main.tf` — Modal CLI reads that var natively.
   (Or alternatively, append `--env "${MODAL_ENVIRONMENT}"` in both scripts.)

Without this, the feature is half-wired for non-`main` envs — URLs point one place, deployment lands somewhere else.

ColeMurray (changes requested):
see comment regarding modal script

ColeMurray:
resolve by https://github.com/ColeMurray/background-agents/pull/687

PR #654, opened 2026-05-18
feat(sandbox): add provider dashboard links

hreiten:
## Summary

- add a provider-neutral `sandbox_dashboard_url` server message and session state field
- broadcast the dashboard URL after a sandbox provider object ID is persisted on spawn, restore, or persistent resume
- render the desktop sandbox status as an external provider-dashboard link when a URL is available

## Why

Modal exposes useful per-sandbox diagnostics in its dashboard, but the web UI currently has no way to link users from a session to the corresponding provider sandbox. This makes debugging stuck, slow, or failed sandboxes harder than it needs to be.

The implementation keeps the shared protocol provider-neutral and confines Modal URL construction to the control-plane Modal path. Existing connected clients receive the URL as soon as the provider object ID is known, while newly subscribed clients receive it through the hydrated session state.

## Prior upstream work checked

- refreshed `upstream/main` before opening this PR; this branch is one commit ahead and zero commits behind
- checked `upstream/main` for `sandboxDashboardUrl`, `sandbox_dashboard_url`, and `modal_sandbox_url`; no existing implementation was present
- searched existing upstream PRs for sandbox/dashboard/Modal dashboard terms; no matching fix for this URL plumbing was found
- noted open PR #629 for non-main Modal environments; this PR defaults the Modal dashboard environment to `main` and does not add new Terraform/env wiring

## Details

- `buildModalSandboxDashboardUrl` returns `null` unless the required Modal workspace and provider object ID are present.
- `SandboxLifecycleManager` centralizes provider object ID persistence and broadcasts the dashboard URL only when configured.
- `SessionDO` wires the builder only for the Modal backend.
- The web socket hook clears stale dashboard URLs when a replacement sandbox starts or the current sandbox reaches a terminal state.

## Tests

- `npm run lint`
- `npm run typecheck`
- `npm test`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Session header shows a clickable sandbox provider dashboard link when available; session state and server messages now include an optional sandbox dashboard URL.

* **Bug Fixes**
  * Dashboard link visibility and updates are more consistent across spawn, restore, and resume flows; dashboard URL cleared or retained appropriately during lifecycle transitions.

* **Tests**
  * Added tests for URL construction, message delivery, session-state hydration, and sandbox access state transitions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ColeMurray/background-agents/pull/654?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/control-plane/src/sandbox/client.ts
+++ b/packages/control-plane/src/sandbox/client.ts
@@ -14,6 +14,7 @@ const log = createLogger("modal-client");
 
 // Modal app name
 const MODAL_APP_NAME = "open-inspect";
+const DEFAULT_MODAL_DASHBOARD_ENVIRONMENT = "main";
 
 /**
  * Construct the Modal base URL from workspace name.
@@ -21,6 +22,16 @@ const MODAL_APP_NAME = "open-inspect";
  */
 function getModalBaseUrl(workspace: string): string {
   return `https://${workspace}--${MODAL_APP_NAME}`;
+}
+
+export function buildModalSandboxDashboardUrl(params: {
+  workspace: string | undefined;
+  environment?: string | undefined;
+  providerObjectId: string | null | undefined;
+}): string | null {
+  if (!params.workspace || !params.providerObjectId) return null;
+  const environment = params.environment || DEFAULT_MODAL_DASHBOARD_ENVIRONMENT;
+  return `https://modal.com/apps/${params.workspace}/${environment}/deployed/${MODAL_APP_NAME}?activeTab=sandboxes&sandboxId=${params.providerObjectId}`;
 }
 
 export interface CreateSandboxRequest {
--- a/packages/control-plane/src/sandbox/client.ts
+++ b/packages/control-plane/src/sandbox/client.ts
@@ -14,6 +14,8 @@ const log = createLogger("modal-client");
 
 // Modal app name
 const MODAL_APP_NAME = "open-inspect";
+
+// Modal's default environment name; unrelated to the git branch named "main".
 const DEFAULT_MODAL_DASHBOARD_ENVIRONMENT = "main";
 
 /**
@@ -30,8 +32,10 @@ export function buildModalSandboxDashboardUrl(params: {
   providerObjectId: string | null | undefined;
 }): string | null {
   if (!params.workspace || !params.providerObjectId) return null;
-  const environment = params.environment || DEFAULT_MODAL_DASHBOARD_ENVIRONMENT;
-  return `https://modal.com/apps/${params.workspace}/${environment}/deployed/${MODAL_APP_NAME}?activeTab=sandboxes&sandboxId=${params.providerObjectId}`;
+  const workspace = encodeURIComponent(params.workspace);
+  const environment = encodeURIComponent(params.environment || DEFAULT_MODAL_DASHBOARD_ENVIRONMENT);
+  const providerObjectId = encodeURIComponent(params.providerObjectId);
+  return `https://modal.com/apps/${workspace}/${environment}/deployed/${MODAL_APP_NAME}?activeTab=sandboxes&sandboxId=${providerObjectId}`;
 }
 
 export interface CreateSandboxRequest {
--- a/packages/control-plane/src/sandbox/lifecycle/manager.test.ts
+++ b/packages/control-plane/src/sandbox/lifecycle/manager.test.ts
@@ -130,6 +130,7 @@ function createMockStorage(
         sandbox.auth_token_hash = data.authTokenHash;
         sandbox.auth_token = null;
         sandbox.modal_sandbox_id = data.modalSandboxId;
+        sandbox.modal_object_id = null;
       }
     }),
     updateSandboxModalObjectId: vi.fn((id: string) => {
@@ -671,6 +672,50 @@ describe("SandboxLifecycleManager", () => {
         {
           type: "sandbox_dashboard_url",
           url: "https://provider.example/new-provider-obj",
+        },
+      ]);
+    });
+
+    it("broadcasts sandbox_dashboard_url after resume when provider object id is unchanged", async () => {
+      const sandbox = createMockSandbox({
+        status: "stopped",
+        modal_object_id: "same-provider-obj",
+        snapshot_image_id: null,
+      });
+      const storage = createMockStorage(createMockSession(), sandbox);
+      const broadcaster = createMockBroadcaster();
+      const provider = createMockProvider({
+        capabilities: { supportsPersistentResume: true },
+        resumeSandbox: vi.fn(async () => ({
+          success: true,
+          providerObjectId: "same-provider-obj",
+        })),
+      });
+      const config = {
+        ...createTestConfig(),
+        sandboxDashboardUrlBuilder: (id: string) => `https://provider.example/${id}`,
+      };
+
+      const manager = new SandboxLifecycleManager(
+        provider,
+        storage,
+        broadcaster,
+        createMockWebSocketManager(false),
+        createMockAlarmScheduler(),
+        createMockIdGenerator(),
+        config
+      );
+
+      await manager.spawnSandbox();
+
+      expect(provider.resumeSandbox).toHaveBeenCalled();
+      expect(storage.calls).not.toContain("updateSandboxModalObjectId:same-provider-obj");
+      expect(
+        broadcaster.messages.filter((m) => (m as { type: string }).type === "sandbox_dashboard_url")
+      ).toEqual([
+        {
+          type: "sandbox_dashboard_url",
+          url: "https://provider.example/same-provider-obj",
         },
       ]);
     });
--- a/packages/control-plane/src/sandbox/lifecycle/manager.test.ts
+++ b/packages/control-plane/src/sandbox/lifecycle/manager.test.ts
@@ -24,6 +24,8 @@ import {
   type CreateSandboxResult,
   type RestoreConfig,
   type RestoreResult,
+  type ResumeConfig,
+  type ResumeResult,
   type SnapshotConfig,
   type SnapshotResult,
 } from "../provider";
@@ -254,15 +256,18 @@ function createMockProvider(
   overrides: Partial<{
     createSandbox: (config: CreateSandboxConfig) => Promise<CreateSandboxResult>;
     restoreFromSnapshot: (config: RestoreConfig) => Promise<RestoreResult>;
+    resumeSandbox: (config: ResumeConfig) => Promise<ResumeResult>;
     takeSnapshot: (config: SnapshotConfig) => Promise<SnapshotResult>;
+    capabilities: Partial<SandboxProvider["capabilities"]>;
   }> = {}
 ): SandboxProvider {
-  return {
+  const provider: SandboxProvider = {
     name: "mock",
     capabilities: {
       supportsSnapshots: true,
       supportsRestore: true,
       supportsWarm: true,
+      ...overrides.capabilities,
     },
     createSandbox:
       overrides.createSandbox ||
@@ -285,6 +290,10 @@ function createMockProvider(
         imageId: "snapshot-img-123",
       })),
   };
+  if (overrides.resumeSandbox) {
+    provider.resumeSandbox = overrides.resumeSandbox;
+  }
+  return provider;
 }
 
 function createTestConfig(): SandboxLifecycleConfig {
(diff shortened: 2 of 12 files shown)


packages/control-plane/src/sandbox/lifecycle/manager.ts
```diff
@@ -751,7 +753,7 @@ export class SandboxLifecycleManager {
       }
```
ColeMurray, line 752: [Automated Review] **Critical — resume with unchanged `providerObjectId` never broadcasts the URL.** This is the *common* resume path (most resumes return the same provider object ID). Combined with the web hook's clearing on `sandbox_status: stale|stopped|failed` (`use-session-socket.ts:391-407`), active subscribers fall into this trap:

1. Sandbox transitions `running → stopped` → web clears `sandboxDashboardUrl`.
2. User triggers a prompt → server runs `resumeSandbox`; provider returns the same ID.
3. Server broadcasts only `sandbox_status: connecting` (no `sandbox_dashboard_url`).
4. Client shows `Sandbox: connecting` with no link until they reload and re-hydrate.

Notice that `code_server_info` *is* re-broadcast unconditionally a few lines down (`L760`) — same path, different behavior for the dashboard URL.

Suggested fix: separate persistence from broadcast and call the broadcast unconditionally on successful resume:
```ts
const finalProviderObjectId = result.providerObjectId ?? providerObjectId;
if (result.providerObjectId && result.providerObjectId !== providerObjectId) {
  this.storage.updateSandboxModalObjectId(result.providerObjectId);
}
this.broadcastSandboxDashboardUrl(finalProviderObjectId);
```
Worth a regression test alongside the existing 'when provider object id changes' case at `manager.test.ts:634`.

packages/control-plane/src/session/durable-object.ts
```diff
@@ -110,6 +110,12 @@ const WS_TOKEN_TTL_MS = 24 * 60 * 60 * 1000; // 24 hours
 
 /** Statuses that indicate a session is finished — metrics are synced to D1 on these transitions. */
 const TERMINAL_STATUSES: SessionStatus[] = ["completed", "failed", "cancelled"];
+const SANDBOX_DASHBOARD_URL_BLOCKED_STATUSES = new Set<SandboxStatus>([
+  "spawning",
+  "stale",
+  "stopped",
+  "failed",
+]);
```
ColeMurray: [Automated Review] **Critical — blocked-status list contradicts the PR's stated motivation.** The PR description says: *'Modal exposes useful per-sandbox diagnostics in its dashboard… debugging stuck, slow, or failed sandboxes harder than it needs to be.'* But this set hides the URL for exactly the states (`stale`, `stopped`, `failed`) where users most need it. Modal does not delete sandboxes immediately on stop/fail — the dashboard remains useful for log/state inspection.

Recommended changes:
- Hydrate (`L1700`): allow the URL whenever `providerObjectId` is non-null, regardless of status. (`spawning` is self-handled since `providerObjectId` is null until the provider returns.)
- Web (`use-session-socket.ts:391-395`): drop `stale|stopped|failed` from the dashboard-URL clearing path — or split clearing of the read-only diagnostic URL from clearing of access-credential URLs like `codeServerUrl`/`ttydToken`.

If there *is* a deliberate reason (e.g., Modal's dashboard 404s for a specific class of pre-spawn failure), please document it inline and narrow the blocklist.

packages/control-plane/src/sandbox/client.ts
```diff
@@ -23,6 +24,16 @@ function getModalBaseUrl(workspace: string): string {
...
 }
 
+export function buildModalSandboxDashboardUrl(params: {
+  workspace: string | undefined;
+  environment?: string | undefined;
+  providerObjectId: string | null | undefined;
+}): string | null {
+  if (!params.workspace || !params.providerObjectId) return null;
+  const environment = params.environment || DEFAULT_MODAL_DASHBOARD_ENVIRONMENT;
+  return `https://modal.com/apps/${params.workspace}/${environment}/deployed/${MODAL_APP_NAME}?activeTab=sandboxes&sandboxId=${params.providerObjectId}`;
```
ColeMurray: [Automated Review] Interpolated values aren't URL-encoded. Modal IDs are alphanumeric+hyphen today, so this works, but a future Modal ID format change or unusual workspace name would break the URL silently. `encodeURIComponent` on `workspace`, `environment`, and `providerObjectId` is cheap insurance (leave `MODAL_APP_NAME` as-is — it's a constant).

packages/control-plane/src/session/durable-object.ts
```diff
@@ -1676,9 +1688,22 @@ export class SessionDO extends DurableObject<Env> {
...
+    providerObjectId: string | null | undefined,
+    sandboxStatus?: SandboxStatus | null
+  ): string | null {
+    if (resolveSandboxBackendName(this.env.SANDBOX_PROVIDER) !== "modal") return null;
+    if (sandboxStatus && SANDBOX_DASHBOARD_URL_BLOCKED_STATUSES.has(sandboxStatus)) return null;
+    return buildModalSandboxDashboardUrl({
+      workspace: this.env.MODAL_WORKSPACE,
+      providerObjectId,
+    });
+  }
```
ColeMurray: [Automated Review] This method has an implicit overload — pass `sandboxStatus` and it filters; omit it and it doesn't. Callers in this PR rely on that distinction (broadcast closure at `L737-740` omits status; hydrate at `L1691` passes it). Consider making the contract explicit:
- Two named methods (`buildDashboardUrlForBroadcast(providerObjectId)` / `buildDashboardUrlForHydration(providerObjectId, status)`), or
- Make the parameter required (`sandboxStatus: SandboxStatus | null`) so the broadcast site explicitly passes `null`.

Related: if you address the blocked-statuses critical above, this method may collapse to a one-liner around `buildModalSandboxDashboardUrl`.

packages/control-plane/src/sandbox/lifecycle/manager.ts
```diff
@@ -1146,6 +1148,14 @@ export class SandboxLifecycleManager {
     return this.wsManager.getConnectedClientCount();
   }
 
+  private storeAndBroadcastProviderObjectId(providerObjectId: string): void {
+    this.storage.updateSandboxModalObjectId(providerObjectId);
+    const url = this.config.sandboxDashboardUrlBuilder?.(providerObjectId);
+    if (url) {
+      this.broadcaster.broadcast({ type: "sandbox_dashboard_url", url });
+    }
+  }
```
ColeMurray, line 1168: [Automated Review] The sibling `storeAndBroadcastCodeServer` (just below at `L1166`) logs at info level when it broadcasts; this helper doesn't log at all. Adding `log.debug('Broadcasting sandbox dashboard URL', { provider_object_id: providerObjectId })` keeps the two helpers consistent and makes the new wire traffic traceable in production.

packages/control-plane/src/sandbox/lifecycle/manager.test.ts
```diff
@@ -525,6 +589,92 @@ describe("SandboxLifecycleManager", () => {
...
+        broadcaster.messages.filter((m) => (m as { type: string }).type === "sandbox_dashboard_url")
+      ).toEqual([
+        {
+          type: "sandbox_dashboard_url",
+          url: "https://provider.example/restored-obj-456",
+        },
+      ]);
+    });
+
+    it("broadcasts sandbox_dashboard_url after resume when provider object id changes", async () => {
```
ColeMurray, line 635: [Automated Review] Missing a complementary test for the *typical* resume path: when `resumeSandbox` returns the same `providerObjectId`. Today the broadcast is skipped in that case — which is the critical bug flagged on `manager.ts:754`. After fixing that, this is the regression test that locks the behavior in.

packages/control-plane/src/sandbox/client.ts
```diff
@@ -14,6 +14,7 @@ const log = createLogger("modal-client");
 
 // Modal app name
 const MODAL_APP_NAME = "open-inspect";
+const DEFAULT_MODAL_DASHBOARD_ENVIRONMENT = "main";
```
ColeMurray, line 19: [Automated Review] Nit: a one-line comment explaining that `'main'` is Modal's default *environment* name (not git's `main` branch) would prevent future confusion. Also — is this constant a leave-behind hook for #629 (non-`main` Modal environments)? Worth a `// TODO(#629): make configurable via env` if so, otherwise consider just inlining.

packages/control-plane/src/sandbox/lifecycle/manager.test.ts
```diff
@@ -285,6 +290,10 @@ function createMockProvider(
         imageId: "snapshot-img-123",
       })),
   };
+  if (overrides.resumeSandbox) {
+    provider.resumeSandbox = overrides.resumeSandbox;
+  }
```
ColeMurray, line 296: [Automated Review] Nit: `resumeSandbox` is only assigned when the override is provided; the other capability methods (`createSandbox`, `restoreFromSnapshot`, `takeSnapshot`) are always present with defaults. Consider always providing a default (or document inline that the absence is intentional). Tests that don't pass `supportsPersistentResume: true` won't reach the resume branch anyway, so a no-op default is safe.

packages/control-plane/test/integration/websocket-client.test.ts
```diff
@@ -31,6 +31,42 @@ describe("Client WebSocket (via SELF.fetch)", () => {
...
+      const { ws, messages } = await openClientWs(name, { subscribe: true });
+      const subscribed = messages!.find((m) => m.type === "subscribed") as Record<string, unknown>;
+      const state = subscribed.state as Record<string, unknown>;
+
+      expect(state.sandboxStatus).toBe(testCase.status);
+      expect(state.sandboxDashboardUrl).toBe(testCase.expectedDashboardUrl);
+
+      ws.close();
+    }
+  });
```
ColeMurray: [Automated Review] Nit: `it.each(cases)('hydrates dashboard URL for status %s', ...)` produces per-status test names, making failure isolation easier than a single `for...of` inside one `it`. Also tightens the loop body since the test name carries the case description.

packages/control-plane/src/session/durable-object.ts
```diff
@@ -728,6 +734,11 @@ export class SessionDO extends DurableObject<Env> {
       };
     }
 
+    const sandboxDashboardUrlBuilder =
+      sandboxBackend === "modal"
+        ? (providerObjectId: string) => this.getSandboxDashboardUrl(providerObjectId)
+        : undefined;
```
ColeMurray, line 734: [Automated Review] Nit: a brief comment (`// Bound here so the lifecycle manager stays provider-agnostic`) would help future readers understand why this wiring lives in `createLifecycleManager` rather than inside the manager itself. The closure captures `this`, but that's safe because DO instances retain identity across method calls; a one-liner would make that intent obvious.

packages/web/src/app/(app)/session/[id]/page.tsx
```diff
@@ -1128,7 +1137,24 @@ function SandboxStatus({ status }: { status?: string }) {
...
+  const className = `text-xs ${colors[status] || colors.pending}`;
+  const label = `Sandbox: ${status}`;
+
+  if (dashboardUrl) {
+    return (
+      <a
+        href={dashboardUrl}
+        target="_blank"
+        rel="noreferrer noopener"
+        title="Open sandbox in provider dashboard"
```
Comment (ColeMurray) [inline packages/web/src/app/(app)/session/[id]/page.tsx:1149]: [Automated Review] Nit: `title="Open Modal dashboard"` would be more specific for the current Modal-only implementation. Revert to a generic label once Daytona (or another provider) also exposes a dashboard URL.

ColeMurray (commented):
[Automated Review] Posting inline feedback from a deep review of PR #654. Headline: 2 critical issues (resume-path broadcast gap, blocked-status list contradicts the PR's stated debugging motivation), 4 suggestions, 5 nitpicks. The architecture itself is clean — provider-neutral protocol, Modal URL construction confined to `client.ts`, and the new `storeAndBroadcastProviderObjectId` helper nicely consolidates three previously-duplicated lifecycle paths. Test coverage spans unit, integration, and React-hook layers. Filing this as a Comment rather than Request Changes — the two criticals are both quick fixes once the design intent is settled.

ColeMurray (dismissed):
see comments

packages/control-plane/src/sandbox/lifecycle/manager.ts
hreiten: Fixed in `990c1b3`. The resume path now computes `finalProviderObjectId = result.providerObjectId ?? providerObjectId`, persists only when the ID changes, and broadcasts the dashboard URL on every successful resume. This matches the code-server rebroadcast behavior and covers the common unchanged-ID path.

packages/control-plane/src/session/durable-object.ts
hreiten: Fixed in `990c1b3`. I removed the dashboard blocked-status list: hydration now returns the Modal dashboard URL whenever the backend is Modal and `modal_object_id` exists. The client also preserves the diagnostic URL for `stale`, `stopped`, `failed`, and `sandbox_error` while still clearing access credentials. To avoid stale links during actual replacements, fresh spawn/restore now clear `modal_object_id` through `updateSandboxForSpawn`, and restore broadcasts `spawning` only after that atomic write.

packages/control-plane/src/sandbox/client.ts
hreiten: Fixed in `990c1b3`. The Modal URL builder now uses `encodeURIComponent` for `workspace`, `environment`, and `providerObjectId`, with coverage for spaces, slashes, and query-like characters.

packages/control-plane/src/session/durable-object.ts
hreiten: Fixed by simplifying the method in `990c1b3`. After removing status-based filtering, `getSandboxDashboardUrl` no longer has the implicit overload; it only takes `providerObjectId` and returns a Modal URL when the Modal backend is active.

packages/control-plane/src/sandbox/lifecycle/manager.ts
hreiten: Fixed in `990c1b3`. Dashboard URL broadcasts now emit a debug log with the provider object ID before sending `sandbox_dashboard_url`.

packages/control-plane/src/sandbox/lifecycle/manager.test.ts
hreiten: Fixed in `990c1b3`. Added a regression test for successful resume with an unchanged provider object ID. The test asserts that we do not rewrite the same ID but still broadcast the dashboard URL.

packages/control-plane/src/sandbox/client.ts
hreiten: Addressed the useful part in `990c1b3`: added a short comment clarifying that `main` is Modal's default environment name, not the git branch. I intentionally did not add a TODO for #629 here because making the environment configurable is separate deployment wiring and outside this PR's scope.

packages/control-plane/src/sandbox/lifecycle/manager.test.ts
hreiten: Intentionally left as-is. `resumeSandbox` being optional in the mock mirrors the provider capability shape, and tests that exercise resume opt in explicitly with `supportsPersistentResume` plus a `resumeSandbox` override. Adding a default implementation would make the helper less faithful and could hide capability-gating mistakes.

packages/control-plane/test/integration/websocket-client.test.ts
hreiten: Intentionally left as a loop. I changed the cases to cover the important behavior changes, including provider-object-ID presence/absence, but kept the shared setup in one test to avoid extra churn. The per-case session name includes status and whether an ID is present, which is enough context if this fails in the worker logs.

packages/control-plane/src/session/durable-object.ts
hreiten: Intentionally left without another comment. The lifecycle manager still receives only a provider-neutral URL builder, and the Modal-only binding is local to `SessionDO`; adding a comment there felt like restating the code rather than clarifying a non-obvious behavior.

packages/web/src/app/(app)/session/[id]/page.tsx
Comment (hreiten) [reply packages/web/src/app/(app)/session/[id]/page.tsx:1149]: Intentionally kept the generic title. Even though the current builder is Modal-only, the wire contract and UI prop are provider-neutral (`sandbox_dashboard_url` / `sandboxDashboardUrl`), so `provider dashboard` keeps the UI aligned with the protocol and avoids another future rename when another backend adds a URL.

ColeMurray (approved):

ColeMurray:
thanks!

PR #661, opened 2026-05-20
fix(github-bot): supersede prior CHANGES_REQUESTED on re-review

donnfelker:
## Summary
- When the bot is re-reviewing a PR, establish a baseline from its most recent opinionated review (APPROVED or CHANGES_REQUESTED) and weigh activity since that baseline rather than the full PR history.
- After CHANGES_REQUESTED, if activity since baseline resolves the blocking issues, submit a follow-up APPROVE that supersedes the prior CHANGES_REQUESTED and unblocks the PR.
- If new commits since baseline introduce a new blocking issue, submit REQUEST_CHANGES — a stale APPROVED is not auto-dismissed unless "Dismiss stale pull request approvals" is enabled in branch protection, so an explicit REQUEST_CHANGES is required.
- Pass the configured bot username (`GITHUB_BOT_USERNAME`) into the prompt so the baseline lookup queries the correct author.
- Mark GitHub comments as a `"request"` treatment in the untrusted-content envelope so the bot can act on them while still refusing instructions that try to override the prompt's safety rules.

## Test plan
- [x] `npm test -w @open-inspect/github-bot`

## Summary by CodeRabbit

* **New Features**
  * Bot now establishes a baseline from its most recent review decision (approval or change request) and intelligently assesses subsequent activity to determine whether to provide feedback, approve, or request changes.

* **Tests**
  * Extended test coverage to verify baseline review filtering and updated prompt generation behavior for improved decision-making logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ColeMurray/background-agents/pull/661?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/github-bot/src/handlers.ts
+++ b/packages/github-bot/src/handlers.ts
@@ -436,6 +436,7 @@ export async function handleIssueComment(
     commentBody,
     commenter: sender.login,
     isPublic: !repo.private,
+    botUsername: env.GITHUB_BOT_USERNAME,
     commentActionInstructions: config.commentActionInstructions,
   });
 
@@ -537,6 +538,7 @@ export async function handleReviewComment(
     commentBody,
     commenter: sender.login,
     isPublic: !repo.private,
+    botUsername: env.GITHUB_BOT_USERNAME,
     filePath: comment.path,
     diffHunk: comment.diff_hunk,
     commentId: comment.id,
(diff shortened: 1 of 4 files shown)


Blue101black:
Would be nice if this also just appended to the existing github review session instead of spinning up a new session so it has better context

PR #679, opened 2026-05-23
feat(sandbox): broker SCM credentials for git operations

hreiten:
## Why

We hit this in our own Open-Inspect deployment: long-running or revisited sessions could no longer run git commands after the GitHub App installation token expired. The common workflow was to kick off an agent, come back more than an hour later, and continue the same session. At that point follow-up fetch/push/ls-remote operations were still using the token captured when the sandbox was created, so git failed even though the session itself was otherwise healthy.

## What changed

- Add a sandbox-authenticated `POST /sessions/:id/scm-credentials` broker endpoint that mints fresh SCM credentials through the configured provider.
- Add provider-owned credential-helper auth for GitHub and GitLab, including GitHub installation token expiry metadata.
- Install a sandbox git credential helper in Modal and Daytona images so git fetch/push/clone/LFS operations request fresh credentials on demand.
- Stop injecting static git tokens into fresh interactive sandboxes. Keep explicit fallback behavior only for image-build, legacy snapshot, and repo-image compatibility paths.
- Rewrite stale credentialed `origin` remotes to plain HTTPS URLs before fetch so older snapshots route through the helper.
- Add a `gh` wrapper for GitHub CLI calls, since `gh` reads env tokens rather than git's credential protocol.

## Security and compatibility

The helper fails closed by scoping credential responses to HTTPS and the configured SCM host. It intentionally preserves the existing installation-wide repository access model for that host, so setup/start hooks can still clone auxiliary private repositories the installation can access. It caches credentials locally with a short refresh buffer and `0600` permissions, and it refuses stale-cache fallback on broker failures. Live sessions also refuse to fall back to static `VCS_CLONE_TOKEN` when control-plane context is present but incomplete.

Modal snapshot restores already mint a fresh fallback token during restore; the stale-token motivation here applies primarily to continuously running sessions and Daytona persistent resumes.

## Validation

- `npm run test:integration -w @open-inspect/control-plane -- test/integration/scm-credentials.test.ts`
- `npm run lint -w @open-inspect/control-plane`
- `cd packages/sandbox-runtime && uv run --extra dev pytest tests/test_git_credential_helper.py tests/test_gh_wrapper.py tests/test_entrypoint_build_mode.py tests/test_entrypoint_urls.py -q`
- `cd packages/sandbox-runtime && uv run --extra dev pytest tests -q`
- `cd packages/sandbox-runtime && uv run --extra dev ruff check src tests && uv run --extra dev ruff format --check src tests`
- `npx prettier --check README.md docs/HOW_IT_WORKS.md packages/control-plane/README.md packages/github-bot/README.md`
- `cd packages/modal-infra && uv run --extra dev pytest tests/test_sandbox_env_vars.py tests/test_web_api_create_sandbox.py -q`
- `cd packages/modal-infra && uv run --extra dev ruff check src tests && uv run --extra dev ruff format --check src tests`
- `python3 -m compileall packages/daytona-infra/src`

## Summary by CodeRabbit

* **New Features**
  * Brokered git credential flow with a sandbox-facing endpoint to request short-lived SCM credentials; providers can emit credential-helper credentials.

* **Documentation**
  * Updated security model, token architecture, HOW_IT_WORKS, provider checklist, and READMEs to describe brokered credentials, token taxonomy, and legacy fallbacks.

* **Infrastructure / Runtime**
  * Sandbox/runtime images and startup configure a git credential helper and GH wrapper; fresh sandboxes avoid embedding clone tokens.

* **Tests**
  * Expanded unit/integration tests for credential brokering, providers, sandbox boot/restore, and helper behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ColeMurray/background-agents/pull/679?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/sandbox-runtime/src/sandbox_runtime/entrypoint.py
```diff
@@ -38,6 +38,42 @@
...
+GH_WRAPPER_BODY = (
+    "#!/bin/sh\n"
+    f'REAL_GH="{GH_WRAPPER_REAL_PATH}"\n'
+    'if [ "${VCS_HOST:-github.com}" != "github.com" ]; then\n'
+    '  exec "$REAL_GH" "$@"\n'
+    "fi\n"
+    "# A real user token wins; a marked fallback token does not.\n"
+    'if [ "$OI_GITHUB_TOKEN_IS_FALLBACK" != "1" ] && '
+    '{ [ -n "$GH_TOKEN" ] || [ -n "$GITHUB_TOKEN" ] || [ -n "$GITHUB_APP_TOKEN" ]; }; then\n'
+    '  exec "$REAL_GH" "$@"\n'
```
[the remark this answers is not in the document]
hreiten: Handled in `17123c2`. I tightened this without accepting every `GITHUB_APP_TOKEN` under `OI_GITHUB_TOKEN_IS_FALLBACK=1`, because the manager-injected fallback sets matching `GITHUB_TOKEN` and `GITHUB_APP_TOKEN` aliases. The wrapper now preserves the intended stale-fallback refresh path, while still respecting real user overrides: `GH_TOKEN` always wins, unmarked GitHub token env wins, and marked aliases only pass through when the `GITHUB_TOKEN`/`GITHUB_APP_TOKEN` values differ. Added shell-level wrapper coverage for the fallback-app-token and user-override cases.

packages/sandbox-runtime/src/sandbox_runtime/entrypoint.py
```diff
@@ -168,27 +206,117 @@ async def _clone_repo(self) -> bool:
...
+            # config baked into the image is the primary path anyway.
+            self.log.warn("credential_helper.shim_write_failed", error=str(e))
+
+        # credential.useHttpPath makes git pass the repo path to the helper,
+        # which it needs to scope credentials to the session repo (not just
+        # the host).
+        for key, value in (
+            ("credential.helper", str(shim_path)),
+            ("credential.useHttpPath", "true"),
+        ):
```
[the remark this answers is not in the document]
hreiten: Handled in `17123c2`. `_ensure_credential_helper_configured()` now tracks whether the shim is actually available: it only writes `credential.helper=/usr/local/bin/oi-git-credentials` when the existing shim matches or the write/chmod succeeds. It still configures `credential.useHttpPath=true` even if the shim write fails. Added a test assertion for the read-only shim path so git is not pointed at a missing helper.

packages/sandbox-runtime/src/sandbox_runtime/credentials/git_credential_helper.py
```diff
@@ -0,0 +1,365 @@
...
+    # helper is installed, so git always passes it. A missing path means a
+    # misconfigured image — fail closed rather than fall back to host-only
+    # scoping, which would re-open the same-host leak.
+    requested_path = input_lines.get("path", "").strip()
+    if not requested_path:
+        return False, "no path provided (credential.useHttpPath unset?)"
+    expected_path = _expected_repo_path()
+    if expected_path is None:
+        return False, "REPO_OWNER/REPO_NAME not configured"
+    if _normalize_repo_path(requested_path) != expected_path:
```
ColeMurray: This path scoping is a behavior change for existing sandbox setup flows.

  Today, fresh sandboxes receive the GitHub App token in env, so repo hooks can clone sibling/private repos during `.openinspect/setup.sh` or `.openinspect/start.sh`, for
  example:

  ```sh
  git clone https://github.com/acme/backend.git
  # or
  git clone https://x-access-token:$GITHUB_TOKEN@github.com/acme/backend.git

```
  With this helper, Git will call the credential helper for `path=acme/backend.git`, but `_is_authorized_request()` normalizes that path and requires it to equal `REPO_OWNER/
  REPO_NAME`. If the session repo is acme/frontend, the helper returns no credentials, and fresh sandboxes also no longer have `GITHUB_TOKEN / GITHUB_APP_TOKEN /
  VCS_CLONE_TOKEN` as a fallback.

That is a good security default, but it is backwards incompatible for users who currently clone auxiliary private repos in setup/start scripts for tests, shared backend
  services, monorepos split across repos, etc.

I think we can relax this constraint since the current system already uses installation-wide tokens

File: packages/modal-infra/src/sandbox/manager.py [hunk 4]
```diff
@@ -247,8 +247,32 @@ async def _write_tunnel_env_file(
...
+    ) -> None:
+        """Inject SCM provider metadata into the sandbox environment.
+
+        For interactive sandboxes ``clone_token`` should be ``None``. Git
+        authenticates per-request via the system git credential helper, which
+        fetches a fresh token from the control plane — embedding a token in
+        env would silently fail once it expires (or immediately, for
+        providers with short-lived tokens like GitHub Apps).
+
+        For image-build sandboxes (one-shot, no control-plane access)
```
ColeMurray, line 264: not required in this PR, but it likely makes sense as a follow-up to have all token auth run through the credentials helper to avoid two different credential approaches

ColeMurray (dismissed):
see comments

ColeMurray:
Small correction on the restore motivation: for Modal snapshot restores, we already mint and inject a fresh auth token. `api_restore_sandbox` calls `_resolve_clone_token()` per restore request, passes that into `restore_from_snapshot`, and the restored entrypoint rewrites `origin` via `_ensure_remote_auth()` before fetching.

Daytona is different: it does not use Modal-style snapshot restore. It resumes the same persistent sandbox via `startSandbox`/`recoverSandbox`, so env vars and credentialed remotes are not rebuilt. The stale-token issue applies to continuously running sandboxes and Daytona persistent resumes, but not Modal snapshot restores.

One extra Daytona note after tracing this more closely: existing Daytona sandboxes created before this helper lands will not automatically get the new helper on resume, since they keep the old filesystem/env. They likely need a fresh spawn or an explicit resume-time migration/patch.

packages/sandbox-runtime/src/sandbox_runtime/credentials/git_credential_helper.py
hreiten: Thanks for the thorough review and for calling out the compatibility break. I agree this should preserve the existing installation-wide token model.

Pushed `ed148b9` to relax the helper authorization from session-repo scoped to HTTPS + configured `VCS_HOST` scoped. Same-host auxiliary/private repo clones now route through the helper again, while non-HTTPS and foreign-host requests are still refused.

I also updated the tests for sibling repo clones, host-only/no-path requests, and missing `REPO_OWNER` / `REPO_NAME`, and kept the refusal coverage for foreign hosts and non-HTTPS. The docs/comments and PR body now include your Modal vs Daytona correction: Modal snapshot restores mint a fresh fallback token; the stale-token issue is primarily long-running sessions and Daytona persistent resumes.

Validation from this pass:

- `cd packages/sandbox-runtime && uv run --extra dev pytest tests/test_git_credential_helper.py -q`
- `cd packages/sandbox-runtime && uv run --extra dev pytest tests/test_git_credential_helper.py tests/test_gh_wrapper.py tests/test_entrypoint_build_mode.py tests/test_entrypoint_urls.py -q`
- `cd packages/sandbox-runtime && uv run --extra dev pytest tests -q`
- `cd packages/sandbox-runtime && uv run --extra dev ruff check src tests`
- `cd packages/sandbox-runtime && uv run --extra dev ruff format --check src tests`
- `npx prettier --check README.md docs/HOW_IT_WORKS.md packages/control-plane/README.md packages/github-bot/README.md`

packages/modal-infra/src/sandbox/manager.py
hreiten: Agreed, that direction makes sense as a follow-up so we do not keep two token paths around longer than necessary. I left it out of this PR per your note and kept this pass focused on preserving same-host auxiliary repo compatibility.

ColeMurray (approved):

ColeMurray:
thanks!

PR #684, opened 2026-05-29
feat(slack): gate private channels/DMs behind allowPrivateChannels toggle

fredwanteeed:
## Summary

Adds a workspace-wide **`allowPrivateChannels`** Slack setting (Settings → Integrations → Slack), mirroring the existing `mentionsPolicy` global-only pattern.
Motivation : https://shopify.engineering/under-the-river

- **Default `true`** — backward-compatible; existing installs are unaffected.
- **When off**, the Slack bot only responds in public channels and declines **private channels, group DMs (mpim), and 1:1 DMs (im)** with a polite "public channels only" reply. No session is created.

### How it works
- A single guard (`refuseIfPrivateContext`) runs as the **first action** in all three session-creating ingresses — `handleAppMention`, `handleDirectMessage`, and the `/interactions` repo-select path (`handleRepoSelection`) — so nothing (status, "Working on…", session) is emitted before the gate. This also covers existing-thread follow-ups.
- Privacy is determined from `conversations.info.is_private` (fail-closed: only a confirmed public channel is allowed). `im`/`mpim` are private with no API call.
- The toggle is read from the control plane (`GET /integration-settings/slack`) via the service binding, cached ~60s in memory with a last-known-good KV fallback, so a previously-observed deny survives a control-plane outage and the bot doesn't storm a downed control plane.
- When the toggle is **on** (default), no `conversations.info` call and no guard work happen — behavior is unchanged with zero added cost.

### Why default-allow (note for maintainers)
Chosen for backward compatibility on upgrade. A coding agent reachable from private DMs is a data-exfiltration surface, so operators concerned about that should turn this off. The kill-switch is bounded, not absolute: a deny set *during* a control-plane outage, or on a cold isolate that never fetched, isn't enforced until a fetch succeeds (in-memory TTL + reconnect). If you'd prefer secure-by-default, it's a one-line default flip in `resolveAllowPrivateChannels`.

## Changes
- `shared`: `SlackGlobalSettings.allowPrivateChannels`, `DEFAULT_ALLOW_PRIVATE_CHANNELS`, shared `resolveAllowPrivateChannels` (`!== false`); `SlackChannelInfo.is_private`.
- `control-plane`: validate `allowPrivateChannels` (global-only boolean); surface it in `resolveSlackSettings`.
- `slack-bot`: new `integration-settings.ts` (cached reader + sticky last-known-good); `isPrivateMessageDispatchable` (im+mpim); `refuseIfPrivateContext` guard wired into the three entry points.
- `web`: toggle Switch + reset-dialog copy in the Slack settings panel.
- `docs/integrations/SLACK.md`: toggle behavior, required Slack scopes (`groups:read`/`im:read`/`mpim:read`), and the propagation SLA.

## Testing
- `npm run typecheck` — clean across all packages
- `npm test` — shared 169, control-plane 1170, slack-bot 101, web 264 — all passing
- New tests cover: setting validation/resolution, `isPrivateMessageDispatchable` (im/mpim/public), the gate end-to-end (private/public/im/mpim/fail-closed/toggle-on/`/interactions` back-door), and `getAllowPrivateChannels` cache + fail-closed behavior.

## Summary by CodeRabbit

* **New Features**
  * Workspace-wide toggle to control whether the Slack bot responds in private channels, group DMs, and direct messages (enabled by default). Toggle added to global Slack settings UI.

* **Behavior**
  * When off, the bot politely declines private-context interactions. Refusals are immediate and fail-closed if channel privacy can't be confirmed. Setting changes may take time to propagate.

* **Documentation**
  * Slack guide updated with configuration, propagation timing, scopes, and troubleshooting.

* **Tests**
  * Expanded coverage for private-channel/DM gating and related flows.


fredwanteeed:
Thanks for the review. Addressed in 46a04a7:

- **Required Slack scopes (valid):** corrected the docs — `im`/`mpim` are declined from the event's `channel_type` without a `conversations.info` call, so the gate needs only `channels:read` (public) and `groups:read` (private), not `im:read`/`mpim:read`.
- **Vacuous test assertion (valid):** added `expect(order).toContain("channelInfo")` before the `indexOf` ordering checks so a regression can't pass via `indexOf === -1`.
- **Cold-start outage re-fetch (nitpick):** left as-is, intentionally. Not caching the permissive default is a deliberate tradeoff so a freshly-set deny is picked up the moment the control plane recovers, rather than waiting out a cache TTL. Slack traffic is human-paced and this path only triggers during a full control-plane outage on a never-configured install, so the request rate is bounded in practice. Happy to add a short negative-cache or single-flight if you'd prefer to pre-empt it.

`npm run typecheck` clean; tests green (shared 169, control-plane 1170, slack-bot 101, web 264).

PR #693, opened 2026-05-31
feat: configurable Modal sandbox CPU and memory

vladgurovich:
## What

  Adds two optional fields to `SandboxSettings` — `cpuCores` and `memoryMib` — letting an operator reserve CPU and memory per sandbox. They map to Modal's  `cpu` / `memory` kwargs on `Sandbox.create`.
  
 ## Why

Sandboxes previously ran at Modal's ~0.125-core floor with bursting, with no way to guarantee resources for heavier install/build/test workloads. This is particularly important when trying to build/test complex applications in compiled languages such as Java in the sandbox.
  
  
## Behavior & design decisions

- **No default reservation.** Unset → the provider's own default applies (Modal's floor), so existing sandboxes see no cost change. Modal bills `max(reservation, actual usage)`, so a default would have raised the billed floor for every sandbox; I intentionally avoided that.
- **Positivity-only validation.** `cpuCores` must be a positive number (fractional allowed); `memoryMib` a positive integer. I do not impose arbitrary min/max ranges — Modal is the source of truth for real limits and rejects anything it can't honor at create time.
- **Advisory / provider-dependent.** Modal honors the fields; providers without resource reservations (Daytona) ignore them.

## What About Daytona

I havent used Daytona before, so i interrogated Claude and there is an eventual path to have Daytona respect cpu/memory config via hot-resize. The only wrinkle is that Daytona ties resources to the snapshot, so the cpu/memory config would be applied with a POST /sandbox/:id/resize AFTER create rather than at create time. CPU/Memory increases in Daytona are "hot" -- they work on a running sandbox with no stop.

## How settings flow

You set CPU/memory the same way as any other sandbox setting — globally or per-repo through the integration-settings API, which stores them in D1.

When a session starts, we resolve the effective settings (repo overrides on top of global defaults) and save the whole object on the session row as-is. The values aren't touched again until the sandbox actually spins up: at that point we re-validate them — the stored blob can come from older or untrusted writes, so the spawn path doesn't trust it — and pass anything still valid to Modal as  `cpu` / `memory`. 

Snapshot restores take the same path, so a resumed session gets the same reservation as a fresh one.

  ## Changes by layer

  | Layer | File | Change |
  |---|---|---|
  | Type | `shared/.../integrations.ts` | `cpuCores` / `memoryMib` on `SandboxSettings`, doc'd as advisory |
  | Validation | `control-plane/.../db/integration-settings.ts` | positive-number / positive-integer validators |
  | Resolved API | `control-plane/.../routes/integration-settings.ts` | exposes both (null when unset) |
  | Boundary | `control-plane/.../lifecycle/manager.ts` | `parseSandboxSettings` forwards positive values |
  | Data plane | `modal-infra/.../manager.py` | `_resource_kwargs` → Modal kwargs at both create sites |
  | UI | `web/.../sandbox-settings.tsx` | CPU/Memory inputs; "provider default" when blank |

## Inheritance guard

Per-repo saves only persist a resource value when it's global, explicitly edited, or already an override — so saving unrelated repo settings no longer pins an inherited global default as a repo override (mirrors the existing child-session-limit behavior). Covered by two regression tests.

## Testing

  - control-plane unit (validation, parse forward/drop) and integration
    (resolved-config defaults) — pass
  - web (render, save payload, positivity-block, inheritance) — pass
  - modal-infra (`_resource_kwargs` mapping + create/restore forwarding) — pass
  - lint, typecheck, ruff — clean

  ## Notes / out of scope

  - Daytona ignores the fields (no resource API); intentional.
  - Warm-pool path is dormant (sessions always create fresh) — not threaded.

## Summary by CodeRabbit

* **New Features**
  * Added CPU (fractional) and Memory (MiB, integer) sandbox resource reservations with a "Resources" editor; supports global defaults, per-repo overrides, and explicit null to indicate no override.
  * Resolved sandbox config now surfaces cpuCores and memoryMib (or null when not overridden).
  * Provider integration now forwards CPU/memory to sandbox creation when set.

* **Bug Fixes / Validation**
  * Stronger validation: CPU must be positive finite, Memory must be positive integer; invalid values are omitted or rejected and not persisted.

* **Tests**
  * Expanded unit and integration tests covering UI, validation, persistence, resolved config, and provider-forwarding.


packages/modal-infra/src/sandbox/manager.py
```diff
@@ -39,6 +39,30 @@
...
+
+    cpu_cores = settings.get("cpuCores")
+    if isinstance(cpu_cores, (int, float)) and not isinstance(cpu_cores, bool) and cpu_cores > 0:
+        kwargs["cpu"] = float(cpu_cores)
+
+    memory_mib = settings.get("memoryMib")
+    if isinstance(memory_mib, int) and not isinstance(memory_mib, bool) and memory_mib > 0:
+        kwargs["memory"] = memory_mib
+
+    return kwargs
```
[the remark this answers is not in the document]
vladgurovich: Addressed in a coming commit

packages/web/src/components/settings/sandbox-settings.tsx
```diff
@@ -86,10 +97,21 @@ function SandboxSettingsEditor({
       globalDefaults?.maxTotalChildSessions ??
       DEFAULT_MAX_TOTAL_CHILD_SESSIONS);
 
+  // Resource reservations have no app-level default — unset means "use the
+  // provider's default reservation", so resolve to an empty string, not a number.
+  const currentCpuCores: number | undefined = isGlobal
+    ? globalDefaults?.cpuCores
+    : (repoSettings?.cpuCores ?? globalDefaults?.cpuCores);
```
ColeMurray: [Automated review] This merges the stored repo override and inherited global default into a single current value, so repo scope can no longer distinguish “this repo explicitly overrides cpuCores” from “this repo is inheriting cpuCores.” That leaks into the save path below: clearing an inherited global value makes the form dirty, but blank values are omitted, so saving cannot express “use provider default for this repo” and the inherited value remains effective. Can we model the field provenance explicitly, either as stored override vs inherited effective value or with an explicit reset/null mode, before wiring the payload logic?

packages/modal-infra/src/sandbox/manager.py
```diff
@@ -39,6 +40,35 @@
 MAX_TUNNEL_PORTS = 10
 
 
+def _resource_kwargs(settings: dict[str, Any] | None) -> dict:
+    """Map sandbox settings to Modal resource kwargs.
+
+    `cpuCores` -> Modal `cpu` (cores, fractional allowed), `memoryMib` -> Modal
+    `memory` (MiB). Only set when explicitly configured and positive; absent or
+    invalid values are omitted so Modal applies its own default reservation.
```
ColeMurray: [Automated review] This makes the Modal edge own another copy of the resource validation policy, and it has different semantics from the settings store: the store rejects invalid values, the session parser silently drops invalid stored values, and this provider mapper silently drops them again. That contract is going to be hard to reason about as these fields evolve. Can we make the control-plane side expose one canonical normalize/validate path for `SandboxSettings`, then keep this helper focused on mapping already-normalized settings to Modal kwargs?

ColeMurray (dismissed):
see comments

packages/modal-infra/src/sandbox/manager.py
vladgurovich: resolved in https://github.com/ColeMurray/background-agents/pull/693/commits/b60a3c779eeb680b9eebda967e0bba6fb6214cd8

packages/web/src/components/settings/sandbox-settings.tsx
vladgurovich: resolved in https://github.com/ColeMurray/background-agents/pull/693/commits/b60a3c779eeb680b9eebda967e0bba6fb6214cd8

vladgurovich:
I addressed your feedback @ColeMurray, please take a look again 🙏

vladgurovich:
checking back in @ColeMurray -- anythng missing from this PR?

ColeMurray (approved):

PR #702, opened 2026-06-04
fix(slack-bot): show all repos in clarification picker via search

misnard:
## Problem

When the Slack bot can't determine which repository a message refers to, it posts a "Which repository should I work with?" picker. That picker was a `static_select` built from `repos.slice(0, 5)` — capped at **5 repos**. Worse, when the LLM classifier returned `alternatives`, the menu showed **only** those alternatives, so any other repo was unreachable. Users with more than a handful of repos couldn't select the one they wanted.

## Change

Replace the capped `static_select` with a searchable `external_select`, mirroring the existing App Home branch-override picker:

- The full repo list shows as soon as the menu opens (`min_query_length: 0`).
- Typing filters across **every** repo by full name.
- Results are capped at Slack's per-response ceiling (`MAX_REPO_SUGGESTION_OPTIONS = 100`), consistent with the App Home picker — so it scales past 100 repos via search.
- A new `block_suggestion` handler serves the options for the `select_repo` action; selection still flows through the existing `handleRepoSelection`, unchanged.
- Dropped the now-unused `repos.slice(0, 5)` fallback in the classifier's error path (there's no basis to suggest specific repos on a classification failure, and the picker now lets the user search the full list).

The classifier's `reasoning` is still shown in the message, so the bot's intent remains visible.

## Testing

- `npm run typecheck -w @open-inspect/slack-bot` ✅
- `npm test -w @open-inspect/slack-bot` ✅ (93 tests)
- `npm run lint -w @open-inspect/slack-bot` ✅

New tests cover the searchable picker returning the full list (capped at 100, well beyond the old 5) and filtering by the typed query.

## Summary by CodeRabbit

* **New Features**
  * Searchable repository picker in Slack clarification flows: users can type to filter and select from up to 100 repo options in real time.

* **Bug Fixes**
  * Removed unreliable repo suggestions from error responses when classification fails; clarifications now avoid presenting unsupported recommendations.

* **Tests**
  * Added tests verifying large-option handling and query-based filtering for the Slack repo picker, and updated expectations for failure responses.

### DIFF
--- a/packages/slack-bot/src/classifier/index.ts
+++ b/packages/slack-bot/src/classifier/index.ts
@@ -301,7 +301,9 @@ export class RepoClassifier {
         confidence: "low",
         reasoning:
           "Could not classify repository from structured model output. Please select a repository.",
-        alternatives: repos.slice(0, 5),
+        // No basis to suggest specific repos on a classification failure; the
+        // picker lets the user search the full list.
+        alternatives: undefined,
         needsClarification: true,
       };
     }
--- a/packages/slack-bot/src/index.test.ts
+++ b/packages/slack-bot/src/index.test.ts
@@ -1506,4 +1506,136 @@ describe("POST /interactions", () => {
       },
     ]);
   });
+
+  it("returns all repos (beyond the old 5-item limit) for the repo clarification picker", async () => {
+    const payload = {
+      type: "block_suggestion",
+      action_id: "select_repo",
+      user: { id: "U123" },
+      value: "",
+    };
+
+    const request = new Request("http://localhost/interactions", {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+        "x-slack-signature": "v0=test",
+        "x-slack-request-timestamp": `${Math.floor(Date.now() / 1000)}`,
+      },
+      body: new URLSearchParams({ payload: JSON.stringify(payload) }),
+    });
+
+    const env = makeEnv();
+    const repos = Array.from({ length: 150 }, (_, idx) => {
+      const number = String(idx + 1).padStart(3, "0");
+      return {
+        id: `acme/repo-${number}`,
+        owner: "acme",
+        name: `repo-${number}`,
+        fullName: `acme/repo-${number}`,
+        defaultBranch: "main",
+        private: true,
+      };
+    });
+
+    (env.CONTROL_PLANE.fetch as unknown as ReturnType<typeof vi.fn>).mockImplementation(
+      async (input: RequestInfo | URL) => {
+        const url = typeof input === "string" ? input : input.toString();
+        if (url.includes("/repos")) {
+          return new Response(JSON.stringify({ repos }), {
+            status: 200,
+            headers: { "Content-Type": "application/json" },
+          });
+        }
+
+        return new Response(JSON.stringify({ enabledModels: ["anthropic/claude-haiku-4-5"] }), {
+          status: 200,
+          headers: { "Content-Type": "application/json" },
+        });
+      }
+    );
+
+    const ctx = makeCtx();
+    const response = await app.fetch(request, env, ctx);
+
+    expect(response.status).toBe(200);
+    expect(ctx.waitUntil).not.toHaveBeenCalled();
+
+    const body = (await response.json()) as {
+      options: Array<{ text: { type: string; text: string }; value: string }>;
+    };
+    // Old behavior capped this at 5; new behavior shows the full list up to
+    // Slack's per-response ceiling.
+    expect(body.options).toHaveLength(100);
+    expect(body.options[0]).toEqual({
+      text: { type: "plain_text", text: "repo-001" },
+      description: { type: "plain_text", text: "repo-001" },
+      value: "acme/repo-001",
+    });
+  });
+
+  it("filters repo clarification suggestions by the typed query", async () => {
+    const payload = {
+      type: "block_suggestion",
+      action_id: "select_repo",
+      user: { id: "U123" },
+      value: "repo-150",
+    };
+
+    const request = new Request("http://localhost/interactions", {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+        "x-slack-signature": "v0=test",
+        "x-slack-request-timestamp": `${Math.floor(Date.now() / 1000)}`,
+      },
+      body: new URLSearchParams({ payload: JSON.stringify(payload) }),
+    });
+
+    const env = makeEnv();
+    const repos = Array.from({ length: 150 }, (_, idx) => {
+      const number = String(idx + 1).padStart(3, "0");
+      return {
+        id: `acme/repo-${number}`,
+        owner: "acme",
+        name: `repo-${number}`,
+        fullName: `acme/repo-${number}`,
+        defaultBranch: "main",
+        private: true,
+      };
+    });
+
+    (env.CONTROL_PLANE.fetch as unknown as ReturnType<typeof vi.fn>).mockImplementation(
+      async (input: RequestInfo | URL) => {
+        const url = typeof input === "string" ? input : input.toString();
+        if (url.includes("/repos")) {
+          return new Response(JSON.stringify({ repos }), {
+            status: 200,
+            headers: { "Content-Type": "application/json" },
+          });
+        }
+
+        return new Response(JSON.stringify({ enabledModels: ["anthropic/claude-haiku-4-5"] }), {
+          status: 200,
+          headers: { "Content-Type": "application/json" },
+        });
+      }
+    );
+
+    const ctx = makeCtx();
+    const response = await app.fetch(request, env, ctx);
+
+    expect(response.status).toBe(200);
+
+    const body = (await response.json()) as {
+      options: Array<{ text: { type: string; text: string }; value: string }>;
+    };
+    expect(body.options).toEqual([
+      {
+        text: { type: "plain_text", text: "repo-150" },
+        description: { type: "plain_text", text: "repo-150" },
+        value: "acme/repo-150",
+      },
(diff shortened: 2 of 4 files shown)


packages/slack-bot/src/index.ts
```diff
@@ -33,12 +33,42 @@ import { createKvCacheStore } from "@open-inspect/shared";
...
+const SELECT_REPO_ACTION_ID = "select_repo";
+
+/**
+ * Repo options for the clarification picker's external_select. Slack queries
+ * this endpoint as the user types; we filter on the repo's full name and cap
+ * the count at Slack's per-response limit. With min_query_length 0 the
+ * unfiltered list shows as soon as the menu opens, and typing surfaces any of
+ * the remaining repos — so users are no longer limited to a fixed handful.
+ */
+async function getRepoClarificationOptions(env: Env, query: string | undefined, traceId?: string) {
```
ColeMurray, line 58: [Automated Review] **Wrong layer + the guard only covers one of two identical pickers.** `index.ts` goes 1216 → 1260 lines (already well past 1k), and this builder's near-twin `getRepoBranchSuggestionOptions` already lives in `app-home/`. Co-locate them (or a shared `suggestions.ts`) instead of growing the monster file.

This matters beyond tidiness: the `try/catch` added around this lookup guards *only* the clarification picker. App Home's identical `getRepoBranchSuggestionOptions` lookup has no such guard and still 500s the `/interactions` endpoint on a repos-fetch failure. Sharing the handler puts the guard in one place and protects both pickers.

packages/slack-bot/src/index.ts
```diff
@@ -33,12 +33,42 @@ import { createKvCacheStore } from "@open-inspect/shared";
...
+  const repos = await getAvailableRepos(env, traceId);
+  const normalizedQuery = query?.trim().toLowerCase();
```
ColeMurray, line 60: [Automated Review] **Duplicated filter logic — extract the canonical helper.** This normalize-and-filter block is verbatim in `app-home/interactions.ts` (`getRepoBranchSuggestionOptions`). It's repo-domain logic, and `classifier/repos.ts` already owns the repo finders (`getRepoByFullName`, `getRepoById`, `getReposByChannel`). Extract once and reuse in both call sites:

```ts
// classifier/repos.ts
/** Case-insensitive filter of repos by free-text query against full name. */
export function filterReposByQuery(repos: RepoConfig[], query: string | undefined): RepoConfig[] {
  const normalized = query?.trim().toLowerCase();
  if (!normalized) return repos;
  return repos.filter((repo) => repo.fullName.toLowerCase().includes(normalized));
}
```

Both call sites then collapse to `filterReposByQuery(repos, query)` — one definition, no drift.

packages/slack-bot/src/index.ts
```diff
@@ -33,12 +33,42 @@ import { createKvCacheStore } from "@open-inspect/shared";
...
+  const filtered = normalizedQuery
+    ? repos.filter((repo) => repo.fullName.toLowerCase().includes(normalizedQuery))
+    : repos;
+
+  return filtered.slice(0, MAX_REPO_SUGGESTION_OPTIONS).map((repo) => ({
+    text: { type: "plain_text" as const, text: repo.displayName },
+    description: { type: "plain_text" as const, text: repo.description.slice(0, 75) },
```
ColeMurray, line 67: [Automated Review] **Re-implements an existing truncation helper.** `repo.description.slice(0, 75)` duplicates the magic `75` and the intent of `truncateSelectOptionText` / `SELECT_OPTION_TEXT_LIMIT` in `app-home/view.ts` (and drops its ellipsis). Worse, the `text` field on the line above (`repo.displayName`) isn't length-guarded at all — a repo name over Slack's 75-char `plain_text` limit would be rejected outright. Route both fields through the canonical `truncateSelectOptionText`. (Pre-existing in the deleted code, but this PR is reworking the exact mapping, so it's the natural place to fix it.)

packages/slack-bot/src/index.ts
```diff
@@ -606,6 +636,32 @@ app.post("/interactions", async (c) => {
   }
 
   if (payload.type === "block_suggestion") {
+    if (payload.action_id === SELECT_REPO_ACTION_ID) {
+      let options: Awaited<ReturnType<typeof getRepoClarificationOptions>> = [];
```
ColeMurray, line 640: [Automated Review] **`Awaited<ReturnType<typeof …>>` is a type smell — name the shape.** This reaches for the function's return type because the option shape was never named. The codebase already has `SlackSelectOption`; the only thing this picker adds is `description`, which Slack option objects support generally:

```ts
// app-home/slack-types.ts
export type SlackSelectOption = { text: SlackPlainText; description?: SlackPlainText; value: string };
```

Then `getRepoClarificationOptions` returns `SlackSelectOption[]` explicitly (App Home's builder is unchanged — it just omits `description`), and this line becomes a plain `const options: SlackSelectOption[]`. See the next comment for dropping the `let`/reassign entirely.

packages/slack-bot/src/index.ts
```diff
@@ -606,6 +636,32 @@ app.post("/interactions", async (c) => {
...
+        // Don't let a lookup failure surface as a 500 on /interactions — return
+        // an empty suggestions payload so Slack just shows no matches.
+        log.error("slack.repo_clarification_options", {
+          trace_id: traceId,
+          query: payload.value,
+          error: e instanceof Error ? e : new Error(String(e)),
+          duration_ms: Date.now() - startTime,
+        });
+      }
+      log.info("http.request", {
```
ColeMurray, line 653: [Automated Review] **Inline-duplicated `http.request` log + divergent structure.** The App Home path returns `{ body, logContext }` and logs once centrally; this branch re-implements the whole `http.request` field set inline, while the fallback `return c.json({ options: [] })` below logs *nothing* — three behaviors for one payload type. Collapse to a single branch that computes options for every sub-case, then logs once:

```ts
if (payload.type === "block_suggestion") {
  const options =
    payload.action_id === SELECT_REPO_ACTION_ID
      ? await getRepoClarificationOptions(c.env, payload.value, traceId).catch((e) => {
          log.error("slack.repo_clarification_options", {
            trace_id: traceId,
            query: payload.value,
            error: e instanceof Error ? e : new Error(String(e)),
          });
          return [];
        })
      : [];
  log.info("http.request", {
    trace_id: traceId, http_method: "POST", http_path: "/interactions", http_status: 200,
    interaction_type: payload.type, action_id: payload.action_id,
    option_count: options.length, duration_ms: Date.now() - startTime,
  });
  return c.json({ options });
}
```

This drops the `let` + `Awaited<ReturnType>`, removes the duplicated log statement, and closes the fallback's missing-log gap.

packages/slack-bot/src/index.test.ts
```diff
@@ -1506,4 +1506,136 @@ describe("POST /interactions", () => {
...
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+        "x-slack-signature": "v0=test",
+        "x-slack-request-timestamp": `${Math.floor(Date.now() / 1000)}`,
+      },
+      body: new URLSearchParams({ payload: JSON.stringify(payload) }),
+    });
+
+    const env = makeEnv();
+    const repos = Array.from({ length: 150 }, (_, idx) => {
```
ColeMurray, line 1529: [Automated Review] **Triplicated test scaffolding.** This 150-repo builder plus the `CONTROL_PLANE.fetch` mock implementation is now copy-pasted across the existing App Home suggestion test and both new tests. Extract `buildNumberedRepos(n)` and `mockReposFetch(env, repos)` helpers so each test shrinks to its actual assertion.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -301,7 +301,9 @@ export class RepoClassifier {
         confidence: "low",
         reasoning:
           "Could not classify repository from structured model output. Please select a repository.",
-        alternatives: repos.slice(0, 5),
+        // No basis to suggest specific repos on a classification failure; the
+        // picker lets the user search the full list.
+        alternatives: undefined,
```
ColeMurray, line 306: [Automated Review] _(minor, not a blocker)_ Setting `undefined` here is correct and consistent with the happy path. Worth noting: after this PR, slack-bot no longer **reads** `ClassificationResult.alternatives` anywhere — the sole reader was the deleted `result.alternatives || repos.slice(0, 5)`. The field stays because it's a shared type still consumed by linear-bot (`webhook-handler.ts`), but slack-bot's happy path now populates it write-only; a one-line comment would save the next reader a phantom-consumer hunt.

Separately: `linear-bot/classifier/index.ts` still has the `alternatives: repos.slice(0, 5)` clarification cap this PR removes here — out of scope, but the two classifiers are drifting and it's worth a follow-up.

ColeMurray (commented):
## [Automated Review] Thermo-nuclear code-quality review

**The core design is right** — swapping the capped `static_select` for an `external_select` backed by a `block_suggestion` endpoint is the correct Slack pattern, it mirrors the App Home branch picker faithfully, and guarding the lookup so it returns `{ options: [] }` instead of a 500 is a genuine robustness win.

The concern is **how** it's built: the change copy-pastes the existing App Home picker machinery (filter logic, option shape, suggestion handler) rather than sharing it, lands the new builder in the already-1260-line `index.ts` away from its sibling, and adds an `Awaited<ReturnType<…>>` type smell plus a duplicated log path. None are behavioral bugs, but they leave the codebase messier than the change requires.

Findings 1–5 are really one coherent move: the clarification picker and the App Home picker are the *same feature* (filter repos by query → cap at `MAX_REPO_SUGGESTION_OPTIONS` → map to options). Share the filter, share the option type, share the guarded handler — and the duplication, the type smell, the divergent log, and the file-bloat all dissolve together. Specifics are in the inline comments.

ColeMurray (approved):

PR #719, opened 2026-06-09
Fail empty OpenCode prompt completions in bridge

### LINKED ISSUE #718: Bridge reports success when OpenCode completes without assistant output
## Summary

The sandbox bridge can report `execution_complete success=true` even when OpenCode emitted no assistant output. This leaves the web app showing a connected, ready sandbox and a completed prompt with no LLM response.

There are two related bridge assumptions involved:

1. The bridge sends `messageID` to OpenCode `prompt_async` and assumes OpenCode will persist that exact user message ID.
2. `_handle_prompt` treats a prompt as successful unless an explicit `error` event was yielded, even if the stream yielded no `token`, `tool_call`, or `step_finish` events.

## Production evidence

Observed in a Vercel-backed production deployment using OpenCode `1.14.41`:

- Web UI showed `Connected`, sandbox ready, and `Execution complete`.
- Control plane events for the reproduced session were only:
  - `ready`
  - `user_message`
  - `execution_complete` with `success: true`
- No `token` events were emitted.
- Cloudflare logs showed the prompt was dispatched to the sandbox successfully and the sandbox reported prompt success.
- Vercel sandbox creation and websocket connection were healthy.

Inspecting the saved sandbox snapshot's OpenCode DB showed:

- A user message row was created with an OpenCode-generated ID, not the `messageID` supplied by the bridge.
- An assistant message row was created with:
  - `cost: 0`
  - `tokens.input: 0`
  - `tokens.output: 0`
  - `tokens.reasoning: 0`
- The `part` table contained only the user text part. There were no assistant text parts.

That means the bridge had no assistant content to forward, but still marked the control-plane message complete successfully.

## Why this matters

This is the same failure shape described in the follow-up note from #630 / #567: silent or empty OpenCode streams can currently become successful completed prompts in the UI. Because `packages/sandbox-runtime` is provider-agnostic, this can affect both Modal and Vercel deployments when OpenCode returns no assistant parts or when message correlation misses the assistant message.

## Expected behavior

- If OpenCode emits an assistant response, bridge forwards it even if OpenCode regenerated the user message ID.
- If OpenCode completes/idles without any assistant output or terminal assistant step, bridge reports `execution_complete success=false` with a clear error.

## Proposed fix

- Track actual OpenCode user message IDs from `message.updated` events where `role == "user"`.
- Accept assistant messages whose `parentID` matches either the bridge-generated `messageID` or the actual OpenCode user message ID discovered from the stream.
- Track whether `_handle_prompt` emitted any prompt output (`token`, `tool_call`, or `step_finish`). If not, fail the execution with a clear `OpenCode completed without emitting assistant output.` error instead of reporting success.

## Regression tests to add

- A stream where OpenCode emits a user `message.updated` with a regenerated ID, then an assistant parented to that actual ID. The bridge should forward the assistant text.
- A stream that connects and idles with no prompt output. `_handle_prompt` should send `execution_complete success=false`.

bath:
## Summary

Fixes #718.

This hardens the sandbox bridge against OpenCode prompt completions that produce no assistant output, and against OpenCode regenerating the user message ID instead of preserving the `messageID` supplied to `prompt_async`.

Changes:

- Track actual OpenCode user message IDs from `message.updated` events where `role == "user"`.
- Accept assistant messages whose `parentID` matches either the generated bridge ID or an actual user message ID observed in the stream.
- Treat a prompt stream with no `token`, `tool_call`, or `step_finish` as a failed execution instead of reporting `execution_complete success=true`.
- Add regression coverage for regenerated user message IDs and empty prompt completions.

## Evidence

This was reproduced in a production Vercel-backed Open-Inspect deployment using OpenCode `1.14.41`.

Observed control-plane event sequence for the reproduced session:

```text
ready
user_message
execution_complete success=true
```

No `token` event was emitted, while the web app showed the sandbox as connected/ready and the prompt as complete.

Cloudflare logs showed the control plane dispatched the prompt successfully and the sandbox reported success. Vercel sandbox creation and the sandbox websocket were healthy, so the failure was inside the sandbox runtime/OpenCode bridge path.

Inspecting the saved OpenCode DB from the sandbox snapshot showed:

- OpenCode created a user message with its own generated ID, not the bridge-supplied `messageID`.
- OpenCode created an assistant message with `cost: 0` and all token counts at zero.
- The `part` table had only the user text part; there were no assistant text parts.

That explains why the bridge had nothing to forward, but the current `_handle_prompt` still sent a successful `execution_complete`: it only considered explicit `error` events as failures.

This is the failure mode called out as a follow-up in #630 / #567: silent or empty OpenCode streams should fail loudly rather than become successful no-reply prompts.

## Why this fix works

The bridge needs two independent guards:

1. Correlation should follow the actual OpenCode message graph. If OpenCode emits a user `message.updated` with a generated ID, assistant messages parented to that real user ID belong to the current prompt and should not be filtered out.
2. Completion success should require evidence of prompt output. If the stream reaches idle/return without any assistant token, tool call, or step finish, the UI should receive a failed `execution_complete` with a clear error.

This keeps the existing parent-based filtering for unrelated prompts, while making it robust to OpenCode regenerating the user message ID.

## Test plan

- [x] `python3 -m compileall packages/sandbox-runtime/src/sandbox_runtime/bridge.py`
- [x] `git diff --check`
- [ ] `python3 -m pytest packages/sandbox-runtime/tests/test_bridge_sse.py packages/sandbox-runtime/tests/test_bridge_message_tracking.py` could not be run in my local environment because the available Python is `3.9.6` and `pytest` is not installed. The package requires Python `>=3.12`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for prompts that produce no output—now properly reported as failures.
  * Enhanced message correlation in streaming mode to reliably track and link assistant responses to their source messages, especially when message IDs change during processing.

### DIFF
--- a/packages/sandbox-runtime/src/sandbox_runtime/bridge.py
+++ b/packages/sandbox-runtime/src/sandbox_runtime/bridge.py
@@ -625,13 +625,26 @@ class AgentBridge:
 
             had_error = False
             error_message = None
+            emitted_output = False
             async for event in self._stream_opencode_response_sse(
                 message_id, content, model, reasoning_effort
             ):
                 if event.get("type") == "error":
                     had_error = True
                     error_message = event.get("error")
+                elif event.get("type") in ("token", "tool_call", "step_finish"):
+                    emitted_output = True
                 await self._send_event(event)
+
+            if not had_error and not emitted_output:
+                had_error = True
+                error_message = "OpenCode completed without emitting assistant output."
+                self.log.error(
+                    "prompt.no_output",
+                    message_id=message_id,
+                    model=model,
+                    reasoning_effort=reasoning_effort,
+                )
 
             if had_error:
                 outcome = "error"
@@ -942,6 +955,7 @@ class AgentBridge:
         cumulative_text: dict[str, str] = {}
         emitted_tool_states: set[str] = set()
         allowed_assistant_msg_ids: set[str] = set()
+        user_message_ids: set[str] = {opencode_message_id}
         pending_parts: dict[str, list[tuple[dict[str, Any], Any]]] = {}
         pending_parts_total = 0
         pending_drop_logged = False
@@ -1114,7 +1128,16 @@ class AgentBridge:
                                         role = info.get("role", "")
                                         finish = info.get("finish", "")
 
-                                        parent_matches = parent_id == opencode_message_id
+                                        if role == "user" and oc_msg_id:
+                                            if oc_msg_id not in user_message_ids:
+                                                self.log.info(
+                                                    "bridge.user_message_id_discovered",
+                                                    expected_id=opencode_message_id,
+                                                    actual_id=oc_msg_id,
+                                                )
+                                            user_message_ids.add(oc_msg_id)
+
+                                        parent_matches = parent_id in user_message_ids
                                         is_compaction_summary = info.get("summary") is True
 
                                         self.log.debug(
@@ -1212,6 +1235,7 @@ class AgentBridge:
                                             opencode_message_id,
                                             cumulative_text,
                                             allowed_assistant_msg_ids,
+                                            user_message_ids=user_message_ids,
                                             compaction_occurred=compaction_occurred,
                                         ):
                                             yield final_event
@@ -1236,6 +1260,7 @@ class AgentBridge:
                                             opencode_message_id,
                                             cumulative_text,
                                             allowed_assistant_msg_ids,
+                                            user_message_ids=user_message_ids,
                                             compaction_occurred=compaction_occurred,
                                         ):
                                             yield final_event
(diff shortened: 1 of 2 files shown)


bath:
Added a second production repro to #718 using `openai/gpt-5.2` in session `5d45271a55775d05d22fca0818cbcaef`.

Key point for this PR: the live Vercel sandbox had the OpenAI OAuth plugin/auth state present, but OpenCode still persisted an assistant message with zero cost/tokens and no assistant `part` rows. The currently deployed bridge lacks this PR's no-output guard, so it emitted `execution_complete success=true` with no tokens. This PR would convert that exact result into a failed completion with an explicit error.

PR #734, opened 2026-06-12
fix(control-plane): use full namespace path as GitLab repo owner

duboff:
## Summary

- The GitLab provider mapped `namespace.path` — the **last** path segment only — into the repository `owner` in `getRepository`, `checkRepositoryAccess`, and `listRepositories`. For nested-group projects (`group/subgroup/project`) this produced owner=`subgroup`, so owner/name pairs handed to consumers couldn't be resolved back to a project: lookups reconstructed `subgroup/project` and GitLab returned 404 (surfaced as "Repository is not installed for the GitHub App" at session creation).
- Switch to `namespace.full_path`, which carries the whole group hierarchy and is identical to `namespace.path` for top-level groups, so flat-namespace deployments are unaffected.

Related: #733 fixes the same nested-group class of bug on the linear-bot side.

## Test plan

- [x] New unit tests asserting `owner` is the full namespace path for nested-group projects across all three methods
- [x] Existing mocks extended with `full_path` (always present in real GitLab API responses)
- [x] `npm test -w @open-inspect/control-plane` — 1291 tests pass
- [x] `tsc --noEmit` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitLab support for nested-group projects by using the complete group path for repository ownership identification, ensuring repositories nested within multiple group levels are correctly handled and accessible.

### DIFF
--- a/packages/control-plane/src/source-control/providers/gitlab-provider.test.ts
+++ b/packages/control-plane/src/source-control/providers/gitlab-provider.test.ts
@@ -37,7 +37,7 @@ describe("GitLabSourceControlProvider", () => {
           name: "My Web App", // display name — should NOT be used
           path: "web", // URL slug — should be used as name
           path_with_namespace: "acme/web",
-          namespace: { path: "acme" },
+          namespace: { path: "acme", full_path: "acme" },
           default_branch: "main",
           visibility: "private",
         })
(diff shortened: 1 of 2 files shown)


ColeMurray:
resolved conflicts in https://github.com/ColeMurray/background-agents/pull/883 and merged

PR #738, opened 2026-06-12
feat: add deepseek model options

cygaar:
## Summary

  - Add DeepSeek as a supported model provider
  - Add DeepSeek V4 Flash and V4 Pro model options
  - Wire DeepSeek model metadata through the shared model definitions
  - Document the required `DEEPSEEK_API_KEY` secret

## Testing

  - `npm test -w @open-inspect/control-plane -- src/utils/models.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Added support for DeepSeek models as opt-in options; users must set `DEEPSEEK_API_KEY` environment variable.
  * Two new DeepSeek model variants now available for selection.

* **Documentation**
  * Updated configuration and troubleshooting guides to cover DeepSeek setup and API key requirements across all deployment environments.


ColeMurray:
Thanks for this! The model additions look good — I confirmed `deepseek-v4-flash` / `deepseek-v4-pro` resolve via the `deepseek` provider in the models.dev catalog (`api.deepseek.com`, keyed on `DEEPSEEK_API_KEY`), so the data-only wiring is complete and nothing extra is needed on the runtime side.

I pushed a small follow-up commit (`83db6a17`) directly to the branch with three tweaks:

- **Reverted the `auth.ts` `debug` change** — it looked unrelated to the DeepSeek work, so it's probably better as its own PR. Happy to take it separately if it was intentional.
- **`HOW_IT_WORKS.md`** — kept your DeepSeek note but split it into its own callout and restored the "Modal injects these automatically" line, since the combined version read as contradictory under the **Daytona and Vercel users** header.
- **`SECRETS.md`** — kept `ANTHROPIC_API_KEY` as the global-secret example rather than swapping it for `DEEPSEEK_API_KEY` (Anthropic is still the default provider).

Also updated the PR description (the models are V4 Flash / V4 Pro rather than Chat / Reasoner). Nothing blocking from my side after this — thanks again!

ColeMurray (approved):

PR #740, opened 2026-06-12
feat: make model api key settings explicit

cygaar:
## Summary

Makes the API key settings for models explicit (similar to how it's setup in Cursor), to make it easier to configure them.

## Changes

- Adds fixed inputs for:
    - ANTHROPIC_API_KEY
    - OPENAI_API_KEY
    - OPENCODE_API_KEY

- Supports both global and per-repo scopes using the same encrypted secrets APIs as the existing Secrets settings.
- Shows whether each key is Set, Inherited, or Not set.
- Allows repo-specific keys to override global keys.
- Adds save/remove actions for direct keys.
- Adds focused tests for global save, repo override save, inherited status, and repo key deletion.

## Testing

- npm test -w @open-inspect/web -- model-api-keys-settings.test.tsx
- npm run typecheck -w @open-inspect/web
- npm run lint -w @open-inspect/web

## Screenshot

Screenshot 2026-06-13 at 1 44 34 PM

## Summary by CodeRabbit

* **New Features**
  * Added a Model API Keys settings UI to manage provider keys (OpenAI, Anthropic, OpenCode Zen) with global and repository-scoped modes, clear status labels, per-key remove, validation, and save notifications.
  * Integrated the API keys UI into the Models settings page; loading state now preserves page context while keys load.

* **Tests**
  * Added a comprehensive test suite covering save, remove, repo/global behavior, and UI states.

### DIFF
--- a/packages/web/src/components/settings/model-api-keys-settings.test.tsx
+++ b/packages/web/src/components/settings/model-api-keys-settings.test.tsx
@@ -1,0 +1,185 @@
+// @vitest-environment jsdom
+/// <reference types="@testing-library/jest-dom" />
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { cleanup, render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import * as matchers from "@testing-library/jest-dom/matchers";
+import { SWRConfig, mutate as globalMutate } from "swr";
+import { ModelApiKeysSettings } from "./model-api-keys-settings";
+
+expect.extend(matchers);
+
+const { reposMock, toastMock } = vi.hoisted(() => ({
+  reposMock: {
+    repos: [
+      {
+        id: 1,
+        fullName: "acme/app",
+        owner: "acme",
+        name: "app",
+        description: null,
+        private: true,
+        defaultBranch: "main",
+      },
+    ],
+    loading: false,
+  },
+  toastMock: {
+    success: vi.fn(),
+    error: vi.fn(),
+  },
+}));
+
+vi.mock("@/hooks/use-repos", () => ({
+  useRepos: () => ({ repos: reposMock.repos, loading: reposMock.loading }),
+}));
+
+vi.mock("sonner", () => ({
+  toast: toastMock,
+}));
+
+beforeEach(() => {
+  Element.prototype.scrollIntoView = vi.fn();
+});
+
+function jsonResponse(body: unknown, status = 200) {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { "Content-Type": "application/json" },
+  });
+}
+
+function renderWithSWR(fallback: Record<string, unknown> = {}) {
+  const fetchMock = vi.fn(async (_input: RequestInfo | URL, init?: RequestInit) => {
+    if (init?.method === "PUT" || init?.method === "DELETE") {
+      return jsonResponse({ status: "ok" });
+    }
+    throw new Error("unexpected fetch");
+  });
+  vi.stubGlobal("fetch", fetchMock);
+
+  const result = render(
+    <SWRConfig
+      value={{
+        provider: () => new Map(),
+        fallback,
+        dedupingInterval: Infinity,
+        revalidateOnFocus: false,
+        revalidateIfStale: false,
+        revalidateOnReconnect: false,
+      }}
+    >
+      <ModelApiKeysSettings />
+    </SWRConfig>
+  );
+
+  return { ...result, fetchMock };
+}
+
+async function selectRepo() {
+  const user = userEvent.setup();
+  await user.click(screen.getByRole("button", { name: /All Repositories/ }));
+  await user.click(screen.getByRole("option", { name: /app/ }));
+}
+
+afterEach(async () => {
+  cleanup();
+  await globalMutate(() => true, undefined, { revalidate: false });
+  vi.restoreAllMocks();
+  reposMock.loading = false;
+  toastMock.success.mockReset();
+  toastMock.error.mockReset();
+});
+
+describe("ModelApiKeysSettings", () => {
+  it("saves model API keys to global secrets by default", async () => {
+    const user = userEvent.setup();
+    const { fetchMock } = renderWithSWR({
+      "/api/secrets": { secrets: [] },
+    });
+
+    await user.type(screen.getByLabelText("Anthropic API key"), "sk-ant-test");
+    await user.type(screen.getByLabelText("OpenAI API key"), "sk-openai-test");
+    await user.click(screen.getByRole("button", { name: "Save API Keys" }));
+
+    await waitFor(() => {
+      expect(fetchMock).toHaveBeenCalledWith(
+        "/api/secrets",
+        expect.objectContaining({
+          method: "PUT",
+          body: JSON.stringify({
+            secrets: {
+              ANTHROPIC_API_KEY: "sk-ant-test",
+              OPENAI_API_KEY: "sk-openai-test",
+            },
+          }),
+        })
+      );
+    });
+  });
+
+  it("shows repo keys as set and global keys as inherited", async () => {
+    renderWithSWR({
+      "/api/secrets": { secrets: [] },
+      "/api/repos/acme/app/secrets": {
+        secrets: [{ key: "OPENAI_API_KEY" }],
+        globalSecrets: [{ key: "ANTHROPIC_API_KEY" }],
+      },
+    });
+
+    await selectRepo();
+
+    expect(screen.getByText("Inherited")).toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


ColeMurray:
@cygaar can you include a screenshot/recording of the feature?

cygaar:
@ColeMurray added a screenshot

ColeMurray:
[Automated Review]

## Summary

- **PR:** #740 — *feat: make model api key settings explicit* · **Author:** @cygaar · **Base:** `main` ← `modeloptions`
- **Size:** 3 files, +590 / −60 (one new component, one new test file, one integration edit)
- **What it does:** Adds a dedicated "Model API Keys" UI (for `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `OPENCODE_API_KEY`) embedded at the bottom of the **Models** settings tab. It writes to the *existing* encrypted-secrets endpoints, supports global + per-repo scope, shows Set/Inherited/Not-set status, and adds save/per-key-delete.

**Overall:** This is a clean, secure, well-tested component, and the underlying behavior is correct end-to-end (verified against the backend — see below). My concerns are not bugs; they are two **design-fit** issues: (1) it forks the existing secrets logic rather than sharing it, and (2) it presents `ANTHROPIC_API_KEY` as a uniform settable key, which is misleading on Modal (the default provider). There are **no correctness or security blockers**.

---

## Design & Fit Assessment

I traced the whole data path before judging, so this is grounded rather than speculative.

**✅ The idea is coherent with the system.** The three keys map *exactly* onto the three model providers in the shared catalog (`packages/shared/src/models.ts`: `anthropic/*`, `openai/*`, and the `opencode/*` "OpenCode Zen" group at lines 164–171). All three are accepted by the secrets API — `ANTHROPIC`/`OPENAI` were deliberately un-reserved in commit `bc2482a4` (PR #88, *"allow LLM API keys to be set as repo secrets"*), and `OPENCODE_API_KEY` was never reserved (`secrets-validation.ts:7-25`). A user-set key reaches the OpenCode runtime via process-env inheritance (`entrypoint.py:792-856`) and *is used* on all three providers. So the curated, Cursor-style "explicit keys" framing is legitimate — on **Daytona and Vercel it is the *only* credential source** for Anthropic (`docs/SECRETS.md:116`), so surfacing it explicitly has real value.

**⚠️ Concern 1 — It contradicts the provider-nuance the system already maintains (the sharpest fit issue).** On **Modal — the default provider** — `ANTHROPIC_API_KEY` is *structurally required and auto-injected* by the platform via a separate Modal Secret (`app.py:42-47`, `required_keys=["ANTHROPIC_API_KEY"]`; attached at `manager.py:409-417`). The docs deliberately tell users to set `ANTHROPIC_API_KEY` **only for Daytona/Vercel** because "Modal injects this automatically" (`docs/SECRETS.md:39`, `docs/HOW_IT_WORKS.md:489-490`). This new UI erases that nuance:
- For a Modal user with a working agent, the Anthropic row renders **"Not set"** — because the badge reflects only the *user-secret store*, not the *effective runtime credential*. That's a misleading prompt to enter a key that isn't needed.
- If they do enter one, it collides with the platform key. Tracing shows the user value *should* win (Modal layers explicit `env=` over `secrets=[]`), so it likely "works" — but **no test in the repo pins Secret-vs-env precedence** (only env-vs-env is tested, `test_sandbox_env_vars.py:40,88`). So the happy path here rests on Modal SDK semantics, not on anything this codebase guarantees.
- Net: the feature is provider-*agnostic* in presentation but provider-*dependent* in reality, and the UI conveys none of it. At minimum the Anthropic row could be annotated/disabled on Modal (or the status could distinguish "provided by platform").

**⚠️ Concern 2 — It forks the existing secrets implementation instead of sharing it.** A generic secrets UI already exists: `secrets-settings.tsx` (scope + repo combobox) + `secrets-editor.tsx` (the SWR/validation/PUT/DELETE engine, already parameterized by `scope`). The new component re-implements the *same logic*: the `GLOBAL_SCOPE = "__global__"` sentinel, the `apiBase` derivation, `getUtf8Size`, the PUT `{ secrets }` body, the per-key DELETE + `mutate`, the repo Combobox config, and the size constants `MAX_VALUE_SIZE = 16384` / `MAX_TOTAL_VALUE_SIZE = 65536`. Those two constants are **already triple-sourced** (server `secrets-validation.ts:3-4`, client `secrets-editor.tsx:14-15`, `docs/SECRETS.md`) — this PR adds a **fourth** copy. The *presentational* fork (fixed provider rows vs. free-form key/value) is defensible; the *logic* fork is not. There's no `useSecrets` hook today, so the clean fix is to extract the scope→apiBase + save/delete/validate logic into a shared hook (and import the byte limits from one source — ideally lifting them into `@open-inspect/shared`).

**Minor fit note:** the same secret is now editable from two tabs (the generic **Secrets** tab and the new **Models** section). They're views over one store so it's consistent, but it's a second surface for the same data.

---

## Critical Issues

None. No correctness, security, or data-integrity blockers — the frontend↔backend contract (`globalSecrets` in the repo response, PUT body shape, per-key DELETE routes, auth) all check out against the live handlers (`routes/secrets.ts`).

---

## Suggestions

- **[Architecture] (strong)** `model-api-keys-settings.tsx` — Extract the duplicated secrets logic (scope→`apiBase`, `getUtf8Size`, size limits, save/delete/`mutate`) into a shared hook reused by both this component and `secrets-editor.tsx`, rather than maintaining a parallel copy. At minimum, import `MAX_VALUE_SIZE`/`MAX_TOTAL_VALUE_SIZE` from a single source instead of re-declaring them (`:13-14`). Rationale: avoids the now-fourth copy of the byte limits and prevents the two secrets UIs from drifting.

- **[Design]** `model-api-keys-settings.tsx` — Make the Anthropic row provider-aware. On Modal it's platform-injected, so "Not set" is misleading and an entered value collides with the platform key. Consider annotating/disabling it for Modal sessions, or reflecting "provided by platform" in the status. This preserves the guidance `docs/SECRETS.md` already encodes.

- **[Testing]** `model-api-keys-settings.test.tsx` — Tests stub `fetch` entirely, so they validate request construction and status rendering but give no signal on the real backend contract (fine for unit scope — the contract was verified manually here). Consider adding cases for the **error toasts** (non-OK save/delete) and the **"empty field is omitted from the payload"** behavior, which is core to the "leave empty to keep current value" semantics and currently untested.

- **[Docs]** `docs/SECRETS.md` / `docs/HOW_IT_WORKS.md` aren't updated. Since they contain the provider-specific Anthropic guidance this UI interacts with, a short note pointing at the new UI (and the Modal caveat) would keep them in sync.

---

## Nitpicks

- Nit: `model-api-keys-settings.tsx` `handleSave`/`handleDelete` — Save is disabled on `saving` but not on `deletingKey`, so Save can fire mid-delete. Both hit the same endpoint and re-`mutate`, so impact is low, but disabling Save while a delete is in flight would tidy it up.
- Nit: The "Repository" `<label>` isn't associated with the Combobox (no `htmlFor`/`id`); minor a11y gap. (The provider inputs correctly use `aria-label`.)
- Nit: Remove triggers an immediate destructive DELETE with no confirmation. Acceptable, but worth confirming it matches the existing `SecretsEditor` UX.

---

## Positive Feedback

- **Security posture is excellent:** `type="password"` inputs, secret values never re-fetched/redisplayed ("Saved values are never displayed again"), only metadata read, and the `.toUpperCase()` comparisons correctly mirror the server's key normalization.
- **Solid component craft:** clean loading/error/ready states, SWR usage consistent with the rest of the app, good `aria-label`s that the tests assert against, and the "only send changed (non-empty) keys" upsert semantics are correct against the backend's `setSecrets` behavior.
- **The `models-settings.tsx` refactor** (early-return → inline conditional) is a genuine improvement — the heading and the new section now stay visible while model preferences load.

---

## Questions

1. `model-api-keys-settings.tsx` — Was composing/extending the existing scope-parameterized `SecretsEditor` (or extracting a shared hook) considered before forking? Knowing the intent helps weigh the duplication trade-off.
2. `MODEL_API_KEYS[0]` (Anthropic) — On Modal the badge shows "Not set" despite the platform providing the key. Is provider-awareness in scope for this PR, or intentionally deferred?
3. Is there a reason this lives in the **Models** tab rather than alongside the generic **Secrets** tab, given both now edit the same store?

---

## Verdict

**Comment.** The implementation is correct, secure, and well-tested, and the feature genuinely fits the system's three-provider model. Not requesting changes on correctness grounds. Before merge, though, it's worth a conscious decision on the two design-fit points — the **logic duplication** and the **Modal `ANTHROPIC_API_KEY` "Not set"/collision** UX — since both touch how the wider system already handles secrets.

<sub>Methodology: the new component file isn't on the local checkout, so it was reviewed from the PR diff; every backend assumption was verified against the current `main` (2 commits behind `origin/main`, neither touching the secrets path).</sub>

PR #604, opened 2026-05-08
feat(terraform): support custom domains for Cloudflare workers

donnfelker:
## Summary
  - Adds two optional Terraform variables — `web_app_domain` and `control_plane_domain` — that override the default  `*.workers.dev` URLs in cross-service config (`NEXTAUTH_URL`, `NEXT_PUBLIC_WS_URL`, `CONTROL_PLANE_URL`, Modal allowed hosts) and outputs.
  - **Cloudflare-only feature.** `web_app_domain` takes effect only when `web_platform = "cloudflare"` — Vercel deployments use Vercel-managed domains and ignore the variable. `control_plane_domain` always applies since the control plane is on Cloudflare regardless of web platform.
  - Both default to empty, so existing deployments produce no diff. The variables describe the URL/identity layer only — binding the hostname to the worker (Cloudflare dashboard or `cloudflare_workers_custom_domain`) is out of scope and left to the operator.
  - Documents the full cutover in `docs/GETTING_STARTED.md` Step 9b (binding, GitHub App callback update, apply, smoke-test, troubleshooting table) plus a pointer in `docs/SETUP_GUIDE.md`. Both are flagged "Cloudflare only".
  - Adds `terraform test` coverage with five plan-mode runs (including `vercel_platform_ignores_web_app_domain` to lock the no-op behavior on the Vercel path) and wires `terraform test` into the existing `validate` CI job.

## Summary by CodeRabbit

* **New Features**
  * Added custom domain support for Cloudflare deployments (web app and control plane hostnames)

* **Tests**
  * Added Terraform plan-mode tests for custom-domain behavior
  * CI now runs Terraform tests during pre-plan validation and reports test results in the PR summary

* **Documentation**
  * Updated setup/getting-started docs with custom-domain setup, examples, callback updates, smoke-test steps, and troubleshooting guidance

[Review Change Stack](https://app.coderabbit.ai/change-stack/ColeMurray/background-agents/pull/604)

PR #747, opened 2026-06-15
feat(web): support custom domain

jehiah:
## Support a custom domain for the Cloudflare web Worker

Adds the ability to serve the web app on a custom domain (e.g. `app.example.com`)
instead of the default `*.workers.dev` URL when `web_platform = "cloudflare"`.

There was an existing `cloudflare_zone_id` terraform variable, but it was unused.

### Changes

- **`variables.tf`** — new optional `cloudflare_custom_domain` variable (the hostname to attach), alongside the existing `cloudflare_zone_id`. Defaults to `null`.
- **`web-cloudflare.tf`** — new `cloudflare_workers_custom_domain.web_app` resource that attaches the hostname to the web Worker. Cloudflare provisions the DNS record and edge certificate. Gated on `local.web_custom_domain_enabled` and depends on the Worker deploy.
- **`locals.tf`** — added `web_custom_domain_enabled` flag and a third case to `web_app_url`. When a custom domain is set, the app URL (used for `NEXTAUTH_URL` and cross-service config) becomes `https://<custom-domain>` instead of the `workers.dev` URL.
- **`terraform.tfvars.example`** — documented the new `cloudflare_custom_domain` input under the Cloudflare zone section.

### Behavior

The custom domain only activates when **all** of the following are true:

- `web_platform = "cloudflare"`
- `cloudflare_zone_id` is set
- `cloudflare_custom_domain` is non-empty

Otherwise the deployment falls back to the existing `workers.dev` URL — no change for existing deployments.

### Notes

- Requires a Cloudflare zone you control (set `cloudflare_zone_id`).
- The Cloudflare API token needs **Workers Routes: Edit** to manage the custom domain.
- `terraform validate` passes cleanly (dropped the deprecated `environment` attribute on the resource).

## Summary by CodeRabbit

* **New Features**
  * Added optional Cloudflare custom-domain support for production web workers, including automatic URL switching to the custom hostname when configured.
* **Safety Checks**
  * Added validation to prevent misconfigured custom-domain/zone settings; falls back to the default workers subdomain when required inputs are missing.
* **Documentation**
  * Expanded production Cloudflare setup guidance and examples for the custom-domain option.

### DIFF
--- a/terraform/environments/production/locals.tf
+++ b/terraform/environments/production/locals.tf
@@ -14,8 +14,13 @@ locals {
   control_plane_url  = "https://${local.control_plane_host}"
   ws_url             = "wss://${local.control_plane_host}"
 
+  # Whether a custom domain is configured for the Cloudflare web Worker
+  web_custom_domain_enabled = var.web_platform == "cloudflare" && var.cloudflare_custom_domain != null && var.cloudflare_custom_domain != "" && var.cloudflare_zone_id != null && var.cloudflare_zone_id != ""
+
   # Web app URL depends on deployment platform
-  web_app_url = var.web_platform == "cloudflare" ? (
+  web_app_url = local.web_custom_domain_enabled ? (
+    "https://${var.cloudflare_custom_domain}"
+    ) : var.web_platform == "cloudflare" ? (
     "https://open-inspect-web-${local.name_suffix}.${var.cloudflare_worker_subdomain}.workers.dev"
     ) : (
     "https://open-inspect-${local.name_suffix}.vercel.app"
--- a/terraform/environments/production/locals.tf
+++ b/terraform/environments/production/locals.tf
@@ -14,16 +14,27 @@ locals {
   control_plane_url  = "https://${local.control_plane_host}"
   ws_url             = "wss://${local.control_plane_host}"
 
+  # Must match the deployed Worker's `name` and the custom-domain `service` binding.
+  web_worker_name = "open-inspect-web-${local.name_suffix}"
+
   # Whether a custom domain is configured for the Cloudflare web Worker
-  web_custom_domain_enabled = var.web_platform == "cloudflare" && var.cloudflare_custom_domain != null && var.cloudflare_custom_domain != "" && var.cloudflare_zone_id != null && var.cloudflare_zone_id != ""
+  web_custom_domain_enabled = (
+    var.web_platform == "cloudflare" &&
+    trimspace(coalesce(var.cloudflare_custom_domain, "")) != "" &&
+    trimspace(coalesce(var.cloudflare_zone_id, "")) != ""
+  )
+
+  # Host the Cloudflare web Worker is served from: custom domain when configured,
+  # otherwise its default workers.dev hostname.
+  web_cloudflare_host = (local.web_custom_domain_enabled
+    ? var.cloudflare_custom_domain
+    : "${local.web_worker_name}.${var.cloudflare_worker_subdomain}.workers.dev"
+  )
 
   # Web app URL depends on deployment platform
-  web_app_url = local.web_custom_domain_enabled ? (
-    "https://${var.cloudflare_custom_domain}"
-    ) : var.web_platform == "cloudflare" ? (
-    "https://open-inspect-web-${local.name_suffix}.${var.cloudflare_worker_subdomain}.workers.dev"
-    ) : (
-    "https://open-inspect-${local.name_suffix}.vercel.app"
+  web_app_url = (var.web_platform == "cloudflare"
+    ? "https://${local.web_cloudflare_host}"
+    : "https://open-inspect-${local.name_suffix}.vercel.app"
   )
 
   # Worker script paths (deterministic output locations)
--- a/terraform/environments/production/terraform.tfvars.example
+++ b/terraform/environments/production/terraform.tfvars.example
@@ -27,7 +27,7 @@ cloudflare_api_token = ""
 cloudflare_account_id = ""
 
 # Cloudflare zone ID (optional, only needed for custom domains)
-# From domains overview list, select the "copy zone id"
+# In the Cloudflare dashboard, open the domain's Overview and use "Copy zone ID".
 # cloudflare_zone_id = ""
 
 # Custom domain (hostname) for the Cloudflare web Worker (optional).
--- a/terraform/environments/production/terraform.tfvars.example
+++ b/terraform/environments/production/terraform.tfvars.example
@@ -27,7 +27,14 @@ cloudflare_api_token = ""
 cloudflare_account_id = ""
 
 # Cloudflare zone ID (optional, only needed for custom domains)
+# From domains overview list, select the "copy zone id"
 # cloudflare_zone_id = ""
+
+# Custom domain (hostname) for the Cloudflare web Worker (optional).
+# Only used when web_platform = "cloudflare". Requires cloudflare_zone_id above
+# and a zone you control. Cloudflare provisions the DNS record + edge cert and
+# the web app URL becomes https://<this-hostname> instead of the workers.dev URL.
+# cloudflare_custom_domain = "" # e.g., "app.example.com"
 
 # Cloudflare Workers subdomain (found in Workers & Pages -> Overview, bottom-right of panel)
 # This is the subdomain used for workers.dev URLs (e.g., "myaccount" in "worker.myaccount.workers.dev")
--- a/terraform/environments/production/variables.tf
+++ b/terraform/environments/production/variables.tf
@@ -19,6 +19,12 @@ variable "cloudflare_zone_id" {
   default     = null
 }
 
+variable "cloudflare_custom_domain" {
+  description = "Custom domain (hostname) to attach to the Cloudflare web Worker (optional). Requires web_platform = 'cloudflare' and cloudflare_zone_id. e.g. 'app.example.com'"
+  type        = string
+  default     = null
+}
+
 variable "cloudflare_worker_subdomain" {
   description = "Cloudflare Workers account subdomain (e.g. 'myaccount' — .workers.dev is appended automatically)"
   type        = string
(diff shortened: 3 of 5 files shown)


jehiah:
@ColeMurray ready for review - I've been able to use this successfully.

jehiah:
I will note there is a similar PR in #604 but it doesn't include code to configure the domain in terraform with `cloudflare_workers_custom_domain`

terraform/environments/production/locals.tf
```diff
@@ -14,8 +14,13 @@ locals {
...
   ws_url             = "wss://${local.control_plane_host}"
 
+  # Whether a custom domain is configured for the Cloudflare web Worker
+  web_custom_domain_enabled = var.web_platform == "cloudflare" && var.cloudflare_custom_domain != null && var.cloudflare_custom_domain != "" && var.cloudflare_zone_id != null && var.cloudflare_zone_id != ""
+
   # Web app URL depends on deployment platform
-  web_app_url = var.web_platform == "cloudflare" ? (
+  web_app_url = local.web_custom_domain_enabled ? (
+    "https://${var.cloudflare_custom_domain}"
+    ) : var.web_platform == "cloudflare" ? (
```
ColeMurray: **[Automated Review]** This turns a previously-binary platform switch into a 3-way nested ternary that conflates two orthogonal questions — *which platform* (cloudflare vs vercel) and *which cloudflare host* (custom vs workers.dev). The `) : var.web_platform == "cloudflare" ? (` mid-expression nesting is hard to scan.

Since a custom domain is cloudflare-only, push that decision into its own local and keep `web_app_url` a binary switch (identical to its pre-PR shape):

```hcl
  # Host the Cloudflare web Worker is served from: custom domain when configured,
  # otherwise its default workers.dev hostname.
  web_cloudflare_host = (local.web_custom_domain_enabled
    ? var.cloudflare_custom_domain
    : "${local.web_worker_name}.${var.cloudflare_worker_subdomain}.workers.dev"
  )

  web_app_url = (var.web_platform == "cloudflare"
    ? "https://${local.web_cloudflare_host}"
    : "https://open-inspect-${local.name_suffix}.vercel.app"
  )
```

Behavior is identical across all three cases (`web_custom_domain_enabled` already implies cloudflare). The nesting disappears and each local answers exactly one question. (`web_worker_name` referenced here is the extraction suggested on `web-cloudflare.tf`.)

terraform/environments/production/locals.tf
```diff
@@ -14,8 +14,13 @@ locals {
   control_plane_url  = "https://${local.control_plane_host}"
   ws_url             = "wss://${local.control_plane_host}"
 
+  # Whether a custom domain is configured for the Cloudflare web Worker
+  web_custom_domain_enabled = var.web_platform == "cloudflare" && var.cloudflare_custom_domain != null && var.cloudflare_custom_domain != "" && var.cloudflare_zone_id != null && var.cloudflare_zone_id != ""
```
ColeMurray: **[Automated Review]** This is a long, unscannable line that spells out the `x != null && x != ""` pattern twice. The file already has a canonical "is this configured" idiom one block up (`google_enabled = trimspace(...) != ""`, line 10). Match it — using `coalesce` because these vars default to `null` (the google vars default to `""`). `trimspace` also defends against a whitespace-only value producing a broken `https:// ` URL:

```suggestion
  web_custom_domain_enabled = (
    var.web_platform == "cloudflare" &&
    trimspace(coalesce(var.cloudflare_custom_domain, "")) != "" &&
    trimspace(coalesce(var.cloudflare_zone_id, "")) != ""
  )
```

ColeMurray (commented):
**[Automated Review] — Thermo-nuclear code quality pass**

Functionally sound and correctly placed: `web_app_url` is the right single source of truth (it propagates to `NEXTAUTH_URL`, the slack/linear/control-plane `WEB_APP_URL` vars, `outputs.tf`, and the Vercel path), and the resource is schema-valid for the pinned provider — I verified `cloudflare_workers_custom_domain` against v5.19 (`hostname`+`service` required; `account_id`/`zone_id` optional; `environment` deprecated), so dropping `environment` is correct.

Inline findings, by priority:
1. **[locals.tf]** The `web_app_url` change grows a 3-way nested ternary that conflates platform-choice with host-choice — reframe via a `web_cloudflare_host` local so it stays binary. *(primary)*
2. **[web-cloudflare.tf]** `service` re-hardcodes the worker name a 4th time; extract `local.web_worker_name` to make the invariant explicit.
3. **[locals.tf]** The `web_custom_domain_enabled` flag is a long line duplicating a null-check pattern; match the file's existing `trimspace(...)` idiom.
4. **[variables.tf]** Optional: guard the silent fallback when `cloudflare_custom_domain` is set without `cloudflare_zone_id`, reusing the existing `checks.tf` pattern.

None are correctness blockers. I'd ask for 1 + 2 before merge; 3 and 4 are strongly recommended.

terraform/environments/production/web-cloudflare.tf
```diff
@@ -119,3 +119,16 @@ resource "null_resource" "web_app_cloudflare_deploy" {
...
+
+# Attach a custom domain to the web Worker (when configured).
+# Cloudflare provisions and manages the DNS record + edge cert for the hostname.
+resource "cloudflare_workers_custom_domain" "web_app" {
+  count = local.web_custom_domain_enabled ? 1 : 0
+
+  account_id = var.cloudflare_account_id
+  zone_id    = var.cloudflare_zone_id
+  hostname   = var.cloudflare_custom_domain
+  service    = "open-inspect-web-${local.name_suffix}"
```
ColeMurray: **[Automated Review]** `service` must equal the deployed Worker's `name`, but that worker name is now copy-pasted across 4 sites — `locals.tf:19`, `web-cloudflare.tf:50` (`WORKER_NAME`), `:67` (`name =`), and here. A typo in any one silently misroutes the domain to a non-existent service.

Extract the invariant into one local in `locals.tf`:

```hcl
  # Must match the deployed Worker's `name` and the custom-domain `service` binding.
  web_worker_name = "open-inspect-web-${local.name_suffix}"
```

then reuse it here (and replace the other three literals):

```suggestion
  service    = local.web_worker_name
```

terraform/environments/production/variables.tf
```diff
@@ -19,6 +19,12 @@ variable "cloudflare_zone_id" {
   default     = null
 }
 
+variable "cloudflare_custom_domain" {
```
ColeMurray, line 22: **[Automated Review]** Optional but recommended: if a user sets `cloudflare_custom_domain` but forgets `cloudflare_zone_id` (separate vars; zone_id is described as "optional"), the feature **silently does nothing** and serves the workers.dev URL. That's exactly the failure mode `checks.tf` already guards for Vercel (`check "vercel_url_matches"`, whose comment notes cross-service references "will silently break").

Reuse that idiom in `checks.tf`:

```hcl
check "cloudflare_custom_domain_config" {
  assert {
    condition = (
      var.web_platform != "cloudflare" ||
      trimspace(coalesce(var.cloudflare_custom_domain, "")) == "" ||
      trimspace(coalesce(var.cloudflare_zone_id, "")) != ""
    )
    error_message = "cloudflare_custom_domain is set but cloudflare_zone_id is empty — the custom domain is ignored and the app falls back to the workers.dev URL."
  }
}
```

terraform/environments/production/terraform.tfvars.example
```diff
@@ -27,8 +27,15 @@ cloudflare_api_token = ""
 cloudflare_account_id = ""
 
 # Cloudflare zone ID (optional, only needed for custom domains)
+# From domains overview list, select the "copy zone id"
```
ColeMurray: **[Automated Review]** Minor: this reads a bit garbled.

```suggestion
# In the Cloudflare dashboard, open the domain's Overview and use "Copy zone ID".
```

terraform/environments/production/locals.tf
```diff
@@ -9,8 +9,13 @@ locals {
...
   ws_url             = "wss://${local.control_plane_host}"
 
+  # Whether a custom domain is configured for the Cloudflare web Worker
+  web_custom_domain_enabled = var.web_platform == "cloudflare" && var.cloudflare_custom_domain != null && var.cloudflare_custom_domain != ""
+
   # Web app URL depends on deployment platform
-  web_app_url = var.web_platform == "cloudflare" ? (
+  web_app_url = local.web_custom_domain_enabled ? (
+    "https://${var.cloudflare_custom_domain}"
+    ) : var.web_platform == "cloudflare" ? (
```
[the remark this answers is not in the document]
jehiah: Already handled — `web_custom_domain_enabled` gates on `cloudflare_zone_id` (non-null and non-empty), so the resource won't be created with a null zone. (Also normalized to the `trimspace(coalesce(...))` idiom per the other thread.)

terraform/environments/production/locals.tf
jehiah: Done. Added a `web_cloudflare_host` local and collapsed `web_app_url` back to a binary platform switch — nesting gone, behavior identical.

terraform/environments/production/locals.tf
jehiah: Done — switched to the `trimspace(coalesce(...)) != ""` idiom matching `google_enabled`.

terraform/environments/production/web-cloudflare.tf
jehiah: Done. Extracted `local.web_worker_name` and replaced all four literals (`WORKER_NAME`, wrangler `name`, `service`, and the `web_app_url` host).

terraform/environments/production/variables.tf
jehiah: Done — added a `cloudflare_custom_domain_config` check block mirroring `vercel_url_matches`.

terraform/environments/production/terraform.tfvars.example
jehiah: Done — reworded.

vladgurovich:
This looks superuseful @ColeMurray -- could you review it?

ColeMurray (approved):

Issue #724, opened 2026-06-10
Improve Web App UI for get-task-status function call


ColeMurray:
When an agent makes a function call to view the task status of a child session, depending on the arguments, it can return the complete response.

It would be nice to have this better formatted and a more rich first-party presentation of this tool call within the UI

current

<img width="792" height="432" alt="Image" src="https://github.com/user-attachments/assets/208992e5-82f5-4dc9-8428-46edabdf275f" />

heat0206:
Hi! I'd like to work on this issue.

ColeMurray:
thanks @heat0206!

>I'm happy to extend this further with a richer custom UI component if maintainers would prefer a more structured presentation of task metadata and final responses.

this was something I had more in mind, basically some kind of rich formatted UI that would better show the output

ColeMurray:
@open-inspect[bot] the code currently doesn't look setup to render custom component UIs for tools. Review through the surrounding code and provide a high level overview of what would be required to do this?

heat0206:
Thanks, this makes sense.

For the renderer registry approach, would you prefer introducing the registry first and migrating the existing `slack-notify` special case to it, then adding `get-task-status` on top? Or would you prefer keeping the existing `slack-notify` implementation as-is and only using the registry for new tool renderers?

ColeMurray:
>introducing the registry first and migrating the existing slack-notify special case to it

this makes sense to me. let's introduce the registry with slack-notify as the first use-case, and then can create a follow-up change with the `get-task-status` specific one

heat0206:
Sounds good, that approach makes sense to me.

I'll take a look at introducing the renderer registry first and moving the existing slack-notify implementation over to it. Once that's in place, we can build the get-task-status renderer on top of the new structure.

PR #758, opened 2026-06-17
refactor(web): introduce tool renderer registry

heat0206:
## Summary

Introduce a simple tool renderer registry and migrate the existing `slack-notify` custom renderer to use it.

## Changes

* Added a `TOOL_RENDERERS` registry mapping tool names to custom renderer components.
* Replaced the hardcoded `slack-notify` conditional in `ToolCallItem` with a registry lookup.
* Preserved the existing generic rendering path for tools without a registered custom renderer.
* Typed the registry using `React.ComponentType<ToolCallItemProps>` to support future custom renderers.

## Motivation

This is a first step toward supporting additional custom tool renderers without adding more tool-specific conditionals inside `ToolCallItem`.

With this structure in place, future renderers (such as `get-task-status`) can be added by registering a component rather than modifying the core rendering logic.

## Summary by CodeRabbit

* **Refactor**
  * Updated tool rendering logic to streamline how different tool types are handled within the application.

### DIFF
--- a/packages/web/src/components/tool-call-item.tsx
+++ b/packages/web/src/components/tool-call-item.tsx
@@ -23,6 +23,11 @@ interface ToolCallItemProps {
   showTime?: boolean;
 }
 
+//Registry
+const TOOL_RENDERERS: Record<string, React.ComponentType<ToolCallItemProps>> = {
+  "slack-notify": SlackNotifyEvent,
+};
+
 function ToolIcon({ name }: { name: string | null }) {
   if (!name) return null;
 
@@ -51,14 +56,11 @@ function ToolIcon({ name }: { name: string | null }) {
 }
 
 export function ToolCallItem({ event, isExpanded, onToggle, showTime = true }: ToolCallItemProps) {
-  if (event.tool === "slack-notify") {
+  const Renderer = TOOL_RENDERERS[event.tool];
+
+  if (Renderer) {
     return (
-      <SlackNotifyEvent
-        event={event}
-        isExpanded={isExpanded}
-        onToggle={onToggle}
-        showTime={showTime}
-      />
+      <Renderer event={event} isExpanded={isExpanded} onToggle={onToggle} showTime={showTime} />
     );
   }
 


packages/web/src/components/tool-call-item.tsx
```diff
@@ -23,6 +23,11 @@ interface ToolCallItemProps {
   showTime?: boolean;
 }
 
+//Registry
+const TOOL_RENDERERS: Record<string, React.ComponentType<ToolCallItemProps>> = {
```
ColeMurray, line 27: **[Automated Review]** This `Record<string, React.ComponentType<ToolCallItemProps>>` typing is the crux of the scalability question. Because the value type is the _whole_ item component, every renderer has to re-implement the collapsible shell (the `py-0.5` wrapper, toggle button, chevron rotation, time span, and expanded container) — exactly what `SlackNotifyEvent` does today, duplicating ~25 lines that are character-identical to `ToolCallItem`'s. With N renderers you get N copies of that shell.

Consider narrowing the contract so a renderer contributes only what varies, and let `ToolCallItem` own the shell once:

```tsx
interface ToolRenderer {
  icon?: (e: ToolCallEvent) => ReactNode;     // defaults to formatToolCall
  summary?: (e: ToolCallEvent) => ReactNode;  // defaults to formatToolCall
  Body: (props: { event: ToolCallEvent }) => ReactNode;
}
```

This also mirrors the existing `eventRenderers` registry in `session-timeline.tsx` (render-fn + shared prop bag + `Partial<Record<...>>`), so the two dispatchers stay consistent. See the main review comment for why `apply_patch` specifically needs this.

Minor: `//Registry` -> `// Registry` (space after `//`), or drop the comment.

ColeMurray (commented):
**[Automated Review]**

Thanks @heat0206 — the direction is right and matches what was agreed in #724 (registry first, `slack-notify` as the first case), and the refactor is behavior-preserving. Before this becomes the foundation that `get-task-status` and other renderers build on, I want to flag one structural issue with the **contract** the registry is built on, since it determines whether the pattern actually scales. Line-specific notes are inline; the summary and the `apply_patch` analysis are here.

## The pattern already exists in this file — mirror it

`session-timeline.tsx` already has this exact registry, one level up, in a cleaner shape:

```tsx
// session-timeline.tsx — dispatch on event type, shared prop bag, single fallback
const eventRenderers: Partial<Record<SandboxEvent["type"], (props: EventRendererProps) => ReactNode>> = {
  user_message: UserMessageEvent,
  artifact:     ArtifactEvent,   // <- this is the screenshot renderer
  // ...
};
const render = eventRenderers[event.type];
if (!render) return null;
return render({ event, /* ... */ });
```

Two takeaways:

1. **Screenshots are already a registered renderer — on the _event_ axis, not the tool axis.** So the new `TOOL_RENDERERS` is a sibling registry one level down (dispatch on tool name _within_ a `tool_call`). Worth being explicit that these are two different axes so future "render types" land in the right one: rich/inline media -> an `artifact` event (existing path); custom framing of a tool's own result -> a tool renderer.
2. The new registry should match the established one's conventions (`Partial<Record<...>>`, render-fn + shared prop bag, single fallback) rather than introduce a parallel `Record<string, ComponentType<FullProps>>` shape.

## `apply_patch` is the real test — and it doesn't fit the current contract

The genuinely in-disguise _tool_ renderer today is `apply_patch`, special-cased inside the generic `ToolCallItem` body (`tool-call-item.tsx:66`, plus the `Patch:` block further down). Moving it into the registry is the litmus test for whether the abstraction earns its keep — `slack-notify` doesn't prove much, because it was already a standalone component.

The catch: **`apply_patch` does not fit the current `ComponentType<ToolCallItemProps>` contract without duplication.** The tell is what each renderer overrides:

| renderer | icon | collapsed summary | expanded body |
|---|---|---|---|
| `slack-notify` | custom | custom | custom |
| `apply_patch` | **generic** | **generic** | generic **+ one extra Patch section** |

`slack-notify` wants _total_ control, so "render the whole item" fits it by accident. `apply_patch` wants the generic collapsed line and most of the generic body, overriding only part of the body — and the full-component contract has no way to express partial override. Migrating it as-is forces you to copy the entire shell **and** re-implement the generic args/output body (and export the private `ToolIcon`), adding ~50 lines of duplication to delete ~4 lines of branching. Net negative — so in practice it wouldn't get migrated, and the registry would stay a one-entry `slack-notify` holder.

Under a slot contract it's a one-liner instead:

```tsx
interface ToolRenderer {
  icon?: (e: ToolCallEvent) => ReactNode;     // optional - defaults to formatToolCall
  summary?: (e: ToolCallEvent) => ReactNode;  // optional - defaults to formatToolCall
  Body: (props: { event: ToolCallEvent }) => ReactNode;
}

const TOOL_RENDERERS: Partial<Record<string, ToolRenderer>> = {
  "slack-notify": { icon: slackIcon, summary: slackSummary, Body: SlackNotifyBody },
  "apply_patch":  { Body: ApplyPatchBody },   // icon/summary fall back to generic
};
```

`ToolCallItem` owns the shell and renders the generic body as the default; `apply_patch` provides only its `Body` (reusing a shared `ToolDetails` with an extra `patch` prop), and the special-casing leaves `ToolCallItem` entirely.

## Recommendation

- **Preferred:** adopt the slot contract `{ icon?, summary?, Body }` + extract a shared `ToolCallShell` / `ToolDetails`, and migrate `apply_patch` in this PR as the proof. Adding the next tool is then genuinely one registry line.
- **Acceptable minimum:** keep `Record<string, ComponentType<...>>` but extract a shared `ToolCallShell` every renderer wraps, so `apply_patch` fits without shell duplication. (Still re-derives icon/summary per renderer; can't register a body-only tool.)
- **Not recommended:** ship the contract as-is — `apply_patch` can't move in cleanly, so the registry never proves itself and the `ToolCallItem` special-casing stays.

Happy to provide a concrete diff for the preferred version if that's useful.

packages/web/src/components/tool-call-item.tsx
```diff
@@ -51,14 +56,11 @@ function ToolIcon({ name }: { name: string | null }) {
 }
 
 export function ToolCallItem({ event, isExpanded, onToggle, showTime = true }: ToolCallItemProps) {
-  if (event.tool === "slack-notify") {
+  const Renderer = TOOL_RENDERERS[event.tool];
```
ColeMurray, line 59: **[Automated Review]** Two small things on the lookup key:

1. **Case-sensitivity divergence.** This uses the raw `event.tool`, but the canonical dispatch in `tool-formatters.ts` normalizes with `tool?.toLowerCase()`. So a renderer registered here only matches an exact-case tool name, while `formatToolCall` matches case-insensitively. Not a regression (the old `=== "slack-notify"` was also exact), but it bakes an inconsistency into the new seam — whoever adds `get-task-status` has to remember the registry is case-sensitive. Suggest keying on `event.tool?.toLowerCase()` to match the rest of the file.
2. This `const Renderer = registry[key]; if (Renderer) { ... }` shape is the same dispatch `EventItem` already does with `eventRenderers` in `session-timeline.tsx`. Aligning the two (same contract shape + fallback) keeps the rendering layer coherent.

heat0206:
Thanks for the detailed review. I spent some time digging into `ToolCallItem`, `SlackNotifyEvent`, and the current `apply_patch` handling and I think I understand the concern with the current `ComponentType<ToolCallItemProps>` contract.

`SlackNotifyEvent` currently owns the entire shell (toggle, timestamp, expanded container, etc.), while `apply_patch` only needs a much smaller customization around how its details are rendered. That makes `apply_patch` a better test of whether the abstraction scales.

I'm exploring a slot-based approach where `ToolCallItem` owns the shared shell and tool-specific renderers provide only the pieces that vary (body, and potentially icon/summary where needed). I'll continue prototyping that direction and see how it fits both the `slack-notify` and `apply_patch` cases.

PR #759, opened 2026-06-17
fix: never parse secret value-field pastes as KEY=VALUE

sudo-vaibhav:
## What was happening

Pasting a secret **value** that contains `=` into the value field silently corrupted it.

Concrete repro — an Azure storage connection string is a single valid value:

```
DefaultEndpointsProtocol=https;AccountName=acct;AccountKey=...==
```

Type the key name yourself, then paste that string into the value field. Instead of being stored as-is, it was reinterpreted as `KEY=VALUE`:

| | |
|---|---|
| stored key | `DEFAULTENDPOINTSPROTOCOL` (parsed from the value!) |
| stored value | `https;AccountName=acct;AccountKey=...==` (leading segment stripped) |
| your typed key | silently discarded |

So the saved connection string was broken and the secret landed under the wrong name — with no error.

## Why it happened — the root cause

The editor has a "smart paste": paste `.env` content and it auto-detects `KEY=VALUE` lines and imports them (`parseMaybeEnvContent`). That handler was wired to **both** the key input **and** the value input.

But auto-detect only has one coherent meaning: *"I'm starting to define secrets, here's structured content."* That intent exists at the **point of entry** — the key field, or a fresh row. It does **not** exist in the value field: by the time you're typing a value, you've already committed to "this row, this value." There is no input you could paste there that should be reparsed as `KEY=VALUE`.

Attaching the parser to the value field was a category error. Because secret values frequently contain `=` (connection strings, URLs with query params, base64, JWTs), the parser would grab the first `=`, treat everything before it as a key, and drop it.

## The fix

Fix the category error instead of guessing around it:

- **Value input: no paste handler at all.** A value is always literal. (This also covers multi-line values like PEM keys or JSON, which a line-count heuristic would still have mangled.)
- **Key input: unchanged**, handler renamed `handlePasteIntoKey` to make the asymmetry explicit and self-documenting. Pasting a single `KEY=VALUE` or a whole `.env` blob into the key field still auto-imports.

The asymmetry *is* the design: the key field is smart (entry point for structured paste), the value field is dumb (you already chose the field).

### Note on an earlier version of this PR

The first cut gated value-field paste on "single line → literal, multi-line → parse." That was a band-aid — it narrowed *when* the bug fired without removing the wrong coupling, and still hijacked multi-line values. Replaced with the root-cause fix above.

## Tests / checks

- Regression test: a full connection-string line parses as exactly one secret (value preserved, including internal `=` and trailing `==`).
- `env-paste` suite: pass · `typecheck` (web, after building `shared`): clean · `eslint` on touched files: clean
- No real secret value appears in any commit; the test uses a fabricated key.

## Release Notes

* **Bug Fixes**
  * Updated the secrets editor paste behavior to better distinguish single-line secret fragments from `.env`-style content.
  * Ensured the key field performs `.env`-style detection while the value field avoids single-line auto-parsing that could corrupt entries containing `=` and other special characters.

* **Tests**
  * Added coverage to confirm single-line assignments (including quoted values with semicolons and `=`) are parsed as complete secret values.


ColeMurray:
@open-inspect[bot] please provide a code review of this

ColeMurray:
thanks!

PR #762, opened 2026-06-18
feat: Github Setting for 'draft' mode

jehiah:
I want to add support for 'draft' mode PRs. This is important for me because it allows iterating on a PR before review is requested from required reviewers based on github `CODEOWNERS` files.

## Summary

Lets coding sessions open pull requests in **draft mode**, controlled two ways:

1. **A new optional `draft` parameter** on the agent's `create-pull-request` tool.
2. **A new "Source Code Mangement" setting — "Always use draft mode"** — with a global default and
   per-repo overrides. This is intentionally **separate from the existing "GitHub Bot" settings**
   (PR reviews / comment actions); it lives in its own integration card.

Draft creation already existed at the source-control provider layer (`github-provider.ts` honors
`config.draft`; the GitLab provider uses the `Draft:` title prefix). This change threads the flag
through the tool call and adds the setting that supplies the default.

**Resolution order:** an explicit setting always wins; when omitted, the
optional tool parameter is used.

## Screenshots
image

### New `draft` tool parameter

- `packages/sandbox-runtime/.../plugins/inspect-plugin.js` — `create-pull-request` accepts an
  optional `draft` boolean and forwards it.
- `packages/control-plane/src/routes/session-runtime-proxy.ts` — validates `draft` is a boolean and
  forwards it to the session Durable Object.
- `packages/control-plane/src/session/http/handlers/pull-request.handler.ts` — `CreatePrRequest`
  carries `draft` through to the service.
- `packages/control-plane/src/session/pull-request-service.ts` — computes the effective draft value
  (`input.draft ?? resolveAlwaysDraftDefault()`) and passes it to the provider.

### New "GitHub" integration setting (separate from "GitHub Bot")

- `packages/shared/src/types/integrations.ts` — new `IntegrationId` `"github-pr"`,
  `GitHubPrSettings { alwaysUseDraftMode? }`, settings-map entry, `GitHubPrGlobalConfig`, and a new
  `INTEGRATION_DEFINITIONS` card named **"GitHub"**. The existing "GitHub Bot" card is unchanged.
- `packages/control-plane/src/db/integration-settings.ts` — validates `alwaysUseDraftMode` is a
  boolean. Global-default + per-repo-override merge is inherited from the existing framework.
- `packages/control-plane/src/routes/integration-settings.ts` — resolved-config branch for
  `github-pr`.
- `packages/control-plane/src/session/durable-object.ts` — `resolveAlwaysDraftDefault()` reads the
  resolved `github-pr` config for the session's repo; returns `false` when D1 is unavailable.
- `packages/web/src/components/settings/integrations/github-pr-integration-settings.tsx` (new) +
  `packages/web/src/app/(app)/settings/integrations/[id]/page.tsx` — settings card with a global
  "Always use draft mode" toggle and per-repo overrides, mirroring the code-server pattern.

## Scope notes

- The setting applies to **all** session-created PRs, including sessions triggered by the GitHub
  bot's @mentions — this is the intended scope, and the setting remains fully decoupled from the
  GitHub Bot configuration.
- **No schema migration required** — the generic `integration_settings` /
  `integration_repo_settings` tables already store arbitrary JSON keyed by integration id.

## Testing

- **PR service** (`pull-request-service.test.ts`): default `draft=false`, explicit flag forwarded,
  setting fallback when unspecified, and explicit `draft=false` overriding an enabled setting.
- **Settings store** (`integration-settings.test.ts`): round-trip global/repo, non-boolean
  rejection, and repo override flipping the global default.
- **Proxy** (`session-runtime-proxy.test.ts`): `draft` forwarding and non-boolean rejection.
- Full control-plane unit suite (1353) and web suite (296) pass; `typecheck` and `lint` are clean on
  all changed files.

## Summary by CodeRabbit

* **New Features**
  * Added draft pull request support across the create-PR flow (including optional draft input from the app/tooling)
  * Introduced an SCM Settings page for managing source-control defaults
  * Added an “Always use draft mode” toggle to default PRs to draft
  * Added per-repository SCM overrides to customize draft defaults by repository
* **Bug Fixes**
  * Added validation to reject non-boolean draft values with a clear 400 error response

### DIFF
--- a/packages/control-plane/src/routes/scm-settings.ts
+++ b/packages/control-plane/src/routes/scm-settings.ts
@@ -1,0 +1,198 @@
+/**
+ * SCM (source-control) settings routes.
+ *
+ * SCM settings are a top-level setting, separate from the integration-settings
+ * framework. They control how sessions open pull/merge requests (e.g. always as
+ * drafts) for both GitHub and GitLab.
+ */
+
+import type { ScmGlobalConfig, ScmSettings } from "@open-inspect/shared";
+import { ScmSettingsStore, ScmSettingsValidationError } from "../db/scm-settings";
+import type { Env } from "../types";
+import { createLogger } from "../logger";
+import {
+  type Route,
+  type RequestContext,
+  parsePattern,
+  json,
+  error,
+  parseJsonBody,
+  extractRepoParams,
+} from "./shared";
+
+const logger = createLogger("router:scm-settings");
+
+async function handleGetGlobal(
+  _request: Request,
+  _env: Env,
+  _match: RegExpMatchArray,
+  ctx: RequestContext
+): Promise<Response> {
+  const store = new ScmSettingsStore(ctx.db);
+  const settings = await store.getGlobal();
+  return json({ settings });
+}
+
+async function handleSetGlobal(
+  request: Request,
+  _env: Env,
+  _match: RegExpMatchArray,
+  ctx: RequestContext
+): Promise<Response> {
+  const body = await parseJsonBody<{ settings?: ScmGlobalConfig }>(request);
+  if (body instanceof Response) return body;
+
+  if (!body?.settings || typeof body.settings !== "object" || Array.isArray(body.settings)) {
+    return error("Request body must include settings object", 400);
+  }
+
+  const store = new ScmSettingsStore(ctx.db);
+
+  try {
+    await store.setGlobal(body.settings);
+    logger.info("scm_settings.updated", {
+      event: "scm_settings.updated",
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return json({ status: "updated" });
+  } catch (e) {
+    if (e instanceof ScmSettingsValidationError) {
+      return error(e.message, 400);
+    }
+    logger.error("Failed to update SCM settings", {
+      error: e instanceof Error ? e.message : String(e),
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return error("SCM settings storage unavailable", 503);
+  }
+}
+
+async function handleDeleteGlobal(
+  _request: Request,
+  _env: Env,
+  _match: RegExpMatchArray,
+  ctx: RequestContext
+): Promise<Response> {
+  const store = new ScmSettingsStore(ctx.db);
+
+  try {
+    await store.deleteGlobal();
+    logger.info("scm_settings.deleted", {
+      event: "scm_settings.deleted",
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return json({ status: "deleted" });
+  } catch (e) {
+    logger.error("Failed to delete SCM settings", {
+      error: e instanceof Error ? e.message : String(e),
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return error("SCM settings storage unavailable", 503);
+  }
+}
+
+async function handleListRepoSettings(
+  _request: Request,
+  _env: Env,
+  _match: RegExpMatchArray,
+  ctx: RequestContext
+): Promise<Response> {
+  const store = new ScmSettingsStore(ctx.db);
+  const repos = await store.listRepoSettings();
+  return json({ repos });
+}
+
+async function handleSetRepoSettings(
+  request: Request,
+  _env: Env,
+  match: RegExpMatchArray,
+  ctx: RequestContext
+): Promise<Response> {
+  const params = extractRepoParams(match);
+  if (params instanceof Response) return params;
+  const { owner, name } = params;
+  const repo = `${owner}/${name}`;
+
+  const body = await parseJsonBody<{ settings?: ScmSettings }>(request);
+  if (body instanceof Response) return body;
+
+  if (!body?.settings || typeof body.settings !== "object" || Array.isArray(body.settings)) {
+    return error("Request body must include settings object", 400);
+  }
+
+  const store = new ScmSettingsStore(ctx.db);
+
+  try {
+    await store.setRepoSettings(repo, body.settings);
+    logger.info("scm_repo_settings.updated", {
+      event: "scm_repo_settings.updated",
+      repo,
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return json({ status: "updated", repo });
+  } catch (e) {
+    if (e instanceof ScmSettingsValidationError) {
+      return error(e.message, 400);
+    }
+    logger.error("Failed to update SCM repo settings", {
+      error: e instanceof Error ? e.message : String(e),
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return error("SCM settings storage unavailable", 503);
+  }
+}
+
+async function handleDeleteRepoSettings(
+  _request: Request,
+  _env: Env,
+  match: RegExpMatchArray,
+  ctx: RequestContext
+): Promise<Response> {
+  const params = extractRepoParams(match);
+  if (params instanceof Response) return params;
+  const { owner, name } = params;
+  const repo = `${owner}/${name}`;
+
+  const store = new ScmSettingsStore(ctx.db);
+
+  try {
+    await store.deleteRepoSettings(repo);
+    logger.info("scm_repo_settings.deleted", {
+      event: "scm_repo_settings.deleted",
+      repo,
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return json({ status: "deleted", repo });
+  } catch (e) {
+    logger.error("Failed to delete SCM repo settings", {
+      error: e instanceof Error ? e.message : String(e),
+      request_id: ctx.request_id,
+      trace_id: ctx.trace_id,
+    });
+    return error("SCM settings storage unavailable", 503);
+  }
+}
+
+export const scmSettingsRoutes: Route[] = [
+  { method: "GET", pattern: parsePattern("/scm-settings"), handler: handleGetGlobal },
+  { method: "PUT", pattern: parsePattern("/scm-settings"), handler: handleSetGlobal },
+  { method: "DELETE", pattern: parsePattern("/scm-settings"), handler: handleDeleteGlobal },
+  { method: "GET", pattern: parsePattern("/scm-settings/repos"), handler: handleListRepoSettings },
+  {
+    method: "PUT",
+    pattern: parsePattern("/scm-settings/repos/:owner/:name"),
+    handler: handleSetRepoSettings,
+  },
+  {
+    method: "DELETE",
+    pattern: parsePattern("/scm-settings/repos/:owner/:name"),
+    handler: handleDeleteRepoSettings,
+  },
+];
--- a/packages/control-plane/src/routes/session-runtime-proxy.ts
+++ b/packages/control-plane/src/routes/session-runtime-proxy.ts
@@ -148,6 +148,10 @@ async function handleCreatePR(
     return error("repoName must be a string");
   }
 
+  if (body.draft !== undefined && typeof body.draft !== "boolean") {
+    return error("draft must be a boolean");
+  }
+
   return ctx.sessionRuntime.fetch(sessionId, SessionInternalPaths.createPr, {
     method: "POST",
     headers: { "Content-Type": "application/json" },
@@ -158,6 +162,7 @@ async function handleCreatePR(
       headBranch: body.headBranch,
       repoOwner: body.repoOwner,
       repoName: body.repoName,
+      draft: body.draft,
     }),
   });
 }
--- a/packages/control-plane/src/session/pull-request-service.test.ts
+++ b/packages/control-plane/src/session/pull-request-service.test.ts
@@ -124,7 +124,7 @@ function createRepositoryRow(overrides: Partial<SessionRepositoryRow> = {}): Ses
   };
 }
 
-function createTestHarness() {
+function createTestHarness(options: { alwaysDraftDefault?: boolean } = {}) {
   const log = createMockLogger();
   const provider = createMockProvider();
   const artifacts: ArtifactRow[] = [];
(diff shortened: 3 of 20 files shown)


packages/shared/src/types/integrations.ts
```diff
@@ -144,6 +156,11 @@ export const INTEGRATION_DEFINITIONS: {
     name: "GitHub Bot",
     description: "Automated PR reviews and comment-triggered actions",
   },
+  {
+    id: "github-pr",
+    name: "GitHub",
```
ColeMurray: [Automated Review] **[Design/UX]** Naming collision: this registers a card named **"GitHub"** (`github-pr`) that renders right next to the existing **"GitHub Bot"** (`github`) card, with no icons to tell them apart. A bare "GitHub" card holding a single draft toggle reads as the primary/overarching GitHub integration, which it isn't. Suggest renaming to **"GitHub Pull Requests"** (keep the `github-pr` id) so the label matches the scope and leaves room for future PR defaults (labels, reviewers, branch naming). Alternatively, if this will only ever hold one toggle, consider folding `alwaysUseDraftMode` into the existing `github` settings rather than a new integration + 391-line component.

packages/control-plane/src/session/pull-request-service.ts
```diff
@@ -188,12 +201,17 @@ export class SessionPullRequestService {
       const fullBody =
         input.body + `\n\n---\n*Created with [${this.deps.appName}](${input.sessionUrl})*`;
 
+      // An explicit `draft` on the request wins; otherwise fall back to the
+      // configured "always use draft mode" GitHub setting for this repo.
+      const draft = input.draft ?? (await this.deps.resolveAlwaysDraftDefault());
```
ColeMurray: [Automated Review] **[Design]** The setting is surfaced as **"Always use draft mode"**, but because an explicit tool `draft` wins here (`input.draft ?? ...`), it's really a *default*, not *always*. The tool description even tells the model it can pass `draft: false` to force a ready-for-review PR — so a user who enables this as a `CODEOWNERS` guardrail can still get a non-draft PR if the model decides the work is "ready." Consider either renaming to **"Use draft mode by default"** to match the soft-default behavior, or making it a hard policy (force draft when enabled, regardless of the tool arg). The resolution code itself is correct.

packages/control-plane/src/routes/integration-settings.ts
```diff
@@ -318,6 +319,18 @@ async function handleGetResolvedConfig(
     });
   }
 
+  if (id === "github-pr") {
```
ColeMurray: [Automated Review] **[Suggestion]** This `github-pr` branch of `handleGetResolvedConfig` (the `/resolved/:owner/:name` endpoint) currently has no caller: the DO reads `getResolvedConfig` in-process and the web UI uses the global + `/repos` endpoints. It also returns `enabledRepos`, which `github-pr` has no concept of (always `null`). Harmless and consistent with siblings, but it's effectively dead/noise today — consider dropping `enabledRepos` here, or adding a comment that the branch exists for parity / future runtime callers.

packages/web/src/components/settings/scm-settings.tsx
```diff
@@ -0,0 +1,391 @@
...
+        </Button>
+        <Button variant="destructive" size="sm" onClick={handleDelete}>
+          Remove
+        </Button>
+      </div>
+    </div>
+  );
+}
+
+function Section({
```
ColeMurray, line 380: [Automated Review] **[Maintainability]** This component is a near-verbatim copy of `code-server-integration-settings.tsx` — this `Section` helper is now duplicated across three integration components, along with the whole save/reset/dirty/toast loop. Not introduced solely by this PR, but each new integration deepens the duplication. A shared `useIntegrationSettings` hook + shared `Section`/`RepoOverrideRow` would shrink this file substantially and keep the integrations consistent.

packages/control-plane/src/routes/session-runtime-proxy.ts
```diff
@@ -94,6 +94,10 @@ async function handleCreatePR(
     return error("headBranch must be a string");
   }
 
+  if (body.draft != null && typeof body.draft !== "boolean") {
```
ColeMurray: [Automated Review] **[Nit]** `body.draft != null` lets an explicit `draft: null` pass validation (also flagged by CodeRabbit). Functionally harmless — downstream `input.draft ?? resolveAlwaysDraftDefault()` treats `null` like "omitted," arguably the more lenient-correct behavior. But the contract is `draft?: boolean`, so for strictness use `body.draft !== undefined && typeof body.draft !== "boolean"`. Low priority.

packages/web/src/components/settings/integrations/github-pr-integration-settings.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+  const handleAdd = async () => {
+    if (!addingRepo) return;
+    const [owner, name] = addingRepo.split("/");
+
+    try {
+      const res = await fetch(`/api/integration-settings/github-pr/repos/${owner}/${name}`, {
+        method: "PUT",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ settings: { alwaysUseDraftMode: true } }),
```
ColeMurray: [Automated Review] **[Nit]** Adding a repo override always seeds `{ alwaysUseDraftMode: true }`. A user who adds an override specifically to *disable* draft for a repo has to add it (gets `true`) → untoggle → save. Consider seeding from the current global default so "Add Override" doesn't silently flip the repo to draft.

packages/web/src/components/settings/scm-settings.tsx
```diff
@@ -0,0 +1,391 @@
...
+  const [alwaysUseDraftMode, setAlwaysUseDraftMode] = useState(
+    settings?.defaults?.alwaysUseDraftMode ?? false
+  );
+  const [saving, setSaving] = useState(false);
+  const [dirty, setDirty] = useState(false);
+  const [initialized, setInitialized] = useState(false);
+  const [showResetDialog, setShowResetDialog] = useState(false);
+
+  useEffect(() => {
+    if (settings !== undefined && !initialized) {
```
ColeMurray, line 102: [Automated Review] **[Nit]** The parent gates rendering on `globalLoading`/`repoSettingsLoading` and only mounts this section once data is resolved, so the `useState` initializer already captures the right value — this `initialized`/`useEffect` re-sync is redundant (and won't reflect later SWR revalidations). Harmless, and it mirrors the existing pattern, so fine to leave.

ColeMurray (commented):
## [Automated Review] Summary

Draft-mode threading is **correct end-to-end** (sandbox tool → proxy → handler → service → provider) and well covered by tests at the store/proxy/service layers. The provider boundary already honored `draft` (GitHub `requestBody.draft`, GitLab `Draft:` title prefix), so this is low-risk plumbing plus a settings surface. **No blocking bugs.** Verdict: **Comment** — the main thing worth resolving before merge is the "GitHub" / "GitHub Bot" naming collision (see inline).

Two findings touch **unchanged files**, so they can't be left inline:

- **[Reliability]** `source-control/providers/github-provider.ts` (`if (config.draft) requestBody.draft = true`): GitHub rejects draft PRs on **private repos under free plans** (HTTP 422), with no fallback here. Enabling the global default could break PR creation for a subset of repos. Worth a note in the setting's help text and/or a graceful retry-without-draft on 422.
- **[Testing]** `github-provider.test.ts` has **zero tests for `createPullRequest`**, so the GitHub draft path this feature now depends on is unverified at the provider layer (service tests mock the provider). GitLab has three draft tests; GitHub should have at least one.

Inline comments below are prefixed `[Automated Review]`.

ColeMurray:
Some use-case questions here:

In the original implementation, this still allows for the agent to override the draft PR setting via an input (this may be a bug). Is the intent of this change to have this ALWAYS be draft mode when selected, or if not explicitly provided.

Related, I'm onboard with adding the optional parameter to the tool as this make sense. Would the intent of this functionality be better served by custom instructions to the agent, as it seems like it may create confusion to the agent, where the setting is hard-coded upstream to "always draft" while the agent itself  called it with `draft:false` and may lead to it using GH cli to then alter it out of draft state.

Related, I don't think putting this under Github PR makes sense, as the draft PR is an source control provider level concept, rather than specific to github. This is evident by the generic SCM provider types being updated.
It may make more sense here to relax the naming around this to just "Pull request" or "SCM settings" / similar.

ColeMurray (commented):
-

ColeMurray (changes requested):

jehiah:
> Some use-case questions here:
> 
> In the original implementation, this still allows for the agent to override the draft PR setting via an input (this may be a bug). Is the intent of this change to have this ALWAYS be draft mode when selected, or if not explicitly provided.
> 

Yes - this is a bug. My intention was a "force" option (but if you have a strong preference for a "default to draft mode" I could go that direction).

> Related, I'm onboard with adding the optional parameter to the tool as this make sense. Would the intent of this functionality be better served by custom instructions to the agent, as it seems like it may create confusion to the agent, where the setting is hard-coded upstream to "always draft" while the agent itself called it with `draft:false` and may lead to it using GH cli to then alter it out of draft state.

Yes my desire is to make the draft functionality generally exposed so tooling is flexible enough to enable guide particular workflows one direction or the other they could (i.e. open draft mode if tests have not been run, etc) - assuming not everyone will want to force it like I do. I think that has value but I've found limited success and at Bitly we prefer to always start as a draft. You can always flip to non-draft but it's not reversible - you can't undo automated review requests that are sent when you don't start in draft mode.

There is a little bit of conflict in exposing a draft option in the tool if it's forced for a particular repo/setup - I think that's ok and won't cause problems - but happy to iterate if it does. I'll update the tool settings to imply that the draft option may be overridden by a remote policy.

> 
> Related, I don't think putting this under Github PR makes sense, as the draft PR is an source control provider level concept, rather than specific to github. This is evident by the generic SCM provider types being updated. It may make more sense here to relax the naming around this to just "Pull request" or "SCM settings" / similar.

I'm totally open on where to set this. I was thinking before i started it would be Github specific - but GitLab supports draft mode too so as you say a top level "SCM settings" or "[Github / GitLab] SCM" might be appropriate.  I've been noodling on how to support repositories that use a fork workflow and that setting structure would be a good home for that too.

packages/control-plane/src/session/pull-request-service.test.ts
```diff
@@ -258,6 +259,46 @@ describe("SessionPullRequestService", () => {
...
+  });
+
+  it("forces draft when always-draft is enabled, even if the request sets draft=false", async () => {
+    harness = createTestHarness({ alwaysDraftDefault: true });
+
+    await harness.service.createPullRequest(createInput({ draft: false }));
+
+    expect(harness.provider.createPullRequest).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({ draft: true })
```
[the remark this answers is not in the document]
jehiah: The test is correct; alwaysDraftDefault overrides the createPullRequest input.

packages/web/src/components/settings/scm-settings.tsx
```diff
@@ -0,0 +1,397 @@
...
+  const availableForOverride = availableRepos.filter(
+    (r) => !overriddenRepos.has(r.fullName.toLowerCase())
+  );
+
+  const handleAdd = async () => {
+    if (!addingRepo) return;
+    const [owner, name] = addingRepo.split("/");
+
+    try {
+      const res = await fetch(`/api/integration-settings/scm/repos/${owner}/${name}`, {
```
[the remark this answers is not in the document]
jehiah: This is a **pre-existing, cross-cutting limitation** — every integration repo-override UI uses the same [owner]/[name] two-segment split and route shape. 

A proper fix means reworking the repo-override route to a catch-all across the whole settings system, which is out of scope for this PR. Worth a follow-up issue to expand GitLab support or reject the subgroup pattern explicitly.

packages/web/src/app/api/scm-settings/repos/[owner]/[name]/route.ts
```diff
@@ -0,0 +1,59 @@
...
+        method: "PUT",
+        body: JSON.stringify(body),
+      }
+    );
+    const data = await response.json();
+    return NextResponse.json(data, { status: response.status });
+  } catch (error) {
+    console.error("Failed to update SCM repo settings:", error);
+    return NextResponse.json({ error: "Failed to update SCM repo settings" }, { status: 500 });
+  }
```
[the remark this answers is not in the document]
Comment (jehiah) [reply packages/web/src/app/api/scm-settings/repos/[owner]/[name]/route.ts:31]: every web API proxy route in the repo uses the same single-try 500 pattern; changing only the scm routes would be inconsistent.

packages/control-plane/src/routes/scm-settings.ts
```diff
@@ -0,0 +1,220 @@
...
+  _match: RegExpMatchArray,
+  _ctx: RequestContext
+): Promise<Response> {
+  if (!env.DB) {
+    return json({ settings: null });
+  }
+  const store = new ScmSettingsStore(env.DB);
+  const settings = await store.getGlobal();
+  return json({ settings });
+}
```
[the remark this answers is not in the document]
jehiah: the sibling `integration-settings.ts` read handlers don't wrap reads either; they fall through to the router's generic handler. Skipped for consistency with the established pattern.

packages/shared/src/types/integrations.ts
jehiah: renamed to "SCM settings" with "scm" key.

packages/control-plane/src/session/pull-request-service.ts
jehiah: clarified intent is for "always use draft mode"

packages/web/src/app/api/scm-settings/route.ts
```diff
@@ -0,0 +1,57 @@
...
+    const response = await controlPlaneFetch("/scm-settings", {
+      method: "PUT",
+      body: JSON.stringify(body),
+    });
+    const data = await response.json();
+    return NextResponse.json(data, { status: response.status });
+  } catch (error) {
+    console.error("Failed to update SCM settings:", error);
+    return NextResponse.json({ error: "Failed to update SCM settings" }, { status: 500 });
+  }
```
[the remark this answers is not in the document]
jehiah: valid - but matches existing code pattern for a single try/catch on routes.

jehiah:
RFRR @ColeMurray 

image

jehiah:
@ColeMurray please take another look - I've rebased and squashed this PR so that it apples cleanly on `main` again. Previous review comments are addressed.

jehiah:
@ColeMurray bumping this feature for draft mode support - I've rebased to resolve merge conflicts so this again will apply cleanly.

ColeMurray:
merged with some revisions in https://github.com/ColeMurray/background-agents/pull/1285

thanks!

Issue #833, opened 2026-06-26
Support no-repo and workspace automation targets


rubenlangeweg:
### Why

Automations should not have to know the repository before they start.

Some automations are simple: fixed repo, fixed branch, run on a schedule. Others start from operational context. The incident.io flow is the clearest example. An alert/incident notification comes in, but the system does not know the repo yet. The first agent should inspect the incident, use incident.io MCP and catalog context to understand the affected service, then decide whether code work is needed.

If code work is needed, that should happen in a separate repo-bound session with the normal access checks and scoped credentials. The no-repository run is the coordinator. It should not be a session with broad repo access.

The same shape applies to team and platform work. A weekly automation to update `AGENTS.md` across selected repos should resolve a repo set, then create separate repo-bound runs per repo. It should not become one session with access to everything.

### Product context

This matches the direction Cursor has already taken in the Agent window. Agents can start from `Home`, with no repository selected, or from a workspace that groups several repositories. Automations could follow the same mental model.

The useful product shape is:

- `Home`: run an automation without a repository. Good for incident triage, Slack or Linear routing, operational checks, reporting, and MCP-driven investigation.
- `Single repository`: current behavior. Good for a known codebase and branch.
- `Workspace` or repo set: run against multiple repositories that belong to the same product, team, or service boundary.

### Target model

Automation creation stores the intended target mode. Each run resolves that mode into an execution target before execution starts.

Current target modes:

- `fixed_single_repo`: existing behavior. Resolve one repo and branch before creating the session.
- `no_repository`: coordinator mode. No clone, no SCM credentials, no PR tool. It can investigate external context and request repo-bound work later.

Future target modes should fit the same model:

- `webhook_payload_single_repo`: resolve the repo from the trigger payload.
- `repo_set`: expand a selected repo set or workspace into separate repo-bound runs.
- parent or child automation runs: coordinate multi-step work without mixing credential scopes.

The unit of product behavior should be the automation run. A sandbox is one execution backend, not the definition of an automation.

### Architecture note

The current PR slice can reasonably reuse the existing session and sandbox pipeline. That gets the feature in users' hands with the existing LLM runtime, MCP injection, logs, cancellation, callbacks, and UI.

However, a sandbox-backed no-repository session could be a good MVP, but probably not the final infrastructure shape. For many no-repo jobs, the work is mostly API and MCP calls. Spinning up Modal and OpenCode for every run may be heavier than needed once volume grows.

The long-term shape should be:

[...]

PR #836, opened 2026-06-26
feat: support automations without repositories

rubenlangeweg:
## Summary

Automations can now run without a repository target, so operational workflows can start agent sessions without cloning a code workspace or inventing a placeholder repo. Repository-backed automations still resolve through SCM access before session creation, while no-repository sessions disable repo-only behavior like code-server, child sessions, and PR creation.

Current boundary: `no_repository` automations are coordinator sessions only; code investigation or PR work should start a separate repo-bound session after a repository is discovered.

## Why

This is the first target-resolution slice for #833. It enables operational automations that do not know a repository at trigger time, such as an incident/catalog workflow that first investigates external context, then starts a separate repo-bound session only after the relevant repository is discovered.

## What Changed

- Adds `fixed_single_repo` / `no_repository` target handling at automation creation and scheduling time.
- Allows automation and session rows to store null repository fields through D1 migration `0029_allow_no_repository_targets.sql`.
- Teaches sandbox startup to skip clone/setup hooks when no repository is present while preserving the existing repository path for normal sessions.
- Updates web session and automation UI labels so no-repository sessions render as a first-class state.

Below is a recording of the working implementation deployed on CF. 

https://github.com/user-attachments/assets/e2d51f44-77dc-4a3e-9279-02105797132c

## Validation

- `npm run typecheck`
- `npm run lint`
- `npm test`
- `packages/control-plane`: analytics integration test via `vitest.integration.config.ts`
- `packages/modal-infra`: focused pytest and ruff on touched files
- `packages/sandbox-runtime`: focused pytest on touched runtime tests
- D1 migration validated against a seeded pre-0029 SQLite database with `PRAGMA foreign_key_check = 0`
- `git diff --check`

Refs #833

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT-5

Made with [Orca](https://github.com/stablyai/orca) 🐋

## Summary by CodeRabbit

* **New Features**
  * Added automation/session “no repository” support with a target mode selector (`fixed_single_repo` vs `no_repository`), including repository configuration UI and consistent “No repository” labeling.
* **Bug Fixes**
  * Improved handling of missing/nullable repo context across automation resolution, session creation, PRs, SCM credentials, sandbox feature gating/tool installs, analytics repo breakdowns, and scheduler/Slack audit + callback payloads.
  * Added clearer validation and compatibility errors for unsupported or incompatible target modes.
* **Documentation**
  * Updated automation setup documentation with revised repository configuration flow, required fields, instruction limits, and supported automation target modes.


rubenlangeweg:
Skipping the CodeRabbit nitpicks intentionally.

- Extra assertion-only tests in automations, Slack notify, sandbox lifecycle, and scheduler specs: not worth the added test churn for this fix.
- Shared Automation/CreateAutomationRequest discriminated unions: useful cleanup, but it spreads type churn across packages and is not required for the no-repository behavior fix.

The actual behavior fixes are committed in d76702a and covered by the final CONTRIBUTING.md checks.

ColeMurray:
@rubenlangeweg can you include a few screenshots of the UI for this

rubenlangeweg:
@ColeMurray attached some screenshots and a recording. 

CleanShot 2026-06-26 at 23 11 01@2x
CleanShot 2026-06-26 at 23 11 08@2x
CleanShot 2026-06-26 at 23 11 27

packages/control-plane/src/session/schema.ts
```diff
@@ -383,6 +383,75 @@ export const MIGRATIONS: readonly SchemaMigration[] = [
...
+      const columnMap = new Map(columns.map((column) => [column.name, column]));
+      if (
+        columnMap.get("repo_owner")?.notnull === 0 &&
+        columnMap.get("repo_name")?.notnull === 0 &&
+        columnMap.get("base_branch")?.notnull === 0
+      ) {
+        return;
+      }
+
+      sql.exec("BEGIN");
```
ColeMurray: [Automated review]
**🔴 Critical (blocking): `sql.exec("BEGIN")` throws in the Durable Object runtime and bricks every pre-existing session on deploy.**

Cloudflare's DO SQLite (`ctx.storage.sql.exec()`) rejects transaction-control statements — the `BEGIN` here and the `COMMIT`/`ROLLBACK` at lines 448/450. Per the [SQLite Storage API docs](https://developers.cloudflare.com/durable-objects/api/sqlite-storage-api/): *"note that `sql.exec()` cannot execute transaction-related statements like `BEGIN TRANSACTION` or `SAVEPOINT`."* The runtime throws:

> To execute a transaction, please use the `state.storage.transaction()` or `state.storage.transactionSync()` APIs instead of the SQL `BEGIN TRANSACTION` or `SAVEPOINT` statements.

(`BEGIN` ≡ `BEGIN TRANSACTION` in SQLite, so the bare form is caught by the same authorizer.)

**Trigger:** the guard at lines 395–401 only returns early when `repo_owner`/`repo_name`/`base_branch` are *already* nullable. Every session DO created before this PR has `repo_owner TEXT NOT NULL` / `base_branch TEXT NOT NULL DEFAULT 'main'` (the pre-PR `SCHEMA_SQL`), so `notnull === 1`, the guard does **not** return, and `BEGIN` executes.

**Impact:** `BEGIN` is *outside* the `try` (which opens at line 404), so the throw propagates out of `initSchema` → `ensureInitialized()` never sets `initialized = true` → the migration re-runs and re-throws on **every** subsequent `fetch()`. All historical sessions return 500 permanently after deploy. No data is lost (it throws before `DROP TABLE session`), but it's a full availability brick of existing sessions. Moving `BEGIN` inside the `try` doesn't help — the `catch` then runs `ROLLBACK`, which is rejected too.

**Fix:** drop the `BEGIN`/`COMMIT`/`ROLLBACK` lines entirely. `applyMigrations()` is fully synchronous (no intervening `await`), so the `DROP TABLE` + `RENAME` sequence already runs in a single DO write turn and is atomic under the runtime's automatic write-coalescing — *"Any series of write operations with no intervening `await` will automatically be submitted atomically."* The explicit transaction is both unnecessary and disallowed. Keep `DROP TABLE IF EXISTS session_new` (line 405) as the crash-recovery guard. (If an explicit transaction were ever needed, the API is `ctx.storage.transactionSync(() => { … })`, which the current `(sql) => void` migration signature doesn't expose.)

**Test gap that hid this:** `schema.test.ts` uses a mock `SqlStorage` that only records call strings and asserts `"BEGIN"`/`"COMMIT"` appear — it never executes real SQLite, so it can't catch the rejection. Recommend an integration test (`@cloudflare/vitest-pool-workers`) that runs `initSchema` against an old `repo_owner TEXT NOT NULL` schema and asserts it succeeds and preserves existing rows.

Same failure mode as [workerd#5411](https://github.com/cloudflare/workerd/issues/5411) ("apply D1 migrations only fails in production").

terraform/d1/migrations/0029_allow_no_repository_targets.sql
```diff
@@ -0,0 +1,228 @@
+-- Allow automation and session records to intentionally have no repository.
+-- SQLite cannot drop NOT NULL constraints in place, so rebuild the tables.
+-- automation_runs references automations(id), so back it up without the foreign
+-- key before dropping automations and recreate the final table after the parent
+-- table has been renamed.
+
+CREATE TABLE automation_runs_backup (
```
ColeMurray: **[Automated Review] 🟠 Make this rebuild re-run-safe — the three temp tables lack `IF NOT EXISTS`.**

`automation_runs_backup` (line 7), `automations_new` (line 36), and `sessions_new` (line 161) are created without `IF NOT EXISTS`. D1 forbids explicit `BEGIN`/`COMMIT`, and `scripts/d1-migrate.sh` records version `0029` only *after* the whole `wrangler d1 execute --file` succeeds. So if the file fails partway (timeout, network blip, constraint error), `0029` isn't recorded and the next apply re-runs the file from the top — failing immediately at `CREATE TABLE automation_runs_backup` because the orphaned temp table already exists. That leaves a stuck migration needing manual D1 cleanup on prod (CI auto-applies on push to `main`).

No permanent data loss — the copy-before-drop ordering keeps the data in at least one table at all times — but there's a window where live `automations` / `automation_runs` / `sessions` queries can fail.

**Fix:** add `IF NOT EXISTS` to the three `CREATE TABLE` statements (and ideally `DROP TABLE IF EXISTS automation_runs_backup`, `…_new` at the top). The per-session DO migration already guards itself this way with `DROP TABLE IF EXISTS session_new`.

For context — the fidelity of the rebuild itself is correct: all `automations` (23) / `sessions` (21) / `automation_runs` columns, defaults, and effective indexes are preserved, and the two `automation_runs` indexes not recreated here (`idx_runs_automation_status`, `idx_runs_active_status`) were already dropped by migrations 0025/0024, so omitting them is right. This finding is purely about partial-failure recovery.

ColeMurray:
**[Automated Review] 🟡 Architecture (non-blocking): consider a discriminated union for `Automation` instead of widening repo fields to `| null`.**

A design call worth making consciously before merge — not a blocker.

The PR widens `repoOwner` / `repoName` / `baseBranch` to `string | null` across `Session` / `SessionState` / `Automation` / `ChildSessionDetail` (`packages/shared/src/types/index.ts`). Because `Automation` already carries the `targetMode` discriminant, independent-nullable modeling lets illegal states be representable at the type layer — e.g. `{ targetMode: "no_repository", repoOwner: "x" }` or `{ targetMode: "fixed_single_repo", repoOwner: null }` — so the "both repo fields exist, or neither does" invariant ends up enforced only by scattered runtime guards.

Consequences visible in this diff:
- The repo-presence predicate `!repo_owner || !repo_name` is hand-rolled across ~11 control-plane files; the one type guard introduced (`sessionHasRepository`, `sandbox/lifecycle/manager.ts:190`) isn't exported and isn't even reused in its own file (`resolveAgentSlackNotifyEnabled` re-inlines it).
- The `"fixed_single_repo"` default + normalize/validate logic is duplicated across the store mapper (`automationTargetMode`), `target-resolution.ts`, and the route; `toAutomation()` throws per-row on an unexpected `target_mode`, so a single bad row 500s the entire automations list endpoint.
- `"No repository"` is hardcoded ~8× even though `NO_REPOSITORY_LABEL` / `formatRepoLabel` now exists (and isn't used in its own package).

Since `targetMode` already exists and `toAutomation()` already branches on it, a sum type is nearly free and lets every consumer narrow on the discriminant to obtain non-null repo fields, deleting most downstream null checks:

```ts
type Automation = AutomationBase & (
  | { targetMode: "fixed_single_repo"; repoOwner: string; repoName: string; baseBranch: string; repoId: number | null }
  | { targetMode: "no_repository"; repoOwner: null; repoName: null; baseBranch: null; repoId: null }
);
```

Recommendation: do this now while it's cheap. The PR is framed as the first #833 slice with future *dynamic repo discovery* — without consolidating here, the next slice repeats the null-handling sweep across the same files. Sessions are a different case: a session's repo-ness is a *derived* fact, so keeping them nullable + one shared exported `sessionHasRepository` guard (rather than adding a `target_mode` column) is the right asymmetry.

Relatedly: `resolveAutomationTarget(env, automation)` (`automation/target-resolution.ts`) can't see the trigger event, but dynamic discovery will need it — and the event payload is already in scope at the `scheduler/durable-object.ts` call site. Threading a `triggerContext` parameter now (even if unused by today's two modes) would make the future mode an additive branch rather than a signature change to the seam this PR is introducing.

packages/control-plane/src/session/schema.ts
ColeMurray: **[Automated Review] 🟡 Beyond the `BEGIN` fix — consider dropping this migration rather than just correcting it.**

Removing `BEGIN`/`COMMIT`/`ROLLBACK` makes the rebuild *correct*. It's worth asking whether it should run at all.

Each session is its own DO, and no-repository sessions are always new sessions → fresh DOs created from the current (already-nullable) `SCHEMA_SQL`. Every *pre-existing* DO holds a repo-bound session whose repo columns are populated and always will be — so no old DO ever needs to store a null repo. The rebuild is schema-convergence hygiene, not a functional requirement.

And the payoff is thin here:
- Migrations run lazily on access, so dormant DOs never converge anyway — the system already tolerates old-schema tables.
- `repo_owner TEXT NOT NULL` on an old DO isn't *stale*; it accurately describes that session's data. We'd be rebuilding a live session table to drop a constraint that can never fire.

**Recommended:** drop the migration (the `notnull === 0` guard already no-ops fresh DOs) and document that pre-existing session DOs may retain `NOT NULL` on repo columns — harmless, since they never hold no-repo rows.
**Minimal alternative:** keep it, but remove `BEGIN`/`COMMIT`/`ROLLBACK` — synchronous `applyMigrations` is already atomic via the runtime's write-coalescing.

The only thing that makes the rebuild load-bearing is a near-term migration that writes `NULL` repos into *existing* session rows (e.g. in-place "convert to no-repo"). I don't see one in this PR — if that's not planned, dropping is the lower-risk call.

rubenlangeweg: > `sql.exec("BEGIN")` throws in the Durable Object runtime and bricks every pre-existing session on deploy.

Fixed in 69d7696 by removing the nullable-session rebuild migration entirely. Fresh session DOs already start from the nullable `SCHEMA_SQL`, and pre-existing DOs are repo-backed sessions that never need to store null repo fields, so the rebuild was unnecessary schema-convergence work. I also replaced the transaction-specific tests with a guard that migrations do not issue transaction-control SQL.

terraform/d1/migrations/0029_allow_no_repository_targets.sql
rubenlangeweg: > Make this rebuild re-run-safe — the three temp tables lack `IF NOT EXISTS`.

Fixed in 69d7696. The temp tables now use `IF NOT EXISTS`, and the copy steps use `INSERT OR REPLACE` so a re-run after a pre-drop failure does not fail on existing temp tables or duplicate primary keys. I intentionally did not add top-of-file `DROP TABLE` statements because after a mid-rebuild failure, the backup/temp table may be the only remaining copy; dropping it would make recovery worse.

rubenlangeweg:
> consider a discriminated union for `Automation` instead of widening repo fields to `| null`.

Done in 69d7696 and fd80ade.

`Automation` and `CreateAutomationRequest` are now discriminated unions on `targetMode`. Repo-backed automations require repo fields. `no_repository` automations require null repo fields.

I also moved `sessionHasRepository` into the shared session types for control-plane code and reused it in the lifecycle manager. The web package now uses its existing `NO_REPOSITORY_LABEL` constant instead of hardcoding that label in the touched UI.

I did not add `triggerContext` yet. The current modes do not use event payload, so that would be a dead parameter. We can add it in the dynamic discovery slice when there is a caller that needs it.

packages/control-plane/src/db/analytics-store.ts
```diff
@@ -142,14 +142,16 @@ export class AnalyticsStore {
...
 
     const groupExpression = isUserBreakdown
       ? "COALESCE(s.user_id, NULLIF(s.scm_login, ''), '__unknown__')"
-      : "s.repo_owner || '/' || s.repo_name";
+      : repoGroupExpression;
 
     const displayNameSelect = isUserBreakdown
       ? "COALESCE(MAX(NULLIF(u.display_name, '')), MAX(NULLIF(s.scm_login, '')), 'Unknown user') AS display_name,"
-      : "";
+      : `CASE WHEN ${repoGroupExpression} = '__no_repository__' THEN 'No repository' ELSE NULL END AS display_name,`;
```
ColeMurray: this looks like a magic constant

packages/control-plane/src/routes/slack-notify.ts
```diff
@@ -60,12 +60,13 @@ export async function handleSlackNotify(
...
 
-  const repo = `${session.repoOwner}/${session.repoName}`;
+  const repoScope =
+    session.repoOwner && session.repoName ? `${session.repoOwner}/${session.repoName}` : null;
   const audit: AuditFields = {
     prompt_author_user_id: session.userId ?? null,
     trigger_source: session.spawnSource ?? null,
     parent_session_id: session.parentSessionId ?? null,
-    repo,
+    repo: repoScope ?? "No repository",
```
ColeMurray: null feels better here vs magic string

packages/control-plane/src/sandbox/lifecycle/manager.ts
```diff
@@ -427,7 +438,7 @@ export class SandboxLifecycleManager {
 
       const mcpServers = await this.loadMcpServers(session);
 
-      const codeServerEnabled = session.code_server_enabled === 1;
+      const codeServerEnabled = hasRepository && session.code_server_enabled === 1;
```
ColeMurray: it seems probably fine to have the codeServer running even if there was no repository

packages/sandbox-runtime/src/sandbox_runtime/entrypoint.py
```diff
@@ -55,6 +55,13 @@ def _port_from_env(env_var: str, default: int) -> int:
     "slack-notify.js": "AGENT_SLACK_NOTIFY_ENABLED",
 }
 
+AGENT_TOOLS_REQUIRING_REPOSITORY = {
```
ColeMurray, line 58: these seem like they should be able to run without the repository

packages/control-plane/src/sandbox/lifecycle/manager.ts
rubenlangeweg: Yes, agree, resolved!

packages/control-plane/src/routes/slack-notify.ts
rubenlangeweg: agree, cleaner!

packages/control-plane/src/db/analytics-store.ts
rubenlangeweg: it is, i removed the sentinel and now group no-repo rows as NULL mapping to "No repository" only at the response boundry.

packages/sandbox-runtime/src/sandbox_runtime/entrypoint.py
rubenlangeweg: @ColeMurray , for now made a conservative change: status/change tools now install without a repo, while spanw-taks remains repo-gated because the child-spawn control plane path still requires a repo target. 

do you want this PR to also support spawning no-repo child sessions end-to-end, or keeping spawn repo-gated the right boundary ?

ColeMurray: if it doesn't introduce several widespread changes, let's proceed updating to allow no-repo child sessions. If it does cause a larger refactor, we can handle in a follow-up PR

ColeMurray:
I think `targetMode` is the wrong durable abstraction here, especially looking ahead to multi-repo support.

This PR currently introduces repository presence as an automation mode (`fixed_single_repo` vs `no_repository`), and the follow-on design extends that into cardinality (`fixed_single_repo`, `fixed_multi_repo`, `no_repository`). That feels like mode explosion: absence of repository context, number of repositories, trigger compatibility, and run materialization strategy all get packed into one enum.

I’d rather model this as optional repository context:

- session creation can omit repository context to create a repo-less session
- automations can store zero, one, or many repository targets
- event-scoped triggers can derive or require repository context at the trigger boundary
- the scheduler can derive launch behavior from resolved session launches:
  - zero repos: one repo-less session
  - one repo: one repo-backed session
  - multiple repos: one session per repo, grouped for history

That keeps “target” scoped to automation configuration and avoids passing repo mode through session/sandbox/runtime layers. The normalized target rows planned for multi-repo still seem useful, but I don’t think `target_mode` should be the authority. A helper like `resolveAutomationSessionLaunches(automation)` returning one or more session launch specs would give the scheduler the behavior it needs without making `fixed_single_repo` / `fixed_multi_repo` / `no_repository` part of the product/data model.

packages/sandbox-runtime/src/sandbox_runtime/entrypoint.py
rubenlangeweg: lets finish/address this no-repo child session AFTER the remove targetMode.

rubenlangeweg:
The recent commit addresses the direction change with focus on supporting `zero` or `one` repository, NOT many. PR #850 would need to be refactored/updated from targetMode to also align with the direction change and support `many`. 

> I think `targetMode` is the wrong durable abstraction here, especially looking ahead to multi-repo support.
> 
> This PR currently introduces repository presence as an automation mode (`fixed_single_repo` vs `no_repository`), and the follow-on design extends that into cardinality (`fixed_single_repo`, `fixed_multi_repo`, `no_repository`). That feels like mode explosion: absence of repository context, number of repositories, trigger compatibility, and run materialization strategy all get packed into one enum.
> 
> I’d rather model this as optional repository context:
> 
> * session creation can omit repository context to create a repo-less session
> * automations can store zero, one, or many repository targets
> * event-scoped triggers can derive or require repository context at the trigger boundary
> * the scheduler can derive launch behavior from resolved session launches:
>   
>   * zero repos: one repo-less session
>   * one repo: one repo-backed session
>   * multiple repos: one session per repo, grouped for history
> 
> That keeps “target” scoped to automation configuration and avoids passing repo mode through session/sandbox/runtime layers. The normalized target rows planned for multi-repo still seem useful, but I don’t think `target_mode` should be the authority. A helper like `resolveAutomationSessionLaunches(automation)` returning one or more session launch specs would give the scheduler the behavior it needs without making `fixed_single_repo` / `fixed_multi_repo` / `no_repository` part of the product/data model.

Issue #863, opened 2026-06-30
Linear agent is silent when workspace OAuth refresh fails


rubenlangeweg:
## Problem

A Linear agent session can appear to start, but the bot never posts any visible status or failure back to Linear when the workspace OAuth token is expired and the refresh token can no longer be used.

## Observed behavior

When a Linear `AgentSessionEvent` arrives for a workspace with an expired `oauth:token:<orgId>` KV entry and the Linear OAuth refresh endpoint returns `invalid_grant`, the worker logs the failure and returns without creating an agent activity, issue comment, or Open-Inspect session.

Repro evidence from a focused local Vitest probe against current `main`:

```text
agent_session.received action=created agent_session_id=agent-session-1 issue_identifier=ORI-229 org_id=org-1
oauth.refresh org_id=org-1
oauth.refresh_failed org_id=org-1 status=400 oauth_error=invalid_grant oauth_error_description="Refresh token has expired."
agent_session.no_oauth_token org_id=org-1 agent_session_id=agent-session-1
```

The same probe configured `LINEAR_API_KEY`, but no `commentCreate` GraphQL request was made. The only outbound call after the webhook was the failed OAuth refresh request.

## Root cause

The Linear bot collapses all OAuth-token failure modes into `null`, and the agent-session handlers treat that as a terminal silent return.

Code path:

1. `packages/linear-bot/src/utils/linear-client.ts#getOAuthToken` reads `oauth:token:<orgId>` from KV.
2. If the token is expired, it calls `https://api.linear.app/oauth/token` with the stored refresh token.
3. If refresh fails, e.g. `invalid_grant`, it logs `oauth.refresh_failed` and returns `null`.
4. `getLinearClient()` returns `null` when `getOAuthToken()` returns `null`.
5. `packages/linear-bot/src/webhook-handler.ts#handleNewSession` and `#handleFollowUp` call `getLinearClient()` before the first visible Linear activity.
6. On `!client`, both handlers only log `agent_session.no_oauth_token` and return.

That return happens before `updateAgentSession()`, `emitAgentActivity()`, repo resolution, control-plane session creation, or follow-up forwarding.

Related existing code: `packages/linear-bot/src/callbacks.ts` already has a `LINEAR_API_KEY` comment fallback for completion callbacks when AgentActivity cannot be posted, via `postIssueComment()` in `packages/linear-bot/src/utils/linear-client.ts`. The new-session and follow-up paths do not use an equivalent fallback.

## Expected behavior

[...]

PR #864, opened 2026-06-30
fix(linear-bot): classify Linear OAuth failures

rubenlangeweg:
## Summary

Linear agent-session auth failures now preserve the concrete reason they failed instead of collapsing every OAuth problem into a bare missing-client return.

When a workspace token is missing, malformed, expired without a refresh token, rejected by Linear during refresh, or blocked by a KV/read failure, the token loader raises a `LinearAuthError` with the failure reason and any available OAuth response details. Agent-session webhooks catch that error, log `agent_session.no_oauth_token` with the org, issue, trace ID, mode, and auth failure reason, then stop before creating or prompting an Open-Inspect session.

The final behavior is intentionally log-only for these agent-session paths. Earlier fallback issue-comment handling was removed after review because a failed OAuth refresh usually means operators should inspect the structured logs rather than trying to notify Linear through the static API-key fallback.

## Validation

- npm run lint
- npm run typecheck
- npm test

## Post-Deploy Monitoring & Validation

- Watch Linear bot logs for `agent_session.no_oauth_token`, `oauth.refresh_failed`, `oauth.refresh_error`, and `oauth.token_read_error`.
- Healthy signal: auth failures include `auth_failure_reason`, issue context, and trace ID, with no control-plane session or prompt created after the auth failure.
- Failure signal: repeated `token_read_error` / `refresh_error` spikes, missing issue context in `agent_session.no_oauth_token`, or unexpected Linear agent-session requests continuing after auth failure.
- Validation window: first deploy after merge; owner: operator handling the Linear bot deploy.

Related: #863

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT-5

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of expired or invalid Linear OAuth tokens, reducing failed agent session starts and follow-ups.
  * When authentication can’t be restored, the system now skips the action safely and records clearer failure details.
  * Refresh failures are now classified more consistently, helping diagnose issues like invalid grants and token read errors.

* **Tests**
  * Added coverage for token expiry, refresh success/failure, and agent session flows.


packages/linear-bot/src/utils/linear-client.ts
```diff
@@ -63,21 +63,65 @@ export async function exchangeCodeForToken(
...
+      ok: false,
+      reason: "malformed_token",
+      reauthorizationRequired: true,
+      retryable: false,
+    };
   }
 
   if (Date.now() < tokenData.expires_at - 5 * 60 * 1000) {
-    return tokenData.access_token;
+    return { ok: true, token: tokenData.access_token };
```
ColeMurray: Rather than having the ok:true/false, would prefer to raise an explicit exception so the caller can explicitly handle

rubenlangeweg: > Rather than having the ok:true/false, would prefer to raise an explicit exception so the caller can explicitly handle

Addressed: replaced the new result-union helpers with `getOAuthTokenOrThrow` / `getLinearClientOrThrow` and a typed `LinearAuthError`. The `AgentSession` path now catches that exception explicitly, logs the auth reason, and triggers the fallback notification flow; the existing nullable helpers still return `null` for current callers.

rubenlangeweg:
> `60 * 1000` is a duration literal; make the millisecond unit explicit in the identifier.

Addressed: added `EXPIRED_TOKEN_AGE_MS` in `packages/linear-bot/src/webhook-handler.test.ts` and updated the `expiredToken()` helper to use the named millisecond constant.

packages/linear-bot/src/webhook-handler.ts
```diff
@@ -173,6 +176,134 @@ async function createSession(
...
+  traceId: string;
+  orgId: string;
+  agentSessionId: string;
+  issue: AgentSessionWebhookIssue;
+  mode: AgentSessionAuthFailureMode;
+  authFailure: LinearAuthFailure;
+}): Promise<void> {
+  const { env, traceId, orgId, agentSessionId, issue, mode, authFailure } = params;
+
+  if (!env.LINEAR_API_KEY) {
```
ColeMurray: adding the linear fallback commenting here feels a bit overkill. if the client was unable to get a oauth token, it likely means that something bigger is going on and checking the logs seems a natural response.

can we slim down the change to just logging the error rather than adding the API key follow-up comment?

ColeMurray (approved):

PR #911, opened 2026-07-08
feat: Photo & Video attachment via web app

Blue101black:
# Summary
Added support for attaching images and videos via the web app. Linear and slack support coming soon.
image

https://github.com/user-attachments/assets/cd06c73f-10cc-48be-9485-0e9681de175a

##
Added `uploads` table to store the R2 uploads against the sessions.

Video support is done by breaking the video down into individual frames, that way agents can "view" the video.
Linear support (coming soon) is super nice because the linear default prompt actually describes what's happening throughout the video as well.

## Testing
Been using in production for my own company last couple of days. Video proof / pictures above.

## Follow Ups
Linear and slack support

### Note

This was vibe-coded with Fable 5, but tested in production and working well :) 

## Summary by CodeRabbit

* **New Features**
  * Added support for attaching images and videos to prompts through file selection, drag-and-drop, and paste.
  * Added upload validation, size limits, session quotas, cleanup, and secure media streaming.
  * Displayed attachment previews in the composer and rendered uploaded media in session timelines.
  * Added video processing and frame extraction for prompt attachments.

* **Bug Fixes**
  * Prevented duplicate submissions and preserved drafts when uploads fail.
  * Improved attachment validation, retry behavior, and warning messages.


packages/control-plane/src/session/http/handlers/uploads.handler.ts
```diff
@@ -0,0 +1,112 @@
...
+        raw = await request.json();
+      } catch {
+        return Response.json({ error: "Invalid request body" }, { status: 400 });
+      }
+      if (!isValidBody(raw)) {
+        return Response.json({ error: "Invalid upload record body" }, { status: 400 });
+      }
+
+      const timestamp = now();
+      const stale = deps.repository.takeStaleUnreferencedUploads(
```
ColeMurray: [Automated Review] This cleanup destroys the durable upload record before the fallible R2 deletion. If the later quota check returns 429, `staleObjectKeys` is never returned to the outer route; even on success, a failed R2 delete is only logged after the row is already gone. Those objects can no longer be discovered or retried, and the deployment has no bucket lifecycle policy to recover them. Please replace this destructive `take` with a retryable claim/delete/ack (or outbox/reaper) flow so ownership is retained until object deletion succeeds.

packages/control-plane/src/routes/session-prompt.ts
```diff
@@ -9,6 +10,46 @@ import { error, parsePattern, type Route } from "./shared";
...
+    return error(`You can attach up to ${MAX_ATTACHMENTS_PER_PROMPT} files per message`, 400);
+  }
+
+  const attachments: Attachment[] = [];
+  for (const item of raw) {
+    const result = attachmentSchema.safeParse(item);
+    if (!result.success) return error("Invalid attachment", 400);
+
+    const attachment = result.data;
+    if (attachment.content) {
```
ColeMurray, line 38: [Automated Review] This validates inline payloads against the 10/50 MB upload caps, then the message queue JSON-serializes them into the Durable Object row even though this PR explicitly documents a 2 MB row ceiling. A modest inline image can therefore pass validation and fail persistence; direct WebSocket prompts bypass these limits entirely. The underlying `Attachment` shape also permits arbitrary combinations of `content`, `url`, and `uploadId`, while upload references are silently marked without proving they exist or deriving MIME/kind from the stored row. Please use discriminated attachment variants and perform one canonical normalize/claim step at enqueue time; upload-backed messages should persist references only.

packages/sandbox-runtime/src/sandbox_runtime/bridge.py
```diff
@@ -922,12 +937,349 @@ def _transform_part_to_event(
...
+            inlined["content"] = base64.b64encode(data).decode("ascii")
+            self.log.info(
+                "prompt.upload_fetched",
+                attachment_name=name,
+                upload_id=upload_id,
+                size_bytes=len(data),
+            )
+            return inlined
+
+        results = await asyncio.gather(*(hydrate_one(attachment) for attachment in attachments))
```
ColeMurray: [Automated Review] This adds an unbounded media pipeline to an already 2,322-line `AgentBridge`: all attachments are downloaded concurrently, each video is joined in memory and base64-encoded, then decoded back to disk before another `gather` starts ffmpeg work concurrently. ffmpeg and ffprobe also have no process deadline. There is a much simpler structure here: extract an attachment processor, stream upload-backed videos directly to temporary files, use bounded concurrency, and wrap subprocesses with timeouts. That removes the encode/decode round trip and keeps HTTP/media policy out of the bridge.

packages/web/src/hooks/use-prompt-attachments.ts
```diff
@@ -0,0 +1,182 @@
...
+   * pending list is left intact so the user can retry.
+   */
+  const uploadAll = useCallback(async (sessionId: string): Promise<Attachment[]> => {
+    const pending = attachmentsRef.current;
+    if (pending.length === 0) return [];
+
+    setIsUploading(true);
+    setAttachmentError(null);
+    try {
+      return await Promise.all(
```
ColeMurray: [Automated Review] `Promise.all` makes this multi-file operation non-retry-safe. When one request rejects, the other requests continue and may successfully create uploads, but their returned IDs are discarded; retrying uploads every file again, consuming the permanent per-session quota. There is no DELETE/cancel route, and the stale cleanup only runs on a later upload. Please retain per-file uploaded state and reuse successful IDs, or introduce a batch draft/commit API with explicit cancellation so partial success has a recoverable owner.

packages/control-plane/src/routes/session-uploads.ts
```diff
@@ -0,0 +1,258 @@
...
+    return error("Session ID and upload ID are required", 400);
+  }
+  if (!/^[A-Za-z0-9-]+$/.test(uploadId)) {
+    return error("Invalid upload ID", 400);
+  }
+
+  const storage = createMediaObjectStorage(env);
+  const objectKey = buildPromptUploadObjectKey(sessionId, uploadId);
+
+  const rangeHeader = request.headers.get("Range");
```
ColeMurray: [Automated Review] This is a second implementation of the range/head/get/metadata/response-header flow already in `session-media-stream.ts`; only `parseByteRangeHeader` was shared. The two paths will drift on 404 handling, MIME fallback, logging, caching, or range semantics. Please extract a canonical `streamStoredMedia` helper parameterized by object resolution and allowed MIME types, then keep both routes as thin lookup adapters.

ColeMurray (commented):
[Automated Review] Thermo-nuclear code quality review.

packages/control-plane/src/session/http/handlers/uploads.handler.ts
Blue101black: Fixed in f5e2f27f. Stale uploads now use a durable claim/delete/ack flow: rows remain discoverable until R2 deletion succeeds, failed deletes release their claims, and abandoned claims become retryable after a timeout.

packages/control-plane/src/routes/session-prompt.ts
Blue101black: Fixed in f5e2f27f. Attachments are now discriminated by source, and both HTTP and WebSocket prompts pass through one queue-boundary normalization step. Upload references must exist and be unclaimed, MIME/kind are derived from the durable upload row, and persisted payload size is capped.

packages/sandbox-runtime/src/sandbox_runtime/bridge.py
Blue101black: Fixed in f5e2f27f. Media handling is extracted into AttachmentProcessor with bounded concurrency, upload-backed videos stream directly to temporary files, and ffmpeg/ffprobe subprocesses have deadlines.

packages/web/src/hooks/use-prompt-attachments.ts
Blue101black: Fixed in f5e2f27f. Successful per-file upload results are retained by pending attachment ID and session, so a retry reuses completed upload IDs instead of consuming quota again. Added a regression test for partial failure.

packages/control-plane/src/routes/session-uploads.ts
Blue101black: Fixed in f5e2f27f. Both artifact media and prompt uploads now use the canonical streamStoredMedia helper for full/range GETs, metadata validation, and response headers, with route-specific lookup and MIME policies.

Blue101black:
@ColeMurray resolved merge conflicts

packages/web/src/app/api/sessions/[id]/uploads/route.ts
```diff
@@ -0,0 +1,52 @@
+import { NextResponse } from "next/server";
+import { getServerSession } from "next-auth";
+import { authOptions } from "@/lib/auth";
+import { controlPlaneFetch } from "@/lib/control-plane";
+
+const SESSION_ID_PATTERN = /^[A-Za-z0-9-]+$/;
+
+// Control plane enforces exact per-kind caps (10 MB images / 50 MB videos);
+// this is just a cheap upper bound including multipart overhead.
+const MAX_UPLOAD_REQUEST_BYTES = 52 * 1024 * 1024;
```
Comment (ColeMurray) [inline packages/web/src/app/api/sessions/[id]/uploads/route.ts:10]: Confirmed on Vercel: uploading an 8 MB video through this route returns `413 Request Entity Too Large` / `FUNCTION_PAYLOAD_TOO_LARGE` before the handler can apply `MAX_UPLOAD_REQUEST_BYTES` (request `sfo1::gxq4l-1784152862480-04f44b2828b5`). Vercel Functions have a 4.5 MB request-body limit, so the advertised 10 MB image and 50 MB video caps are not reachable on the default Vercel deployment.

For comparison, Cloudflare-hosted web deployments do not hit the same platform-level blocker: Workers accept request bodies up to 100 MB on Free/Pro, 200 MB on Business, and 500 MB by default on Enterprise. There, the application limits dominate: 52 MiB for the multipart request, 10 MiB per image, and 50 MiB per video. However, Workers have a 128 MB memory limit per isolate, and this route buffers the entire multipart body before the control plane parses and copies it again. A 50 MiB video may therefore still be close to the practical memory ceiling, especially under concurrent uploads. Cloudflare limits: https://developers.cloudflare.com/workers/platform/limits/

Could we switch this flow to a direct client upload, such as a short-lived signed R2 URL, followed by server-side validation and finalization? That would avoid the Vercel hard limit and reduce Cloudflare Worker memory pressure. R2 supports single uploads up to 5 GiB: https://developers.cloudflare.com/r2/objects/upload-objects/

Vercel recommends direct-to-storage uploads for payloads over its limit: https://vercel.com/kb/guide/how-to-bypass-vercel-body-size-limit-serverless-functions

packages/sandbox-runtime/src/sandbox_runtime/attachment_processor.py
```diff
@@ -0,0 +1,341 @@
...
+    def info(self, event: str, **kwargs: Any) -> None: ...
+
+    def warn(self, event: str, **kwargs: Any) -> None: ...
+
+
+class AttachmentProcessor:
+    """Resolve prompt attachments with bounded I/O and subprocess concurrency."""
+
+    VIDEO_EXTENSIONS: ClassVar[set[str]] = {"mp4", "mov", "webm", "m4v", "avi", "mkv"}
+    MAX_VIDEO_FRAMES = 8
```
ColeMurray, line 27: i'm not sold on this approach of auto-selecting at most 8 frames from the video. It feels potentially confusing/misleading for a user.  Depending on the length of the video, it can potentially miss quite a lot of context and leave the user wondering why the model seems to fail to understand.

thoughts on dropping this from the PR?

packages/control-plane/src/session/http/handlers/uploads.handler.ts
```diff
@@ -0,0 +1,149 @@
...
+        return Response.json({ error: "Session not found" }, { status: 404 });
+      }
+
+      let raw: unknown;
+      try {
+        raw = await request.json();
+      } catch {
+        return Response.json({ error: "Invalid request body" }, { status: 400 });
+      }
+      if (isCompleteCleanupBody(raw)) {
```
ColeMurray, line 95: the cleanup behavior here feels a bit surprising / unexpected. In this case, the objects can exist indefinitely until another upload comes and triggers them.

it may be preferable to just not handle clean-up here and handle it separately in a (not yet existing) old session cleanup scheduled task

packages/sandbox-runtime/src/sandbox_runtime/attachment_processor.py
```diff
@@ -0,0 +1,341 @@
...
+            if upload_id:
+                ok = await self._download_upload_to_file(str(upload_id), video_path)
+            elif content:
+                try:
+                    video_path.write_bytes(base64.b64decode(content, validate=True))
+                    ok = video_path.stat().st_size <= self.MAX_VIDEO_BYTES
+                except (ValueError, OSError):
+                    ok = False
+            elif url:
+                ok = await self._download_url_to_file(str(url), self.MAX_VIDEO_BYTES, video_path)
```
ColeMurray, line 142: [Automated Review] **Do not let video attachments fetch arbitrary URLs.**

The normal web flow uploads a file and sends an `uploadId`, but the API also accepts a URL. For videos, the sandbox downloads that URL and follows redirects without checking where they lead.

This means someone who can submit a prompt could make the sandbox request an internal address such as `localhost` or a private service. Because the request comes from inside the sandbox, it may reach systems that the user could not access directly.

Please require videos to use trusted `uploadId` references. If remote URLs must remain supported, restrict them to approved hosts and validate every redirect before downloading.

ColeMurray:
Thanks @Blue101black for building and testing the attachment foundation here. We are going to take over the next iteration on a maintainer-owned branch based on the current PR head, so your existing commits and attribution will be preserved without substantially rewriting the branch in your fork.

For the replacement PR, we plan to narrow the initial scope to image attachments, apply a portable 4 MiB web upload limit for the default Vercel deployment, and remove the automatic eight-frame video pipeline and arbitrary remote-video download path. Direct browser-to-R2 uploads, explicit upload lifecycle cleanup, and a more deliberate video-understanding design will follow separately.

We will keep this PR open while preparing the replacement, link it here once it is ready for review, and then close this PR without merging so the original discussion and contribution history remain available. Thank you again for the substantial work that established this path.

packages/sandbox-runtime/src/sandbox_runtime/attachment_processor.py
Blue101black: Dropping video support completely? Opencode doesn't natively support videos so its an issue. What I liked is that it was better than nothing. 

What I like about linear is that they give an AI transcript of the video in the prompt with any videos you attached.

PR #914, opened 2026-07-08
fix(session): provider-agnostic participant presence name

duboff:
## Summary

- Participant presence names are derived as `scm_name || scm_login || user_id`. The `scm_*` fields are only populated for GitHub auth, so **non-GitHub providers (e.g. Google) fall through to the raw `user_id`** — surfacing the Google OAuth `sub` (a long numeric string) as the participant's display name in the UI.
- This plumbs a provider-agnostic display name end to end:
  - the web `ws-token` route forwards `authName` (`user.name`) for all providers
  - the control-plane `ws-token` route + handler persist it to a new `participants.auth_name` column (schema + migration, COALESCE update)
  - presence/name derivation now falls back `scm_name || scm_login || auth_name || user_id` in `durable-object.ts` and `message-queue.ts`
- GitHub behavior is unchanged (`scm_name` still wins); the new fallback only engages when the SCM fields are absent.

## Test plan

- [x] `npm test -w @open-inspect/control-plane` — added coverage in `ws-token.handler.test.ts` (persists `authName`; Google/`authName`-only path) and updated `repository.test.ts` for the new column/param
- [x] `tsc --noEmit` (control-plane + web)
- [ ] Manual: sign in with a non-GitHub (Google) provider, open a session, confirm the participants strip shows the display name instead of the numeric account id

## Summary by CodeRabbit

* **New Features**
  * Added an optional, provider-agnostic display name (`authName`) to session presence and WebSocket token requests.
  * The display name is now stored and included for presence and message author attribution.

* **Bug Fixes**
  * Improved participant display-name resolution using a consistent precedence order, including after reconnects and WebSocket hibernation recovery.

* **Chores**
  * Updated session persistence and validation, plus refreshed tests to cover the new `authName` field.


duboff:
Addressed both nitpicks:
- Extracted a `resolveParticipantName(...)` helper (new `session/participant-name.ts`) and used it at all four sites (durable-object.ts ×3, message-queue.ts ×1), so the fallback precedence lives in one place.
- Added a test for the update path with a non-null `authName` — asserts it reaches `updateParticipantCoalesce` for an existing participant.

ColeMurray:
Superseded by #959, which preserves the original commits and authorship, resolves the migration conflict against current main, and adds focused persistence/display fallback coverage.

PR #1008, opened 2026-07-14
feat(sandbox): add E2B sandbox provider

tomasvarga:
## Summary

Adds **[E2B](https://e2b.dev)** as a sandbox provider, alongside Modal / Daytona / Vercel /
OpenComputer. The control plane talks to the E2B REST API directly at runtime; a small
`packages/e2b-infra` package builds the sandbox template image.

## What's included

- **`E2BSandboxProvider`** implementing the `SandboxProvider` interface (create / resume / stop),
  wired into `provider-factory.ts` via `createE2BProviderFromEnv`.
- **`e2b-rest-client.ts`** — a hand-rolled client mirroring the E2B REST API
  (create / pause / resume / connect / set-timeout / kill, plus per-session env delivery over envd).
- **`packages/e2b-infra`** — builds the sandbox **template** (pinned toolchain Dockerfile + the
  `oi-launch` launcher + staged `sandbox-runtime`) via the E2B Template SDK; a Terraform module
  rebuilds it when the source changes.
- Multi-repo / environment sessions supported (nullable `repoOwner`/`repoName` + `repositories[]`).

## Lifecycle

- Stop is a **resumable pause** (`autoPause` + `autoResume`), so the shared lifecycle manager's
  persistent-resume path drives idle-pause / resume with no E2B-specific plumbing. A lapsed TTL
  pauses (recoverable) rather than kills; only a sandbox that never connected is killed. One config
  knob: **`E2B_AUTO_PAUSE`**.
- Per-session env is delivered via an envd file write — the template start command runs once at
  build, so the launcher waits for the file, then execs the supervisor.

## Testing

- Unit + workerd integration tests pass (including multi-repo).
- Verified **live end-to-end** on a deployed instance, via both the internal API and the web UI:
  session create → E2B sandbox spawn → bridge connect → agent runs in-sandbox → live `*.e2b.app`
  preview. See the demo below.

## Demo

e2b-ui-demo

## Summary by CodeRabbit

* **New Features**
  * Added **E2B** as a supported sandbox provider, including create/resume/stop with resumable pause and configurable timeout/auto-pause behavior.
  * Added support for session environment handoff and code-server/tunnel URL generation.
  * Added automation to build and pre-warm the E2B sandbox template via Terraform and CI.
* **Documentation**
  * Added E2B-specific setup, configuration, lifecycle, and troubleshooting guides; updated getting started and Terraform examples.
* **Tests**
  * Added unit test coverage for the E2B REST client and provider, including request composition, error mapping, and timeout handling.
* **Chores**
  * Updated CI/Terraform workflows to trigger on E2B infra changes and pass required E2B variables.


packages/control-plane/src/sandbox/e2b-rest-client.ts
```diff
@@ -0,0 +1,285 @@
...
+export interface E2BCreateSandboxParams {
+  templateID: string;
+  envVars?: Record<string, string>;
+  metadata?: Record<string, string>;
+  timeout?: number;
+  /** Pause (not kill) the sandbox when its timeout expires. */
+  autoPause?: boolean;
+  /** Wake a paused sandbox on inbound activity (only meaningful with autoPause). */
+  autoResume?: boolean;
+}
```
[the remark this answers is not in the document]
tomasvarga: Addressed in 2dd6b2be — but kept this **seconds-based** rather than converting to milliseconds.

Sandbox lifetime/TTL is a shared, cross-provider value standardized in seconds: `DEFAULT_SANDBOX_TIMEOUT_SECONDS` and `CreateSandboxConfig.timeoutSeconds` in `provider.ts`, and Modal/Daytona both pass `config.timeoutSeconds`. E2B's REST field is seconds too. Converting only E2B to ms would make it the lone outlier vs the `config.timeoutSeconds` it receives.

Fixed the actual violation (the bare `timeout`) by renaming `E2BCreateSandboxParams.timeout` → `timeoutSeconds` (wire key unchanged). `connectSandbox`/`setSandboxTimeout` already used `timeoutSeconds`, and the fetch/abort deadlines stay `*_MS` (`timeoutMs`).

packages/control-plane/src/sandbox/provider-factory.ts
```diff
@@ -112,7 +119,31 @@ function createDaytonaProviderFromEnv(env: Env): DaytonaSandboxProvider {
...
+
+  return createE2BProvider(client, {
+    scmProvider: resolveScmProviderFromEnv(env.SCM_PROVIDER),
+    codeServerPasswordSecret: env.E2B_API_KEY,
+    sandboxTimeoutSeconds: parseNumericEnv(
+      "E2B_SANDBOX_TIMEOUT_SECONDS",
+      env.E2B_SANDBOX_TIMEOUT_SECONDS,
+      DEFAULT_E2B_SANDBOX_TIMEOUT_SECONDS
+    ),
+    autoPause: parseBooleanEnv(env.E2B_AUTO_PAUSE, DEFAULT_E2B_AUTO_PAUSE),
```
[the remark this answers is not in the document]
tomasvarga: Done in 2dd6b2be — `parseBooleanEnv` now takes the variable name and accepts only `true`/`1`/`false`/`0`, otherwise throws `${name} must be a valid boolean` (mirrors `parseNumericEnv`). Added a `provider-factory.test.ts` case asserting `E2B_AUTO_PAUSE: "tru"` throws.

packages/control-plane/src/sandbox/providers/e2b-provider.test.ts
```diff
@@ -0,0 +1,337 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { computeHmacHex } from "@open-inspect/shared";
+import { E2BSandboxProvider, type E2BProviderConfig } from "./e2b-provider";
```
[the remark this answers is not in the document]
tomasvarga: Skipped: `packages/control-plane/vitest.config.ts` sets `environment: "node"` globally for the package, and no other test under `src/**` carries a `@vitest-environment node` pragma. Adding it only to this file would be redundant and inconsistent with the rest of the suite.

packages/control-plane/src/sandbox/e2b-rest-client.ts
```diff
@@ -0,0 +1,285 @@
...
+    if (!config.apiUrl) throw new Error("E2BRestClient requires apiUrl");
+    if (!config.apiKey) throw new Error("E2BRestClient requires apiKey");
+    if (!config.templateId) throw new Error("E2BRestClient requires templateId");
+    this.baseUrl = config.apiUrl.replace(/\/+$/, "");
+  }
+
+  async createSandbox(params: E2BCreateSandboxParams): Promise<E2BSandboxCreated> {
+    const startMs = Date.now();
+    try {
+      return await this.request<E2BSandboxCreated>("POST", "/sandboxes", TIMEOUT_CREATE_MS, {
```
ColeMurray, line 109: [Automated Review] **Make secure envd access an explicit invariant.**

The uploaded file contains `SANDBOX_AUTH_TOKEN` and user secrets, but this create payload never sends `secure: true`; `writeSessionEnv` silently performs an unauthenticated upload when no token is returned. The E2B SDK explicitly defaults secure access on, while the raw API field has no documented server default.

Please send `secure: true`, require a non-null `envdAccessToken` at this provider boundary, and remove the incorrect “null/absent on Hobby” assumption—null means the sandbox was created without secure access.

https://e2b.dev/docs/sandbox/secured-access

packages/e2b-infra/e2b.Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+# which stages packages/sandbox-runtime/src/sandbox_runtime and applies the COPY /
+# WORKDIR / start-command steps programmatically (API-key auth, no access token).
+#
+# Start command (set by build-template.py / Terraform, not ENTRYPOINT here):
+#   python /usr/local/bin/oi-launch
+
+FROM python:3.12-slim-bookworm
+
+# Pinned toolchain versions (keep in sync with daytona-infra/src/toolchain.py).
+ARG OPENCODE_VERSION=1.14.41
```
ColeMurray: [Automated Review] Could we bump this to `1.17.18` so the E2B image stays in parity with the other sandbox images? They were updated in [7f4e058b](https://github.com/ColeMurray/background-agents/commit/7f4e058be028eabf0d32df222ee413407e5ad419), but this Dockerfile is still on `1.14.41`.

packages/control-plane/src/sandbox/providers/e2b-provider.ts
```diff
@@ -0,0 +1,311 @@
...
+      });
+      const metadata = this.buildMetadata(config);
+      const sandbox = await this.client.createSandbox({
+        templateID: this.client.config.templateId,
+        metadata,
+        timeoutSeconds: config.timeoutSeconds ?? this.providerConfig.sandboxTimeoutSeconds,
+        autoPause: this.providerConfig.autoPause,
+        // Auto-resume tracks auto-pause: a recoverable pause you can't auto-wake
+        // would be pointless.
+        autoResume: this.providerConfig.autoPause,
```
ColeMurray: [Automated Review] Can we keep resume under the lifecycle manager instead of also enabling provider-side auto-resume? `resumeSandbox()` already calls `connectSandbox()` for the normal stopped/stale follow-up flow. With this enabled, E2B can also wake from ordinary traffic to an old tunnel URL, while `SessionDO` currently rejects sandbox WebSocket reconnects whenever the stored status is stopped or stale. That seems to allow a stale preview request to wake a sandbox outside the control plane’s state transition. Would setting `autoResume: false` preserve the intended manager-driven lifecycle?

terraform/environments/production/e2b.tf
```diff
@@ -0,0 +1,39 @@
...
+data "external" "e2b_source_hash" {
+  count = local.use_e2b_backend ? 1 : 0
+
+  program = ["bash", "-c", <<-EOF
+    cd ${var.project_root}
+    if command -v sha256sum &> /dev/null; then
+      hash=$(find packages/e2b-infra packages/sandbox-runtime/src \
+        -type f \
+        -not -path 'packages/e2b-infra/.venv/*' -not -path 'packages/e2b-infra/sandbox_runtime/*' -not -path '*/__pycache__/*' \
+        \( -name "*.py" -o -name "*.ts" -o -name "*.js" -o -name "Dockerfile*" -o -name "*.Dockerfile" -o -name "*.toml" -o -name "uv.lock" -o -name "*.sh" \) \
```
ColeMurray: [Automated Review] **Make the source hash cover the files that are actually staged.**

`build-template.py` copies the entire `sandbox_runtime` tree, but this hash only includes selected extensions. The existing `skills/**/SKILL.md` files are already excluded, so a skill-only change would not rebuild the E2B template and the deployed image could remain stale.

Can we hash the full staged runtime tree with exclusions only for generated/cache files, rather than maintaining an extension allow-list? Ideally this should reuse the same source-hashing policy as the other runtime-image builders.

ColeMurray:
[Automated Review]

This PR doesn’t contain support for Open Inspect’s prebuilt repository or environment images. E2B is currently excluded from the image-build workflow, advertises snapshots/restores as unsupported, and always starts from the configured base template rather than a selected prebuilt artifact.

This is important because teams use [`.openinspect/setup.sh`](https://github.com/ColeMurray/background-agents/blob/main/README.md#repository-lifecycle-scripts) for potentially expensive provisioning: installing dependencies, compiling the application, warming caches, downloading tooling, or preparing a local development database—for example, creating a SQLite database and running migrations or seed scripts.

With prebuilt images, that work runs once during the background image build and the resulting filesystem is reused. Without them, every fresh E2B session must clone the repository and rerun the full setup script before the agent is ready. For teams creating many sessions or child agents, this can repeatedly add several minutes to startup and introduce more opportunities for package downloads, migrations, or other external setup steps to fail.

This doesn’t necessarily need to block this PR—we can fast-follow separately—but it is critical functionality for the overall team experience. E2B supports reusable sandbox snapshots, so the underlying platform capability appears to exist. We should explicitly track connecting that capability to the image-build workflow and spawn-time prebuilt-image selection before considering the E2B provider at full feature parity.

ColeMurray (dismissed):

packages/e2b-infra/e2b.Dockerfile
tomasvarga: Done in f0f33985 — bumped `OPENCODE_VERSION` 1.14.41 → 1.17.18 to match `daytona-infra/src/toolchain.py` (`CODE_SERVER_VERSION` 4.109.5 and `AGENT_BROWSER_VERSION` 0.21.2 were already in sync).

terraform/environments/production/e2b.tf
tomasvarga: Done in f0f33985 — switched the source hash from an extension allow-list to **exclude-only** (mirroring the opencomputer builder), so it now hashes the whole staged `sandbox_runtime` tree and excludes only generated/cache dirs. Verified it now covers the previously-missed `skills/**/SKILL.md` files, so a skill-only change rebuilds the template.

packages/control-plane/src/sandbox/e2b-rest-client.ts
tomasvarga: Done in 1a7c44e6. Create now sends `secure: true`; `writeSessionEnv` **requires** the returned `envdAccessToken` (fails closed — kills the sandbox — if it's absent, so the session env is never uploaded anonymously); and the misleading "null on Hobby" comment is fixed.

Verified live on a deployed instance: a `secure: true` sandbox reaches `ready` (the authenticated env upload works), and an **unauthenticated** envd write now returns **401** — it returned 200 before this change.

packages/control-plane/src/sandbox/providers/e2b-provider.ts
tomasvarga: Done in 1a7c44e6 — set `autoResume: false` so resume stays manager-driven (`resumeSandbox` → E2B `/connect`); provider-side auto-resume is no longer enabled.

Verified live: idle → control-plane inactivity pause at ~10.6 min → a prompt resumes it (`connecting` → `ready` in ~11s). A spike also confirmed that with `autoResume: false` a stray request to a paused sandbox's tunnel does **not** wake it, while explicit `/connect` does — so resume can't happen outside the control-plane state machine.

tomasvarga:
Re: prebuilt-image / snapshot support — agreed, and thanks for calling it out explicitly.

This PR scopes E2B to the create / pause-resume lifecycle; it deliberately advertises `supportsSnapshots` / `supportsRestore: false` and stays out of the image-build workflow, so every session currently clones and reruns `.openinspect/setup.sh`. Prebuilt-image parity is the right fast-follow — E2B supports reusable snapshots (pause/resume already leans on persisted filesystem state), so the plan is to implement E2B snapshot create/restore, wire it into `image-builds` (+ `resolveImageBuildProvider`), and add spawn-time prebuilt-image selection, mirroring the Vercel / OpenComputer path.

Tracked in #1032 — will pick it up as a follow-up.

ColeMurray (approved):

ColeMurray:
thanks!

PR #1083, opened 2026-07-21
fix: cascade cancellation to nested tasks

ColeMurray:
## Summary
- add a `cancelNested` option to the `cancel-task` tool, defaulting to `true`
- recursively discover active descendants and cancel them deepest-first
- preserve target-only cancellation with `cancelNested: false`
- tolerate descendants that reach a terminal state during cancellation and report other partial failures
- add unit and integration coverage for recursive and opt-out behavior

## Verification
- `npm test -w @open-inspect/control-plane -- src/db/session-index.test.ts`
- `npm run test:integration -w @open-inspect/control-plane -- child-session-ops.test.ts`
- ESLint and Prettier checks on changed files

---
*Created with [Open-Inspect](https://open-inspect-prod.vercel.app/session/89add2663f40ceee91e5b14164b48dc5)*

## Summary by CodeRabbit

* **New Features**
  * Child task cancellation now supports deep nested cancellation by default (cascades to active descendants).
  * Added `cancelNested: false` to cancel only the specified child.
  * Updated the `cancel-task` tool to accept the `cancelNested` option and report cancelled nested counts.

* **Bug Fixes**
  * Improved request validation for `cancelNested` (malformed/invalid JSON now returns clear 400 errors).

* **Tests**
  * Extended unit/integration coverage for recursive descendant listing/cancellation, ordering, and error/edge cases (including tolerated conflict handling).

### DIFF
--- a/packages/control-plane/src/db/session-index.test.ts
+++ b/packages/control-plane/src/db/session-index.test.ts
@@ -146,7 +146,8 @@ class FakeD1Database {
       const descendants: Array<{ row: SessionRow; depth: number }> = [];
       let parents = [args[0] as string];
       let depth = 1;
-      while (parents.length > 0) {
+      // Mirrors the CTE's MAX_DESCENDANT_DEPTH cycle guard.
+      while (parents.length > 0 && depth <= 10) {
         const children = Array.from(this.rows.values()).filter((row) =>
           parents.includes(row.parent_session_id ?? "")
         );
--- a/packages/control-plane/src/db/session-index.test.ts
+++ b/packages/control-plane/src/db/session-index.test.ts
@@ -54,6 +54,7 @@ const QUERY_PATTERNS = {
   DELETE_SESSION: /^DELETE FROM sessions WHERE id = \?$/,
   SELECT_BY_PARENT:
     /^SELECT \* FROM sessions WHERE parent_session_id = \? ORDER BY created_at DESC$/,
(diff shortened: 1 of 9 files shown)


ColeMurray:
Follow-up commit `47ff3061` from local review:

- **Body validation**: replaced the hand-rolled JSON parse with `cancelChildSessionRequestSchema` in `@open-inspect/shared` (next to `spawnChildSessionRequestSchema`). Empty body still means defaults, so older body-less clients are unaffected.
- **Response semantics change**: the route now returns `200 {status: "cancelled", cancelledDescendantIds}` whenever the cascade did work — including the retry path where the direct child is already terminal (previously that returned the child's 409 even though descendants were just cancelled, which made `cancel-task` report `Cannot cancel` after doing real work). The 409 passthrough remains only when nothing was actionable; 502 partial failures now also list `cancelledDescendantIds`.
- **Tool output**: `cancel-task` reports how many nested tasks were cancelled.
- **Hygiene**: `TERMINAL_STATUS_SQL` derived from a `SessionStatus[]`-typed array; descendant CTE depth bounded at 10 as cycle insurance; unit test builds its route match via `parsePattern`; new integration test pins the terminal-child-with-no-active-descendants 409.

Verified: shared 447 unit tests, control-plane 2022 unit + 616 integration tests, typecheck (control-plane, shared, web), eslint + prettier on changed files.

PR #1087, opened 2026-07-21
feat(slack): forward image attachments from Slack messages to sessions

sameer-ahmd:
## Summary

The Slack bot currently forwards only message text. This PR makes images attached to Slack messages reach the agent as prompt attachments, using the session-attachments pipeline introduced in #1019 — the bot becomes a second producer alongside the web composer. No control-plane or sandbox-runtime changes.

**Flow**: Slack file (`url_private`, bot-token download) → multipart POST to the existing `/sessions/:id/attachments` store → `{attachmentId, name}` references on the prompt.

## What's covered

- **DMs** — including image-only messages: Slack delivers file uploads as the `file_share` subtype, possibly with no text, so DM dispatchability accepts them and an image-only message gets a synthesized prompt body.
- **App mentions** — `app_mention` events don't carry the message's `files` array, so files are recovered with a single-message `conversations.history`/`replies` lookup (`latest` + `inclusive=true` + `limit=1`; setting `oldest=ts` too would create a zero-width window Slack returns empty). The lookup runs concurrently with the channel-info fetch to stay off the critical path, and lookup failures are logged rather than silently treated as "no files".
- **Thread follow-ups** — images on replies are attached to the existing session's next prompt.
- **Clarification flow** — file metadata rides the pending request, so images survive the "which repo?" round-trip.

## Security

File objects arrive on webhook payloads, so their URLs are not trusted: the bot token is only ever sent to `https://slack.com` / `https://*.slack.com`, remote files (`mode: "external"`, whose `url_private` is registrant-supplied) are skipped, and downloads use `redirect: "manual"` so the token can't follow a redirect off Slack infrastructure. Bodies are streamed with a hard byte cap rather than buffered before the size check.

## Limits and failure behavior

The existing caps are enforced bot-side before upload (`MAX_SESSION_ATTACHMENTS_PER_MESSAGE`, `SESSION_ATTACHMENT_IMAGE_MAX_BYTES` — now single-sourced in shared, png/jpeg/webp/gif). Oversized files, failed downloads, and rejected uploads never block the message — each drop is logged with a reason, and once the prompt outcome is known the bot posts a threaded note telling the user how many images were dropped, with guidance matched to the cause (the `files:read` hint only when downloads failed).

## Setup

Requires the `files:read` bot scope (added to GETTING_STARTED; app reinstall needed).

## Testing

- Unit tests for the attachments module: download/upload flow, the token-protection cases (non-Slack hosts, external mode, redirects), streaming cap, drop accounting, and notice wording per reason.
- App-level tests for the follow-up upload path and the app_mention history recovery; unit tests for `getMessageFiles` (both lookup shapes, failure arm), `file_share` DM dispatch, launcher reference passing, and the pending-request file round-trip.
- Running in production on our own single-tenant deployment.

## Summary by CodeRabbit

* **New Features**
  * Forward Slack image attachments into sessions for direct messages, mentions, follow-ups, and new sessions, including image-only requests.
  * Added support for fetching attached files from Slack messages and thread replies for use in session prompts.
* **Bug Fixes**
  * Improved resilience for Slack file lookup and attachment handling failures—no longer throws on API errors and properly handles cases where sessions can’t start due to missing/undeliverable images.
* **Documentation**
  * Updated Slack setup and upgrade instructions to include the `files:read` scope (and `files:write` where applicable) for image forwarding.

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -376,7 +376,8 @@ Queued delivery applies to every Slack completion, including text-only replies. 
 
 1. Add **Account | Queues | Edit** to the Cloudflare API token used by Terraform. Terraform needs
    this permission to create the completion queue, dead-letter queue, Worker binding, and consumer.
-2. Add the Slack bot scope `files:write`, reinstall the app once for the workspace, and update
+2. Add the Slack bot scopes `files:write` and `files:read` (needed to forward images attached to
+   Slack messages into sessions), reinstall the app once for the workspace, and update
    `slack_bot_token` if Slack issued a replacement.
 3. Run `terraform apply`, then verify a text completion and a generated-media attachment. If the
    token lacks Queue access, the apply fails while provisioning the new resources; grant the
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -360,6 +360,7 @@ Skip this step if you don't need Slack integration.
    - `groups:read`
    - `im:history`
    - `im:read`
+   - `files:read` (lets the bot read images attached to messages and forward them to sessions)
    - `files:write`
    - `reactions:write`
 3. Click **"Install to Workspace"**
--- a/packages/shared/src/slack/client.ts
+++ b/packages/shared/src/slack/client.ts
@@ -404,6 +404,8 @@ export interface SlackMessageFile {
   url_private?: string;
   url_private_download?: string;
   size?: number;
+  /** "external" marks remote files whose url_private is third-party-hosted. */
+  mode?: string;
 }
 
 /**
@@ -412,14 +414,15 @@ export interface SlackMessageFile {
  * `app_mention` events don't include the message's `files` array, so when an
  * event arrives without files we recover them from conversation history. Pass
  * `threadTs` when the message is a thread reply; otherwise the top-level
- * message at `ts` is fetched.
+ * message at `ts` is fetched. Returns the failure arm on API errors so callers
+ * can distinguish "the message has no files" from "the lookup failed".
  */
 export async function getMessageFiles(
   token: string,
   channelId: string,
   ts: string,
   threadTs?: string
-): Promise<SlackMessageFile[]> {
+): Promise<SlackEnvelope<{ files: SlackMessageFile[] }>> {
   // Single-message fetch: `latest=<ts>&inclusive=true&limit=1`. Do NOT also set
   // `oldest=<ts>` — an equal oldest/latest is a zero-width window that Slack
   // returns empty for. See conversations.history docs ("Retrieving messages").
@@ -439,9 +442,9 @@ export async function getMessageFiles(
           inclusive: "true",
           limit: "1",
         });
-  if (!res.ok) return [];
+  if (!res.ok) return res;
   const message = res.messages?.find((m) => m.ts === ts);
-  return message?.files ?? [];
+  return { ok: true, files: message?.files ?? [] };
 }
 
 export interface SlackUser {
--- a/packages/shared/src/slack/client.ts
+++ b/packages/shared/src/slack/client.ts
@@ -395,6 +395,55 @@ export async function getThreadMessages(
   return { ok: true, messages };
 }
 
+/** A file object as it appears on a Slack message (subset of fields we use). */
+export interface SlackMessageFile {
+  id?: string;
+  name?: string;
+  title?: string;
+  mimetype?: string;
+  url_private?: string;
+  url_private_download?: string;
+  size?: number;
+}
+
+/**
+ * Fetch the files attached to a single message.
+ *
+ * `app_mention` events don't include the message's `files` array, so when an
+ * event arrives without files we recover them from conversation history. Pass
+ * `threadTs` when the message is a thread reply; otherwise the top-level
+ * message at `ts` is fetched.
+ */
+export async function getMessageFiles(
+  token: string,
+  channelId: string,
+  ts: string,
+  threadTs?: string
+): Promise<SlackMessageFile[]> {
+  // Single-message fetch: `latest=<ts>&inclusive=true&limit=1`. Do NOT also set
+  // `oldest=<ts>` — an equal oldest/latest is a zero-width window that Slack
+  // returns empty for. See conversations.history docs ("Retrieving messages").
+  type HistoryResult = { messages?: Array<{ ts?: string; files?: SlackMessageFile[] }> };
+  const res =
+    threadTs && threadTs !== ts
+      ? await slackGet<HistoryResult>(token, "conversations.replies", {
+          channel: channelId,
+          ts: threadTs,
+          latest: ts,
+          inclusive: "true",
+          limit: "1",
+        })
+      : await slackGet<HistoryResult>(token, "conversations.history", {
+          channel: channelId,
+          latest: ts,
+          inclusive: "true",
+          limit: "1",
+        });
+  if (!res.ok) return [];
+  const message = res.messages?.find((m) => m.ts === ts);
+  return message?.files ?? [];
+}
+
 export interface SlackUser {
   id: string;
   name: string;
--- a/packages/shared/src/slack/client.ts
+++ b/packages/shared/src/slack/client.ts
@@ -423,18 +423,23 @@ export async function getMessageFiles(
   ts: string,
   threadTs?: string
 ): Promise<SlackEnvelope<{ files: SlackMessageFile[] }>> {
-  // Single-message fetch: `latest=<ts>&inclusive=true&limit=1`. Do NOT also set
-  // `oldest=<ts>` — an equal oldest/latest is a zero-width window that Slack
-  // returns empty for. See conversations.history docs ("Retrieving messages").
+  // Single-message fetch. The two endpoints sort differently, so the window
+  // anchor differs: conversations.history returns newest-first, so
+  // `latest=<ts>&inclusive=true&limit=1` yields the target; conversations.replies
+  // returns oldest-first, so the anchor must be `oldest=<ts>&inclusive=true` (a
+  // `latest` anchor would yield the thread root instead). Never set oldest and
+  // latest to the same ts — an equal pair is a zero-width window that Slack
+  // returns empty for. `limit=2` on replies tolerates the thread root being
+  // included alongside the target; the find-by-ts below is the source of truth.
   type HistoryResult = { messages?: Array<{ ts?: string; files?: SlackMessageFile[] }> };
   const res =
     threadTs && threadTs !== ts
       ? await slackGet<HistoryResult>(token, "conversations.replies", {
           channel: channelId,
           ts: threadTs,
-          latest: ts,
+          oldest: ts,
           inclusive: "true",
-          limit: "1",
+          limit: "2",
         })
       : await slackGet<HistoryResult>(token, "conversations.history", {
           channel: channelId,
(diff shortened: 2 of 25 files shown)


sameer-ahmd:
@coderabbitai Addressed in e8bd9046:

- **Channel-context for image-only mentions**: `getChannelInfo` is now fetched unconditionally in `handleAppMention` — image-only mentions get channel name/description as classifier signal, and since `filesPromise` is awaited in the same `Promise.all`, this adds no latency.
- **Docstring coverage**: added doc comments to the remaining undocumented functions in the changed files (`handleIncomingMessage`, `handleAppMention`, `handleDirectMessage`, `attachmentName`, `downloadSlackFile`, `uploadToSession`).

sameer-ahmd:
been using this internally and i find this very important to the workflow

packages/shared/src/slack/client.ts
```diff
@@ -395,6 +395,58 @@ export async function getThreadMessages(
...
+  // Single-message fetch: `latest=<ts>&inclusive=true&limit=1`. Do NOT also set
+  // `oldest=<ts>` — an equal oldest/latest is a zero-width window that Slack
+  // returns empty for. See conversations.history docs ("Retrieving messages").
+  type HistoryResult = { messages?: Array<{ ts?: string; files?: SlackMessageFile[] }> };
+  const res =
+    threadTs && threadTs !== ts
+      ? await slackGet<HistoryResult>(token, "conversations.replies", {
+          channel: channelId,
+          ts: threadTs,
+          latest: ts,
```
ColeMurray: [Automated Review] Slack documents that conversations.replies returns the earliest messages in the oldest/latest range first. With ts set to the thread root, latest set to this reply, and limit=1, this returns the root rather than the target reply, so the timestamp lookup below produces no files for threaded mentions. Could we instead use oldest: ts with inclusive=true and omit latest, or paginate until the exact timestamp? The current unit test mocks two messages despite limit=1, which masks this behavior.

packages/slack-bot/src/attachments.ts
```diff
@@ -0,0 +1,315 @@
...
+  sessionId: string,
+  files: SlackMessageFile[] | undefined,
+  traceId?: string
+): Promise<SlackAttachmentUploadResult> {
+  const images = extractImageFiles(files);
+  if (images.length === 0) return { references: [], dropped: [] };
+
+  const references: SessionAttachmentReference[] = [];
+  const dropped: SlackAttachmentDropReason[] = [];
+  for (const file of images.slice(0, MAX_SESSION_ATTACHMENTS_PER_MESSAGE)) {
```
ColeMurray: [Automated Review] This serial loop can outlive the HTTP Worker background window. Each file may spend up to 10 seconds downloading plus 10 seconds uploading, so six files can take roughly 120 seconds, while ctx.waitUntil work is canceled 30 seconds after the response. The event has already returned 200 and stored its dedupe key, so cancellation can lose the prompt without a retry. Could this work move to a durable queue, or otherwise be bounded to finish within the waitUntil window?

packages/slack-bot/src/events/message-handler.ts
```diff
@@ -118,6 +134,8 @@ async function handleIncomingMessage(params: IncomingMessageParams): Promise<voi
     );
     return;
   }
+  // An image-only message still needs prompt content for the agent to act on.
+  const promptText = messageText || "See the attached image(s).";
```
ColeMurray: [Automated Review] Product edge case: when this is an image-only message and every image is dropped because files:read is missing, a file is oversized, or an upload is rejected, promptText is still "See the attached image(s)." and sendPrompt is called with an empty attachment list. That launches a meaningless run before the warning is posted. Should image-only requests require at least one successful reference, or should a test document the intended fail-soft behavior?

docs/GETTING_STARTED.md
```diff
@@ -360,6 +360,7 @@ Skip this step if you don't need Slack integration.
    - `groups:read`
    - `im:history`
    - `im:read`
+   - `files:read` (lets the bot read images attached to messages and forward them to sessions)
```
ColeMurray, line 363: [Automated Review] files:read is required for url_private downloads, but this update only adds it to the new-install scope list. The existing-deployment upgrade steps below still instruct operators to add only files:write, and terraform/README.md also omits files:read. Could we update those canonical upgrade and scope lists so existing deployments actually enable this feature?

packages/slack-bot/src/pending-requests/pending-request-store.ts
```diff
@@ -4,12 +4,24 @@ import type { Env } from "../types";
...
 const PENDING_REQUEST_TTL_MS = 60 * 60 * 1000;
 
+const pendingFileSchema = z.object({
+  id: z.string().optional(),
+  name: z.string().optional(),
+  title: z.string().optional(),
+  mimetype: z.string().optional(),
+  url_private: z.string().optional(),
+  url_private_download: z.string().optional(),
+  size: z.number().optional(),
```
ColeMurray: [Automated Review] pendingFileSchema omits mode, and Zod strips unknown keys by default, so mode: "external" is lost across the clarification and KV round-trip. The Slack-host allowlist still prevents token exfiltration, but the explicit remote-file skip no longer applies on that path. Could we preserve mode here, or discard external files before persistence, and cover the round-trip?

packages/slack-bot/src/events/message-handler.ts
```diff
@@ -91,11 +98,18 @@ interface IncomingMessageParams {
   threadTs?: string;
   channelName?: string;
   channelDescription?: string;
+  /** Files attached to the Slack message; images are forwarded to the session. */
+  files?: SlackMessageFile[];
```
ColeMurray: [Automated Review] This makes the raw Slack transport shape the attachment model across dispatch, message handling, pending KV state, target selection, session launch, and upload. The same files are re-filtered in several layers, while pendingFileSchema independently re-declares the shape and has already drifted. Also, storePendingRequest serializes the original runtime objects before any schema parse, so extra Slack payload fields and private URLs enter KV. Can we normalize once at ingress into a validated SlackImageAttachment model and make downstream APIs accept only that? For clarification state, persisting file IDs or the source-message locator and re-fetching at delivery would avoid storing raw URL-bearing objects entirely.

packages/slack-bot/src/sessions/control-plane-client.ts
```diff
@@ -110,7 +111,8 @@ export async function sendPrompt(
   content: string,
   authorId: string,
   callbackContext?: CallbackContext,
-  traceId?: string
+  traceId?: string,
+  attachments?: SessionAttachmentReference[]
```
ColeMurray: [Automated Review] Attachment delivery is now an implicit protocol duplicated in the existing-session and new-session paths: upload, send the prompt, notify about drops only after success, and interpret stale or transient failure. sendPrompt has also grown to seven positional arguments, with attachments after two optionals. Can we put this behind one typed delivery operation using an options object, so it owns upload -> prompt -> notification and returns the typed stale or transient outcome? That would remove the duplicated sequencing and make the attachment boundary explicit.

packages/shared/src/slack/client.ts
sameer-ahmd: Fixed in a994f9c2 — you're right, `conversations.replies` is oldest-first, so the `latest` anchor returned the thread root and the find-by-ts came up empty for threaded mentions. The lookup now anchors on `oldest=<ts>&inclusive=true` (with `limit=2` to tolerate the parent being included in the page; find-by-ts stays the source of truth). The unit test was rewritten to mock realistic ordering — target first, later reply second — instead of the two-message shape that masked this.

packages/slack-bot/src/attachments.ts
sameer-ahmd: Addressed in d5d57889 by bounding the work rather than adding a queue: the per-file download and upload pipelines now run concurrently (`Promise.all` over the ≤6 capped files, order preserved), so attachment overhead is bounded by one 10s download timeout plus one 10s upload timeout regardless of file count — down from ~120s worst case to ~20s, inside the waitUntil window. A durable queue would be the fully robust answer but pulls in new infra (Queues binding + Terraform + consumer), which felt out of scope for this PR; happy to follow up with that if you'd prefer it.

packages/slack-bot/src/events/message-handler.ts
sameer-ahmd: Went with requiring at least one successful reference — d5d57889. Image-only requests that lose every image no longer launch a run: the launcher downloads image bytes *before* `createSession` and bails (with a "didn't start on this request" notice) when nothing survives, so no orphan session is created either; on the follow-up path, `deliverPrompt` skips the prompt when zero references survive the upload. One subtlety: when the uploads all fail with 404 the failure is stale-session noise, not a real drop, so that case maps to the existing stale outcome and the thread retries against a fresh session instead of telling the user their image was unreadable. Covered by unit tests at the delivery/launcher/target-selection layers plus an end-to-end test (`starts nothing for an image-only DM whose images all fail to download`).

docs/GETTING_STARTED.md
sameer-ahmd: Fixed in b6664dcd — `files:read` is now in the existing-deployment upgrade steps in docs/GETTING_STARTED.md and in both the scope list and upgrade note in terraform/README.md.

packages/slack-bot/src/pending-requests/pending-request-store.ts
sameer-ahmd: Resolved in d5d57889 by removing the file objects from KV entirely (per the suggestion on the broader modeling comment): pending requests now persist only a source-message locator (`{ ts, threadTs }`) plus an `imageOnly` flag, and the files are re-fetched from Slack at delivery and re-normalized through the same ingress validation (which drops `mode: "external"` files). With no file shape persisted, this drift class can't recur; `storePendingRequest` also parses through the schema before writing so only schema-known fields reach KV. Round-trip covered in the target-selection tests, including the image-only case where the re-fetch fails.

packages/slack-bot/src/events/message-handler.ts
sameer-ahmd: Done in d5d57889. Raw `SlackMessageFile` payloads are now normalized exactly once at event ingress (`toImageAttachments`: supported mime type, Slack-hosted https URL, external/remote files excluded) into a validated `SlackImageAttachment` model, and every downstream API — session launch, download, upload, delivery — accepts only that type. The transport shape stays confined to the dispatcher/handlers and the shared Slack client. For clarification state I took the re-fetch option: pending requests persist only the source-message locator and files are re-fetched and re-normalized at delivery, so no URL-bearing objects are stored (details on the pending-store thread). The trusted-host check is still re-verified at download time as defense in depth.

packages/slack-bot/src/sessions/control-plane-client.ts
sameer-ahmd: Done in d5d57889. There's now a single typed operation, `deliverPrompt(env, options)` in `sessions/prompt-delivery.ts`, that owns upload → prompt → drop-notification and returns a typed outcome (`ok` | `stale` | `transient` | `no_images_delivered`); both the follow-up path and the session launcher call it, so the sequencing (including notify-only-after-success) lives in one place. `sendPrompt` itself now takes an options object instead of seven positional arguments. The download stage is exposed separately (`prepareImageAttachments`) because the launcher runs it before `createSession` to avoid creating a session for an image-only request whose images are all lost.

ColeMurray:
thanks!

PR #1107, opened 2026-07-25
Fix/189 cascade child cancellation

mauricedesaxe:


mauricedesaxe:
Hey, sorry for this, a mistake from my agent that I didn't catch in time. I'll put better guardrails in place so it doesn't happen again 🙏🏻

PR #1143, opened 2026-07-28
fix(slack): forward shared message content to the agent

duboff:
## Summary

Forwarding a Slack message into a request left the agent with only the comment typed alongside it.

Slack puts a shared message's content in the new message's `attachments` array — flagged
`is_share`/`is_msg_unfurl`, carrying its own `author_name`, `channel_name`, `channel_id`, `ts`,
`from_url`, `text`, `fallback`, and `files` — and never in `text`. The bot read only `text`, so
`@bot deal with this` plus a forwarded incident report reached the agent as `deal with this`. The
`attachments` field was already typed on the event payload in `events/dispatcher.ts` but never read
anywhere.

Everything the attachment carries now reaches the session:

- **Body and links** — quoted verbatim under the author and source channel, in a
  `Slack messages forwarded with this request` block matching the existing thread/channel context
  blocks. The quote leads and the user's own text lands last, where it reads as the instruction.
- **Images** — the shared message's `files` are Slack-hosted message files in the same shape as a
  message's own, so they go through the existing normalize → download → session-attachment path and
  share its caps.
- **Provenance** — permalink, channel id, and message ts on a `Source:` line, so an agent with Slack
  tooling of its own can pull the original thread for wider context.

Also:

- A forward with no comment of its own is now a complete request — previously such a DM was dropped
  by `isDmDispatchable` as having no content.
- Link unfurls land in the same array and are skipped: they restate a link the message text already
  carries, and their `text` is usually boilerplate (`"Monitor your errors"`).
- `getMessageFiles` becomes `getMessageDetails` and also returns the message's attachments, since
  `app_mention` events may omit them just as they omit `files`. The mention path fills gaps from the
  single-message lookup it already makes; whatever the event carried wins.
- Bodies are capped at 10 messages and 4000 chars each so one forward can't dominate the prompt.

Field shapes were taken from real `conversations.history` payloads for both explicit forwards and
pasted-permalink unfurls, including one whose share carried two `image/png` files.

Not covered: Slack Message automation triggers still match on the message's own text only. That path
normalizes events for the control plane's condition matching, so including forwarded bodies would
change trigger semantics — documented as a limitation instead.

## Test plan

- [x] `npm test -w @open-inspect/slack-bot` (354 passed) and `npm test -w @open-inspect/shared`
      (500 passed)
- [x] New `forwarded-messages.test.ts` unit tests cover attribution, the source line, verbatim links,
      shared files, image-only shares, the `fallback` path for shared app posts, link-unfurl
      skipping, caps, and truncation
- [x] New end-to-end `index.test.ts` cases: a mention whose forwarded body, source ids, and image are
      recovered from conversation history, and a comment-less forwarded DM. Both fail on `main`
- [x] `npm run typecheck` for shared + slack-bot, `npm run lint:fix`, `npm run format`

## Summary by CodeRabbit

* **New Features**
  * Forwarded Slack messages (shared/unfurl), including quoted source/author context and any forwarded images, are now incorporated into prompts for DMs and mentions—even when no user text is provided.
  * Mention and DM handling now recognizes forwarded-message content carried in Slack attachments.
  * Message lookups now retrieve both files and attachments to improve forwarded-image recovery.

* **Documentation**
  * Expanded Slack forwarding guidance (prompt transformation, truncation rules, and skipped link previews), plus updated channel-trigger limitations and troubleshooting permissions.

* **Tests**
  * Added coverage for forwarded-message collection and updated Slack message-detail retrieval expectations.


duboff:
Thanks — all three applied in 3d0e79c9:

- Whitespace-only `text` now counts as absent, so a usable `fallback` is no longer shadowed (a real drop for shared app posts). Added a test.
- The truncation marker counts against `MAX_FORWARDED_BODY_LENGTH` instead of pushing the body past it; the test now asserts the exact cap.
- Docs state the concrete limits: ten forwarded messages per request, 4,000 characters per shared body.

ColeMurray (approved):

PR #1145, opened 2026-07-28
fix(slack-bot): split long answers across sections instead of truncating at 2000 chars

jackmiller92:
## Problem

Completion messages put the agent's whole response into a **single** Slack section block capped at 2000 characters (`packages/slack-bot/src/completion/blocks.ts`):

```ts
const TRUNCATE_LIMIT = 2000;
const text = truncateForSlack(response.textContent, TRUNCATE_LIMIT);
```

Slack allows **3000 characters per section block** and **50 blocks per message**, so a threaded reply can carry far more than 2000 characters. Any answer longer than that stops mid-sentence — I hit this asking a research question whose answer cited several sources, and the reply was cut off partway through the final paragraph.

Raising the constant to 3000 would only move the wall, so this splits instead.

## Change

- Splits the response on paragraph boundaries into consecutive section blocks, so headings stay with their prose.
- Oversized paragraphs (long tables, large code blocks) fall back to line boundaries, then to a hard slice for a single unbroken line.
- Fenced code blocks are closed at the end of a section and reopened at the start of the next, so a split inside a fence can't leak monospace formatting through the rest of the message.
- Past `MAX_RESPONSE_SECTIONS` (20) the tail is truncated with a pointer to the session — the existing **View Session** button already serves as the escape hatch, so nothing is unreachable.
- Empty responses still render the `_Agent completed._` placeholder.
- Removes the now-unused local `truncateForSlack` helper (the identically-named export in `@open-inspect/shared` is a different function and is untouched).

Behaviour for responses under 3000 characters is unchanged: one section block, same as before.

## Tests

`packages/slack-bot/src/completion/blocks.test.ts` gains five cases:

- a long multi-paragraph answer survives whole across several sections
- no section exceeds Slack's per-section limit
- the block budget is enforced, with a truncation marker and the session link
- code fences stay balanced when a fenced block spans a split
- an empty response still renders the placeholder

`vitest run src/completion/blocks.test.ts` → 11 passed (6 existing + 5 new). `npm run lint` and `npm run typecheck` clean.

Note: `packages/slack-bot/src/index.test.ts` has 10 pre-existing failures on `main` (POST /events, POST /interactions) unrelated to this change — verified they fail identically without it.

## Summary by CodeRabbit

* **New Features**
  * Long Slack responses are now split across multiple readable sections while preserving paragraph formatting and code blocks.
  * Responses respect Slack’s section size limits and provide a link to view the full session when content exceeds the message limit.

* **Bug Fixes**
  * Empty or whitespace-only responses now display an “Agent completed.” placeholder.
  * Improved handling prevents content loss when splitting lengthy responses.

### DIFF
--- a/packages/slack-bot/src/completion/blocks.test.ts
+++ b/packages/slack-bot/src/completion/blocks.test.ts
@@ -181,3 +181,86 @@ describe("buildCompletionBlocks", () => {
     expect(createPrButton?.url).toBe(fallbackUrl);
   });
 });
+
+describe("long response handling", () => {
+  // Regression: responses were capped at 2000 chars in a single section, so
+  // multi-part answers (headings + citations) stopped mid-sentence even though
+  // Slack accepts 3000 per section and 50 blocks per message.
+  const sectionTexts = (blocks: ReturnType<typeof buildCompletionBlocks>): string[] =>
+    blocks
+      .filter((block) => block.type === "section")
+      .map((block) => block.text?.text ?? "")
+      // Drop the artifacts section, which is built separately.
+      .filter((text) => !text.startsWith("*Created:*"));
+
+  it("keeps a long multi-paragraph answer whole across several sections", () => {
+    const paragraphs = Array.from(
+      { length: 8 },
+      (_, i) => `## Section ${i}\n\n${"detail ".repeat(120)}`
+    );
+    const textContent = paragraphs.join("\n\n");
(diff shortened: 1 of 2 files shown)


duboff:
Superseded in https://github.com/ColeMurray/background-agents/pull/1147 (we're part of the same org)

ColeMurray:
thanks!

PR #1147, opened 2026-07-28
fix(slack): split long completions across sections instead of truncating at 2000 chars

duboff:
## Summary

Completion messages were capped at 2000 characters in a single `section` block, so any long answer stopped mid-sentence — even though Slack accepts 3000 characters per section and 50 blocks per message. Multi-part answers (headings plus citations, or a summary followed by a code sample) lost their tail.

`buildCompletionBlocks` now splits the response across as many section blocks as it needs, preferring paragraph boundaries, falling back to line boundaries, then to a hard slice. A fenced code block that lands on a split is closed at the end of one section and reopened — with its language — at the start of the next, so monospace formatting doesn't leak across the rest of the message. Past a 20-section budget the tail is truncated with a pointer to the session, which leaves comfortable headroom under Slack's 50-block limit.

Two details that are easy to get wrong, and which the tests here pin:

- **The fence repair costs characters.** The closing fence adds 4 and the reopen prefix adds `3 + len(info) + 1`. Measuring only the section body lets an in-fence section render at 3004 against Slack's 3000 cap — and Slack rejects an oversized section by failing the whole message, so a long fenced answer posts *nothing* rather than being trimmed. Every fit check therefore measures what Slack actually receives: reopen prefix + body + closing fence.
- **The hard-slice path must advance by what it kept.** Slicing the reopen-prefixed string to the cap while advancing the cursor by the full cap silently drops the characters the prefix displaced — content vanishing from the middle of a code block, not a truncated tail.

The fence info string is bounded to its first whitespace-delimited token (32 chars) because `reopenPrefix` re-emits it on every continuation section, so an unbounded capture from a pathological opener line could push sections past the cap.

## Test plan

- [x] `packages/slack-bot/src/completion/blocks.test.ts` — 18 tests pass
- [x] Long multi-paragraph answer arrives whole across several sections, with no truncation marker
- [x] No section exceeds 3000 characters, including when a split falls inside a fence
- [x] No characters are lost when hard-slicing a >3000-char line inside a fence
- [x] Fences stay balanced in every section, including the truncated final one
- [x] Continuation sections reopen the fence with the original language
- [x] Tail truncation past the section budget still renders the View Session button
- [x] Empty/whitespace responses still render the `_Agent completed._` placeholder
- [x] `prettier --check` clean on the touched files

Most of this is @jackmiller92's work; the character-budget accounting is mine.

## Summary by CodeRabbit

* **New Features**
  * Long Slack completions now render across multiple section blocks instead of truncating into one.
  * Code fences (including the language/info line) are preserved correctly across section boundaries.
  * Empty/whitespace responses show a friendly “completed” placeholder; truncated replies include a clear truncation indicator and a view option for the remainder.
* **Bug Fixes**
  * Prevents oversized sections and ensures generated blocks stay within Slack character/block limits while keeping fences balanced and well-formed.
* **Tests**
  * Added regression tests for long, multi-paragraph splitting, truncation behavior, fence correctness, and strict limits.

### DIFF
--- a/packages/slack-bot/src/completion/blocks.test.ts
+++ b/packages/slack-bot/src/completion/blocks.test.ts
@@ -293,6 +293,31 @@ describe("long response handling", () => {
     }
   });
 
+  // The truncation cut can land inside a fence the final section both opened and
+  // closed, which a trailing-fence check cannot detect. Sweep the section length
+  // through the window where slicing actually happens, moving the fence across the
+  // cut point, and assert the invariant holds for every shape.
+  it("keeps fences balanced when the truncation cut lands inside a fence", () => {
+    const fence = "```";
+    for (let sectionLen = 2949; sectionLen <= 3000; sectionLen += 1) {
+      for (const codeLen of [20, 45, 200]) {
+        const block = `\n${fence}ts\n${"h".repeat(codeLen)}\n${fence}`;
+        const fill = sectionLen - block.length;
+        if (fill < 1) continue;
+        const paragraphs = [
+          ...Array.from({ length: 19 }, () => "f".repeat(2900)),
+          "g".repeat(fill) + block,
+          ...Array.from({ length: 5 }, () => "z".repeat(2900)),
+        ];
+        const sections = splitIntoSlackSections(paragraphs.join("\n\n"));
+        const last = sections[sections.length - 1];
+        expect(last).toContain("truncated");
+        expect(last.length).toBeLessThanOrEqual(3000);
+        expect((last.match(/```/g) ?? []).length % 2).toBe(0);
+      }
+    }
+  });
+
   it("carries the fence language across a split", () => {
     const sections = splitIntoSlackSections(`\`\`\`python\n${"c".repeat(6500)}\n\`\`\``);
     expect(sections.length).toBeGreaterThan(1);
--- a/packages/slack-bot/src/completion/blocks.test.ts
+++ b/packages/slack-bot/src/completion/blocks.test.ts
@@ -312,3 +312,42 @@ describe("long response handling", () => {
     expect(sectionTexts(blocks)[0]).toBe("_Agent completed._");
   });
 });
+
+describe("fence info bounding", () => {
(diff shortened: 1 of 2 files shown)


duboff:
Cross-referencing the review on #1145 (the earlier PR for this change, from the same org — superseded by this one). CodeRabbit raised two findings there. Both are addressed here, so noting how, since that review isn't visible on this PR:

**1. `appendChunk` reused the paragraph separator for line-continuation merges**, which could insert a spurious blank line into a reconstructed oversized paragraph. Splitting now passes an explicit separator per token — `\n\n` at a paragraph boundary, `\n` for a line continuation within one. The same comment suggested extracting the fence-tracking state; it now lives in an immutable `FenceState` with `advanceFence` / `reopenPrefix` / `closeSuffix`, which removed the mutable-closure ordering dependency between the old `appendChunk` and `consumeFences`.

**2. The `maxSections` truncation slice could cut through a fenced block**, unbalancing the final section. Handled in `withTruncationMarker`, and hardened further in 42cd639f: it now counts the fence markers in the slice and closes when the count is odd, rather than relying on a trailing-fence check — a cut can land inside a fence the section both opened *and* closed, which a trailing check cannot see. The suggested regression test is there too, sweeping the section length across the window where slicing actually happens with the fence at three offsets relative to the cut.

For transparency: I could not construct an input that failed the previous trailing-fence logic across a 156-shape sweep, so that second change is hardening rather than a fix for an observed break.

Two further fixes in this PR that weren't in #1145, both with tests:

- **In-fence sections could render at 3004 against the 3000 cap.** The closing fence and reopen prefix are added *after* the fit check, so a section filled to the cap overflowed. Slack rejects an oversized section by failing the whole message, so a long fenced answer posted nothing at all. Fit checks now measure reopen prefix + body + closing fence.
- **The hard-slice path dropped characters.** It sliced the reopen-prefixed string to the cap but advanced the cursor by the full cap, silently losing whatever the prefix displaced — 12 of 7000 characters in a fenced payload. It now advances by exactly what it kept.

PR #1191, opened 2026-07-30
feat(slack): workspace session instructions for Slack-initiated sessions

sameer-ahmd:
## Summary

Adds a **Session Instructions** field to the Slack integration settings (Settings → Integrations → Slack), mirroring Linear's *Issue Session Instructions*. When set, the instructions are appended to the first prompt of every Slack-initiated session as an `## Additional Instructions` section — standing guidance like coding standards, preferred tools, or PR conventions.

## Implementation (follows the Linear pattern)

- **shared**: `sessionInstructions?: string` on `SlackGlobalSettings` (global-only, like `mentionsPolicy`/`routingRules`). New `MAX_SESSION_INSTRUCTIONS_LENGTH = 10000` constant, now used by both the Linear and Slack validators instead of two inline `10000` literals.
- **control-plane**: added to the strict Slack settings allowlist (global level only — rejected per-repo); validated as a string capped at `MAX_SESSION_INSTRUCTIONS_LENGTH`.
- **web**: textarea in the Slack Defaults section with `maxLength` bound to the shared constant. Saves go through the existing `mergedGlobalDefaults` helper so routing rules are preserved; an empty value omits the key; reset clears it (dialog copy updated). Saves now also seed the SWR cache with the saved blob — previously, saving Defaults and then Routing Rules before revalidation completed would merge against the stale snapshot and silently drop the just-saved fields (this race predates this PR; the new field made it more visible).
- **slack-bot**: new `getSlackSessionConfig` fetcher returns the workspace default model *and* session instructions from a single control-plane roundtrip (replacing the launcher's `getSlackDefaultModel` call rather than adding a second fetch of the same endpoint; App Home callers are unchanged). Best effort — fails open to an empty config so a control-plane hiccup never blocks session creation. Follow-up messages in an existing thread are not re-decorated, matching Linear's first-prompt-only injection.
- **docs**: new "Session instructions" section in `docs/integrations/SLACK.md`.

## Tests

- control-plane: round-trip, non-string rejection, over-cap rejection, per-repo rejection
- web: save merges instructions with existing defaults/routing rules; clearing omits the key; SWR resync populates the textarea; cache seeding on save; maxLength bound
- slack-bot: `getSlackSessionConfig` (model+instructions from one fetch, invalid model, unset/whitespace/500/throw) and launcher prompt-append
- Suites on this branch: slack-bot 386, control-plane unit 2142, web slack-settings 31 — all passing; full lint/typecheck clean

## Verified in production

Deployed on our instance: instructions saved in the web UI are appended to the first prompt of new Slack sessions (mentions and DMs), and thread follow-ups are unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added workspace-wide session instructions for Slack integrations.
  * Session instructions are appended to the first prompt of new Slack sessions.
  * Added a 10,000-character limit and validation for session instructions.
  * Added Slack settings controls to create, edit, clear, and reset session instructions.
  * Improved settings updates to preserve changes immediately and prevent stale values.

* **Documentation**
  * Documented Slack session instructions, including their scope and behavior for new sessions and thread follow-ups.

### DIFF
--- a/docs/integrations/SLACK.md
+++ b/docs/integrations/SLACK.md
@@ -165,6 +165,16 @@ How matching works:
 Routing rules do not override an active thread: a keyword in a thread reply does not move that
 conversation to a different repository.
 
+### Session instructions
+
+Administrators can define workspace-wide instructions for Slack-started sessions under **Settings →
+Integrations → Slack → Session Instructions** in the web app. When set, the instructions are
+appended to the first prompt of every new Slack-initiated session as an `## Additional Instructions`
+section — use them for standing guidance such as coding standards, preferred tools, or PR
+conventions. They apply to new sessions only (thread follow-ups continue with the session's existing
+context), are limited to 10,000 characters, and mirror the Linear integration's **Issue Session
+Instructions**.
+
 ---
 
 ## Threaded Conversations
--- a/packages/control-plane/src/db/integration-settings.test.ts
+++ b/packages/control-plane/src/db/integration-settings.test.ts
@@ -1059,6 +1059,33 @@ describe("IntegrationSettingsStore", () => {
       ).rejects.toThrow(IntegrationSettingsValidationError);
     });
 
+    it("round-trips global session instructions", async () => {
+      await store.setGlobal("slack", {
+        defaults: { sessionInstructions: "Always run tests before pushing changes." },
+      });
+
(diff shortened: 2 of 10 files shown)


ColeMurray:
thanks!

ColeMurray (approved):

PR #1217, opened 2026-07-31
feat(slack): let channel-trigger automations decline to reply

duboff:
## Summary

A Slack channel-trigger automation fires on ambient messages, so a run's honest outcome is often that nothing was needed — the message turned out to be chatter between people, or a follow-up addressed to someone else.

Completion delivery had no way to express that. `processSlackCompletion` always posts the agent's final text, falling back to `_Agent completed._` when there is none, so an agent instructed to stay quiet posts its *reasoning about staying quiet* instead. In practice a triage automation on a busy channel becomes the noise it was meant to triage: every reply in a watched thread steers the session (by design — conditions gate new runs, not replies), and every steer produces another post.

This adds a declined-reply path: for automation-sourced completions, a final message of `NO_REPLY` (or an empty one) posts nothing and only clears the 👀 reaction.

### Identifying an automation turn

A thread follow-up completes through `/callbacks/complete`, the same route as an interactive `@mention` turn, which hardcodes `source: "session"`. Gating on the route alone would let only an automation's *first* run decline while every follow-up kept posting — and follow-ups are the bulk of the noise.

Only the control plane knows which automation owns a thread, so `steerSession` now stamps `automationId` onto the Slack callback context and the route derives the job source from it. The field is optional and additive; an interactive session never carries one.

### Scoping, so nothing silently disappears

- **Interactive sessions never decline.** An `@mention` has a person waiting on a visible answer, so the `_Agent completed._` fallback stays exactly as it is.
- **Failed runs never decline** — the operator still sees the error.
- **Runs that produced artifacts never decline.** Work that landed outside Slack needs a pointer to it regardless of what the agent wrote.

Opt-in by nature: an agent only emits the sentinel if the automation's instructions tell it to, so existing automations are unaffected.

## Test plan

- [x] `npm test -w @open-inspect/slack-bot` (397), `npm test -w @open-inspect/control-plane` (2197), `npm test -w @open-inspect/shared` (525)
- [x] `npm run typecheck`, `npm run lint:fix`
- [x] New unit tests for `shouldDeclineReply`: sentinel casing/trailing period, empty text, sentinel embedded in a real answer (rejected), interactive source, failed run, artifacts and media artifacts
- [x] New delivery tests: an automation that declines posts nothing but still clears the reaction; a session emitting the same text posts the ordinary completion
- [x] New callback-route test: a `/callbacks/complete` payload carrying `automationId` queues an automation-sourced job
- [x] Extended the existing steer test to assert the context marker

## Summary by CodeRabbit

* **New Features**
  * Automations can suppress Slack replies by returning `NO_REPLY` or an empty successful response.
  * Processing reactions are cleared when a reply is suppressed.
  * Runs that produce artifacts, fail, or occur in interactive sessions continue to post responses.
  * Slack thread follow-ups now preserve automation context for consistent handling.

* **Documentation**
  * Updated Slack automation documentation to describe reply suppression and exceptions.

### DIFF
--- a/docs/AUTOMATIONS.md
+++ b/docs/AUTOMATIONS.md
@@ -258,6 +258,14 @@ finishes, the agent's final response is posted as a reply in that message's thre
 any pull requests it opened and to the full web session — and the reaction is cleared. A failed run
 posts a short failure notice in the thread instead.
 
+A run can also **decline to reply**: if the agent's entire final message is `NO_REPLY` (or empty),
+nothing is posted and only the reaction is cleared. Because every reply in a watched thread wakes
+the automation (see below), an automation that watches a busy channel will otherwise answer messages
+that need nothing from it. Mention the sentinel in the automation's instructions — for example _"if
+the message needs nothing from you, reply with exactly `NO_REPLY` and nothing else"_ — since the
+agent will answer everything it is woken for unless told otherwise. A run that opened a pull request
+or produced other artifacts always posts.
+
 Every reply in a thread **continues the same session** — during the run and after it finishes — for
 up to 7 days after the thread's first trigger, exactly like replying in an `@mention` thread. The
 reply is enqueued as a follow-up turn on that session (re-spawning it from a snapshot if it had gone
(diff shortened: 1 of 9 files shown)


ColeMurray:

PR #1218, opened 2026-07-31
feat(slack): give channel triggers the thread they fired in

duboff:
## Summary

A channel-trigger automation receives the triggering message and nothing else — channel label, actor id, text. When that message is a reply, the agent gets an isolated sentence whose subject lives in the thread parent.

Real example: someone posted a request with a CSV attached, a colleague replied "could we do it via the admin tool?", and that reply was the automation's entire input. The agent's first act was to ask what "it" referred to.

**Thread history is fetched lazily and rendered as untrusted data.**

### Ownership — who reads, and when

`slack-bot` exposes a signed `/internal/thread-context` endpoint, so `SLACK_BOT_TOKEN` and display-name resolution stay in the bot and `normalizeSlackEvent` stays token-free.

The scheduler asks for context only after an invocation is **atomically admitted**. Condition matching alone is not sufficient — `startInvocation` can still return a concurrency skip or a dedup — so `startInvocation` gained a lazy `instructionsOverrideFactory` resolved just after `insertInvocationGuarded` confirms admission. `handleEvent` memoizes one request per event, so several automations matching the same reply share a single Slack read.

No read happens for top-level messages, unmatched events, successful steers, concurrency skips or deduplicated firings. Any failure launches with the plain context block.

### Rendering — untrusted by construction

Messages are serialized as JSON records rather than interpolated into a line-oriented layout. Slack text can contain newlines and literal tags, so `speaker: text` lines let any participant forge a speaker turn or close the block and open another. `JSON.stringify` escapes newlines and quotes, and every left angle bracket is then replaced with its unicode escape, so no delimiter can appear in the payload at all. The block is labelled untrusted data rather than instructions.

### Shared policy instead of a second one

`selectThreadWindow` and `classifyThreadSpeaker` are pure helpers in `@open-inspect/shared/slack`, used by this path **and** by `message-handler.fetchThreadHistory`, which had drifted: two limits, two bot-classification orders, two timestamp rules. Slack access and name resolution stay in slack-bot; each caller formats its own prompt.

Two bugs that drift had already produced are fixed in the shared helper:

- a reply landing between the trigger and the fetch was presented as *prior* context (now a numeric `beforeTs` bound);
- an app message carrying both `bot_id` and `user` rendered as a person (now `bot_id` first).

The window also pins the thread's opening message, which a plain tail would drop — on a long thread that is usually the request itself.

### Also

`file_share` messages no longer bypass triggers, so a request posted with an attachment is not invisible while its follow-ups start runs. Its text matches like any other message; the attachment is still not forwarded. The DM path already made this exception.

Note for reviewers: `conversations.replies` returns oldest-first with no reverse option, and `latest` is a time bound rather than a tail selector — Slack cannot be asked for the last N replies, which is why the fix is "read only when needed" rather than "read less".

## Test plan

- [x] `npm test -w @open-inspect/shared` (538), `-w @open-inspect/slack-bot` (413), `-w @open-inspect/control-plane` (2208), `-w @open-inspect/web` (852)
- [x] Injection regression tests: forged speaker line, forged `</thread_context>` + `<user_content>`, escaping round-trip
- [x] Endpoint tests: signature rejection, malformed payload, no read for a top-level message, display names, `bot_id` precedence, ordering, exclusion of the trigger and anything newer, count and length caps with the root pinned, empty context on Slack failure and on throw
- [x] Scheduler ownership tests: no context for unmatched / steered / concurrency-skipped / deduplicated / top-level; one request per admitted event; reuse across several matching automations; launch without history when the request fails
- [x] Shared helper tests: numeric ts comparison, exclusions applied before the limit, root pinning, speaker classification

## Summary by CodeRabbit

* **New Features**
  * Slack automations now trigger from messages containing file attachments when text is present; image-only messages remain excluded.
  * Reply prompts can include the opening message and up to 20 recent thread messages.
  * Thread context includes speaker labels, bot identification, and safeguards for untrusted content.
* **Bug Fixes**
  * Thread retrieval failures no longer prevent eligible automations from running.
  * Duplicate and ineligible runs no longer fetch unnecessary thread history.
  * Attachments on replies and forwarded-message bodies remain excluded from prompts.
* **Documentation**
  * Updated Slack automation documentation with attachment matching and thread-handling details.

### DIFF
--- a/packages/control-plane/src/scheduler/durable-object.test.ts
+++ b/packages/control-plane/src/scheduler/durable-object.test.ts
@@ -2261,6 +2261,159 @@ describe("SchedulerDO", () => {
       expect(mockStore.insertInvocationGuarded).not.toHaveBeenCalled();
     });
 
+    describe("lazy thread context", () => {
+      /** A slack-bot binding that records thread-context calls. */
+      function threadContextEnv(threadContext = "<thread_context>[]</thread_context>") {
+        const slackFetch = vi.fn(async () => Response.json({ threadContext }));
+        return {
+          slackFetch,
+          env: createEnv({
+            SLACK_BOT: { fetch: slackFetch } as unknown as Fetcher,
+            SERVICE_AUTH_SECRET_SLACK_BOT: "test-secret",
+          } as Partial<Env>),
+        };
+      }
+
+      function threadContextCalls(slackFetch: ReturnType<typeof vi.fn>) {
+        return slackFetch.mock.calls.filter((call) =>
+          String(call[0]).includes("/internal/thread-context")
+        );
+      }
+
+      it("does not request context for an unmatched reply", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([sampleSlackAutomation]);
+        mockStore.getLatestSteerableRunForThread.mockResolvedValue(null);
+        const { slackFetch, env } = threadContextEnv();
+
+        // Fails the automation's text condition, so no run is admitted.
+        await createSchedulerDO(env).fetch(slackEventRequest({ text: "unrelated chatter" }));
+
+        expect(threadContextCalls(slackFetch)).toHaveLength(0);
+      });
+
+      it("does not request context for a successfully steered reply", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([sampleSlackAutomation]);
+        mockStore.getLatestSteerableRunForThread.mockResolvedValue(
+          sampleRunRow({ id: "active-run", session_id: "sess-running" })
+        );
+        const { slackFetch, env } = threadContextEnv();
+
+        await createSchedulerDO(env).fetch(
+          slackEventRequest({ text: "also update the changelog" })
+        );
+
+        expect(threadContextCalls(slackFetch)).toHaveLength(0);
+      });
+
+      it("does not request context when admission is skipped for concurrency", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([sampleSlackAutomation]);
+        mockStore.getLatestSteerableRunForThread.mockResolvedValue(null);
+        mockStore.getActiveRunForKey.mockResolvedValue(sampleRunRow({ id: "busy" }));
+        const { slackFetch, env } = threadContextEnv();
+
+        await createSchedulerDO(env).fetch(slackEventRequest());
+
+        expect(threadContextCalls(slackFetch)).toHaveLength(0);
+      });
+
+      it("does not request context when the invocation is deduplicated", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([sampleSlackAutomation]);
+        mockStore.getLatestSteerableRunForThread.mockResolvedValue(null);
+        mockStore.insertInvocationGuarded.mockRejectedValue(
+          new Error("UNIQUE constraint failed: automation_invocations.trigger_key")
+        );
+        const { slackFetch, env } = threadContextEnv();
+
+        await createSchedulerDO(env).fetch(slackEventRequest());
+
+        expect(threadContextCalls(slackFetch)).toHaveLength(0);
+      });
+
+      it("requests context once for an admitted run and splices it into the prompt", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([sampleSlackAutomation]);
+        mockStore.getLatestSteerableRunForThread.mockResolvedValue(null);
+        const { slackFetch, env } = threadContextEnv();
+        const stub = env.SESSION.get(env.SESSION.idFromName("any"));
+
+        await createSchedulerDO(env).fetch(slackEventRequest());
+
+        expect(threadContextCalls(slackFetch)).toHaveLength(1);
+        const prompt = await getPromptBody(vi.mocked(stub.fetch));
+        const content = String(prompt.content);
+        // Rebuilt block: history sits ahead of the triggering message. Assert both
+        // markers exist first — indexOf returns -1 when absent, and -1 < n passes.
+        const threadIndex = content.indexOf("<thread_context>");
+        const userIndex = content.indexOf("<user_content>");
+        expect(threadIndex).toBeGreaterThanOrEqual(0);
+        expect(userIndex).toBeGreaterThanOrEqual(0);
+        expect(threadIndex).toBeLessThan(userIndex);
+        expect(content).toContain("please deploy the api");
+      });
+
+      it("reuses one context result across several matching automations", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([
+          sampleSlackAutomation,
+          { ...sampleSlackAutomation, id: "auto-slack-2" },
+        ]);
+        mockStore.getLatestSteerableRunForThread.mockResolvedValue(null);
+        const { slackFetch, env } = threadContextEnv();
+
+        await createSchedulerDO(env).fetch(slackEventRequest());
+
+        expect(mockStore.insertInvocationGuarded).toHaveBeenCalledTimes(2);
+        // Two admitted runs, one Slack read.
+        expect(threadContextCalls(slackFetch)).toHaveLength(1);
+      });
+
+      it("launches without history when the context request fails", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([sampleSlackAutomation]);
+        mockStore.getLatestSteerableRunForThread.mockResolvedValue(null);
+        const slackFetch = vi.fn(async () => new Response("nope", { status: 500 }));
+        const env = createEnv({
+          SLACK_BOT: { fetch: slackFetch } as unknown as Fetcher,
+          SERVICE_AUTH_SECRET_SLACK_BOT: "test-secret",
+        } as Partial<Env>);
+        const stub = env.SESSION.get(env.SESSION.idFromName("any"));
+
+        await createSchedulerDO(env).fetch(slackEventRequest());
+
+        const prompt = await getPromptBody(vi.mocked(stub.fetch));
+        expect(String(prompt.content)).toContain("A message was posted in #ops.");
+        expect(String(prompt.content)).not.toContain("<thread_context>");
+      });
+
+      it("launches without history when the context request is aborted", async () => {
+        mockGetSlackAutomationsForChannel.mockResolvedValue([sampleSlackAutomation]);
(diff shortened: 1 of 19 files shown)


packages/shared/src/triggers/slack/normalizer.ts
```diff
@@ -16,27 +27,77 @@ export interface SlackMessageInput {
...
+    const speaker =
+      message.user === botUserId
+        ? "you (this assistant)"
+        : message.user
+          ? `user ${message.user}`
+          : message.bot_id
+            ? `app ${message.bot_id}`
+            : "unknown";
+    const text = message.text.trim().slice(0, SLACK_THREAD_CONTEXT_MAX_MESSAGE_LENGTH);
+    return `${speaker}: ${text}`;
```
ColeMurray: [Automated Review] Message text is inserted directly into a line-based, tag-delimited prompt. Slack messages can contain newlines and literal tags, so a participant can forge a line such as "you (this assistant): ..." or close <thread_context> and open <user_content>. Please serialize typed speaker records with a collision-safe format such as JSON, mark the block as untrusted, and add regression tests for multiline text and delimiter injection. Escaping only the closing tag would not prevent forged speaker lines.

packages/slack-bot/src/channel-trigger.ts
```diff
@@ -97,19 +102,30 @@ export async function handleChannelTrigger(
...
+async function fetchChannelMeta(
+  env: Env,
+  channel: string,
+  ts: string,
+  threadTs?: string
+): Promise<SlackChannelMeta> {
+  const [info, link, thread] = await Promise.all([
     getChannelInfo(env.SLACK_BOT_TOKEN, channel),
     getPermalink(env.SLACK_BOT_TOKEN, channel, ts),
+    threadTs ? getThreadMessages(env.SLACK_BOT_TOKEN, channel, threadTs) : undefined,
```
ColeMurray: [Automated Review] This fetches the full thread before the control plane knows whether history is needed. Existing-session replies use only the new message, and unmatched replies use no history, so most of these reads are discarded and long threads are repeatedly re-read. Please defer the history request until routing determines that at least one new session needs initial context, then fetch it once and reuse it across matching automations. Limiting the result to 20 messages does not eliminate the unnecessary requests.

packages/shared/src/triggers/slack/normalizer.ts
```diff
@@ -16,27 +27,77 @@ export interface SlackMessageInput {
...
+ * agent can read a reply in the conversation it belongs to. Without this a
+ * trigger arrives as an isolated sentence ("could we do it via the admin
+ * tool?") whose subject lives in the thread parent, and the agent cannot tell
+ * whether the message was even meant for it.
+ *
+ * Each line is `<speaker>: <text>`. Speakers are raw Slack ids — the bot's own
+ * messages are marked so the agent recognizes its earlier turns, and other apps
+ * are marked as bots so their output is not mistaken for a person asking.
+ */
+function buildThreadContext(
```
ColeMurray: [Automated Review] This creates a second thread-history policy alongside message-handler.fetchThreadHistory. The two paths implement timestamp filtering, bot classification, limits, and formatting differently; the future-message and bot-id bugs are already examples of that drift. Please extract shared pure helpers for selecting the message window and classifying speakers. Keep Slack API access and name resolution in slack-bot, and let each caller format its own prompt.

packages/shared/src/triggers/slack/normalizer.ts
duboff: Fixed. Messages are no longer interpolated into a line-oriented layout — they are serialized as JSON records (`{speaker, text}`) inside the block, and every left angle bracket in the payload is replaced with its unicode escape, so no delimiter can appear in attacker-controlled text at all. `JSON.stringify` handles newlines and quotes. The block is explicitly labelled untrusted data rather than instructions.

Regression tests in `packages/slack-bot/src/thread-context.test.ts` cover a forged `you (this assistant):` speaker line, a forged `</thread_context>` + `<user_content>` sequence, and a round-trip check that escaping preserves the original text.

packages/slack-bot/src/channel-trigger.ts
duboff: Fixed, and thanks — this was the right boundary. The fetch has moved out of the ingress path entirely.

`slack-bot` now exposes a signed `/internal/thread-context` endpoint, so `SLACK_BOT_TOKEN` and display-name resolution stay in the bot and the normalizer stays token-free. `startInvocation` gained a lazy `instructionsOverrideFactory` resolved only after `insertInvocationGuarded` reports the invocation admitted — not merely after condition matching, since a skip or dedup can still follow. `handleEvent` memoizes one request per event, so several automations matching the same reply share a single read.

No context is requested for top-level messages, unmatched events, successful steers, concurrency skips or deduplicated firings, and any failure launches with the plain context block. Eight cases in `durable-object.test.ts` cover exactly those.

packages/shared/src/triggers/slack/normalizer.ts
duboff: Fixed. `selectThreadWindow` and `classifyThreadSpeaker` now live in `packages/shared/src/slack/thread-context.ts` as pure helpers, and `message-handler.fetchThreadHistory` uses them too. Slack API access and name resolution stay in slack-bot, and each caller still formats its own prompt — `[Name]: text` for the interactive path, JSON records for the trigger path.

Both bugs you pointed at came from that drift and are fixed in the shared helper: `bot_id` is now checked before `user`, and the window filters on a numeric `beforeTs` so a reply landing between the trigger and the fetch cannot be presented as prior context. The helper also pins the thread root, which a plain tail could drop — on a long thread that is usually the request itself.

packages/shared/src/triggers/slack/normalizer.ts
```diff
@@ -16,27 +27,77 @@ export interface SlackMessageInput {
...
+  messages: SlackThreadMessage[],
+  triggeringTs: string,
+  botUserId: string
+): string[] {
+  const earlier = messages
+    .filter((message) => message.ts !== triggeringTs)
+    .slice(-SLACK_THREAD_CONTEXT_MAX_MESSAGES);
```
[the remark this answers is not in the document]
duboff: Fixed — `selectThreadWindow` now filters on a numeric `beforeTs` (the triggering message's ts) rather than only excluding the exact message, so a reply that lands between the trigger and the fetch cannot appear as prior context. Comparison is numeric, since Slack `ts` values compare wrongly as strings.

packages/shared/src/triggers/slack/normalizer.ts
```diff
@@ -16,27 +27,77 @@ export interface SlackMessageInput {
...
+
+  const rendered = earlier.map((message) => {
+    const speaker =
+      message.user === botUserId
+        ? "you (this assistant)"
+        : message.user
+          ? `user ${message.user}`
+          : message.bot_id
+            ? `app ${message.bot_id}`
+            : "unknown";
```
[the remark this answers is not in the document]
duboff: Fixed — `classifyThreadSpeaker` checks `bot_id` before `user`, after the self-bot check, so an app message carrying both is labelled as an app rather than a person. Covered in `packages/shared/src/slack/thread-context.test.ts`.

packages/slack-bot/src/channel-trigger.ts
```diff
@@ -99,19 +100,30 @@ export async function handleChannelTrigger(
...
   ]);
   return {
     channelName: info.ok ? info.channel.name : undefined,
     permalink: link.ok ? link.permalink : undefined,
+    threadMessages: thread?.ok ? thread.messages : undefined,
```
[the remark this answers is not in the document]
duboff: Addressed by the lazy-context rework rather than by bounding the fetch. As you and @ColeMurray both noted, bounding the page walk would not have helped the cases where the read is discarded entirely — the retrieval now happens only after an invocation is admitted, once per event, and never for unmatched events, steers, skips or duplicates.

Worth recording why the original suggestion was not implementable as written: `conversations.replies` returns oldest-first, has no reverse or descending option, and `latest` is a time bound rather than a tail selector, so Slack cannot be asked for the last N replies directly.

packages/control-plane/src/scheduler/durable-object.ts
```diff
@@ -1177,6 +1210,68 @@ export class SchedulerDO extends DurableObject<Env> {
...
+        channel: event.channelId,
+        threadTs: event.threadTs,
+        ts: event.ts,
+      };
+      const signature = await computeHmacHex(JSON.stringify(body), secret);
+      const response = await binding.fetch("https://internal/internal/thread-context", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ ...body, signature }),
+      });
```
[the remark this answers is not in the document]
duboff: Fixed, and a fair catch — this was the weak point of moving the read onto the admission-to-launch path. The binding fetch now carries `AbortSignal.timeout(SLACK_THREAD_CONTEXT_TIMEOUT_MS)`, defined as a named constant beside the other scheduler constants at 10s to match `CALLBACK_ATTEMPT_TIMEOUT_MS` in callback-delivery.

An abort lands in the existing catch, so the run launches with the plain context block rather than stranding children in `starting`. Added a test that throws a `TimeoutError` from the binding and asserts the prompt still goes out without thread history.

packages/control-plane/src/scheduler/durable-object.test.ts
```diff
@@ -2261,6 +2261,133 @@ describe("SchedulerDO", () => {
...
+        const stub = env.SESSION.get(env.SESSION.idFromName("any"));
+
+        await createSchedulerDO(env).fetch(slackEventRequest());
+
+        expect(threadContextCalls(slackFetch)).toHaveLength(1);
+        const prompt = await getPromptBody(vi.mocked(stub.fetch));
+        const content = String(prompt.content);
+        // Rebuilt block: history sits ahead of the triggering message.
+        expect(content.indexOf("<thread_context>")).toBeLessThan(content.indexOf("<user_content>"));
+        expect(content).toContain("please deploy the api");
```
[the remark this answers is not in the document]
duboff: Fixed — the test now asserts both markers are present (`toBeGreaterThanOrEqual(0)`) before comparing their positions, so a regression that dropped the splice entirely fails instead of passing on `-1`.

ColeMurray:

ColeMurray:

ColeMurray:
superseded by: https://github.com/ColeMurray/background-agents/pull/1326

PR #1277, opened 2026-08-05
fix(linear-bot): honor owner/repo replies to the repository clarification

### LINKED ISSUE #1276: linear-bot: replies to the repository clarification are ignored, elicitation loops forever
## Bug

When the Linear bot cannot classify a repository it emits an elicitation ("Please reply with the repository name (e.g., `owner/repo`)"). Replying with an exact `owner/repo` re-triggers the identical elicitation, forever. The user can never get a session started on an ambiguous issue.

## Root cause

Two independent gaps:

1. **The reply text never reaches target resolution.** The reply arrives as an `AgentSessionEvent` with `action: "prompted"`. Since the elicitation created no session, there is no issue→session mapping, so `handleAgentSessionEvent` routes to `handleNewSession` — which passes `webhook.agentSession.comment` (the *original* trigger comment) into `resolveSessionTarget`. The reply lives in `webhook.agentActivity.content.body` and is never read on this path (`packages/linear-bot/src/webhook-handler.ts`).

2. **No deterministic `owner/repo` match exists.** Even if the reply text were passed through, `resolveSessionTarget` only feeds the comment into the LLM classification prompt; there is no explicit-match step, so classification re-runs with effectively identical inputs and returns the identical uncertain verdict (`packages/linear-bot/src/target-resolution.ts`).

Docs promise the opposite behavior: `docs/integrations/LINEAR.md` — "If Open-Inspect asks for clarification, reply with `owner/repo`. That answer is used on the next resolution attempt."

## Reproduction

1. Deployment with ≥2 repos, an issue whose text doesn't clearly match either repo's description.
2. Delegate/mention the agent → clarification is emitted.
3. Reply `owner/repo` (exact, correct casing or not).
4. Same clarification is emitted again, with reasoning that never mentions the reply.

## Fix

PR incoming:
- `handleNewSession` prefers `agentActivity.content.body` as the trigger comment when `action === "prompted"` (a prompted event with no session mapping *is* the clarification reply).
- `resolveSessionTarget` gains an explicit-match rung before the suggestions API / LLM: a single unambiguous `owner/repo` naming an available repository resolves deterministically (case-insensitive, boundary-guarded, multiple matches still fall through to classification).

rhlsthrm:
Fixes #1276.

## Problem

Replying `owner/repo` to the bot's own repository clarification re-triggers the identical clarification forever — the documented flow ("That answer is used on the next resolution attempt", `docs/integrations/LINEAR.md`) never worked on the no-session path.

## Changes

- **`webhook-handler.ts`** — a `prompted` event that reaches `handleNewSession` (i.e. no issue→session mapping exists) is a clarification reply; its text lives on `agentActivity.content.body`, so it now takes precedence over the session's original trigger comment when building the resolution context.
- **`target-resolution.ts`** — new rung in the resolution ladder, ahead of Linear's suggestions API and LLM classification: `matchExplicitRepo` resolves deterministically when the comment names exactly one available repository. Case-insensitive (repos are stored lowercase), boundary-guarded (`acme/api` does not match inside `acme/api-legacy` or `notacme/api`), and multiple distinct matches still fall through to classification, since that is a genuine ambiguity.

## Testing

- Unit tests for `matchExplicitRepo` (single/multi/none/boundary/case).
- End-to-end handler test: `prompted` event with no session mapping + `acme/backend` reply → session created against `acme/backend`; the control-plane stub throws on any unexpected fetch, so the LLM classifier being consulted would fail the test.
- `npm test -w @open-inspect/linear-bot`: 208/208. Typecheck + eslint clean.

Observed in production: an issue describing UI work didn't match either repo description, and three consecutive exact `owner/repo` replies each re-ran cold classification with identical reasoning.

## Summary by CodeRabbit

* **New Features**
  * Explicit repository names in comments are now recognized with case-insensitive matching.
  * Repository resolution prioritizes clear repository references before heuristic classification.
  * Repository matching preserves the repository’s full name when resolved.

* **Bug Fixes**
  * Improved clarification replies so explicitly specified repositories resolve correctly.
  * Prevented ambiguous, partial, or incorrectly delimited repository-name matches.
  * Ensured prompted activity is attributed to the responding actor when creating a session.


packages/linear-bot/src/target-resolution.ts
```diff
@@ -29,6 +31,29 @@ import { createLogger } from "./logger";
...
+ */
+export function matchExplicitRepo(text: string, repos: RepoConfig[]): RepoConfig | null {
+  const haystack = text.toLowerCase();
+  const boundary = /[\w/-]/;
+  const named = repos.filter((repo) => {
+    const needle = repo.fullName.toLowerCase();
+    for (let at = haystack.indexOf(needle); at !== -1; at = haystack.indexOf(needle, at + 1)) {
+      const before = at === 0 ? "" : haystack[at - 1];
+      const after = haystack[at + needle.length] ?? "";
+      if (!boundary.test(before) && !boundary.test(after)) return true;
```
ColeMurray: [Automated Review] The boundary check treats a period as punctuation, so when only acme/backend is available, text such as acme/backend.docs or not.acme/backend matches acme/backend. Because this rung returns before suggestions and classification, that can create a session against the wrong repository. Please treat a period as a path continuation when it connects to another repository-path character, while still accepting terminal punctuation such as acme/backend., and add those regression cases.

packages/linear-bot/src/webhook-handler.ts
```diff
@@ -450,7 +450,14 @@ async function handleNewSession(
...
   const agentSessionId = webhook.agentSession.id;
-  const comment = webhook.agentSession.comment;
+  // A "prompted" event that reaches new-session handling is a reply to an
+  // elicitation — no issue→session mapping existed, so no session was ever
+  // created. The reply text lives on the agent activity, not on the session's
+  // original trigger comment, and it is the clarification the resolver asked
+  // for — so it must win.
+  const replyBody =
+    webhook.action === "prompted" ? webhook.agentActivity?.content?.body?.trim() : undefined;
+  const comment = replyBody ? { body: replyBody } : webhook.agentSession.comment;
```
ColeMurray: [Automated Review] This selects the prompted reply body, but the new-session actor still comes from agentSession.comment.userId or creatorId through getNewSessionActorUserId. If user B answers an elicitation originally created by user A, the session is signed as linear:A and user A preferences and profile are used even though user B supplied the reply. Please select agentActivity.userId on this prompted path, ideally pairing reply body and actor in one helper, and add a regression test with distinct creator and replier IDs.

packages/linear-bot/src/target-resolution.ts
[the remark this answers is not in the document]
rhlsthrm: Fixed in c0818574: a period now invalidates a match only when it connects to a repository-path character — `acme/backend.docs` and `not.acme/backend` no longer match `acme/backend`, while terminal punctuation (`use acme/backend.`) still does. Regression cases added for all three.

rhlsthrm: Fixed in c0818574: boundary check now treats a period as a path continuation when adjacent to another repository-path character (`not.acme/backend`, `acme/backend.docs` rejected), while terminal punctuation `acme/backend.` still matches. Regression tests added.

packages/linear-bot/src/webhook-handler.ts
rhlsthrm: Fixed in c0818574: the prompted path now pairs reply body and actor in one helper (`getNewSessionInput`) — the session actor is `agentActivity.userId` (the replier), falling back to the session comment/creator only when the activity carries no user. Regression test added with distinct creator (`creator-user-1`) and replier (`replier-user-2`) asserting the signed `X-OpenInspect-Actor` header.

PR #1291, opened 2026-08-05
fix(slack): split long agent notifications instead of truncating them

duboff:
## Summary

An agent posting through the notify route has its message cut at 2900 characters, mid-sentence. The response reports `truncated: true`, so the agent finds out — after the fact, with the text already gone and no way to recover it. The cut lands on the tail, which for any analysis writeup is where the conclusions and recommendations live.

Observed in practice: an agent posted a long investigation into a Slack thread, and the message ended `1. No p…`. Everything after that — the entire "what to do about it" section, the only part the reader needed — was discarded. The agent then had to guess what had been lost and post a follow-up.

**The completion path already solved this.** `splitIntoSlackSections` spreads long text across consecutive section blocks and keeps code fences balanced across the seam, so a long answer arrives whole. The notify route never got the same treatment — the same worker had two different answers to the same Slack limit.

This moves the splitter into `@open-inspect/shared/slack` and uses it in both places:

- `slack-notify` sanitizes against `RAW_TEXT_INPUT_MAX_LENGTH` (the cap it already validates and enforces), splits the result, and emits one section block per chunk. `SLACK_TEXT_MAX_LENGTH` is gone.
- The splitter's ceiling (`MAX_RESPONSE_SECTIONS` × `SECTION_TEXT_MAX_CHARS` = 60k) sits far above that 12k input cap, so `truncated` in the response now reflects only the documented raw-input limit rather than a Slack rendering detail.
- The completion path is unchanged: it imports the same function from shared instead of defining it locally.

## Notes

The splitter's tests deliberately stay in `packages/slack-bot/src/completion/blocks.test.ts`. They exercise it *through* `buildCompletionBlocks`, so leaving them in place also asserts that moving the function left the completion output byte-identical. Moving them to a shared-level suite would have lost that guarantee for no gain.

The fallback text passed to `chat.postMessage` is now the opening section rather than the whole body — a section-split message has no single text body, and Slack only uses that field for the notification preview.

Worth a separate look: the `slack-notify` tool description doesn't mention any length behaviour, so an agent has no way to budget for it. Now that long messages split cleanly that matters less, but stating it would let the model structure long output deliberately rather than discovering the limit from a response flag.

## Test plan

- [x] `npm test -w @open-inspect/shared` (545), `-w @open-inspect/slack-bot` (406), `-w @open-inspect/control-plane` (2312)
- [x] New regression test in `src/routes/slack-notify.test.ts`: a message well over one section keeps every section within 3000 chars, emits more than one section, reports `truncated: false`, and — the point of the test — still contains its closing recommendation. It fails on `main`, where that tail is cut.
- [x] The existing 406 slack-bot tests pass unchanged, confirming the completion path is unaffected by the move.

## Summary by CodeRabbit

* **Bug Fixes**
  * Slack notifications now split long responses into multiple sections without losing the final recommendation.
  * Each section respects Slack’s 3,000-character limit.
  * Responses preserve paragraph, line, Unicode, and code-block formatting where possible.
  * Extremely long responses are capped with a truncation indicator after 20 sections.

* **Tests**
  * Added regression coverage for oversized Slack messages, section limits, and truncation behavior.


duboff:
Taken — the module doc now points at `SECTION_TEXT_MAX_CHARS` instead of restating 3000, and the test imports the constant for both assertions.

Skipped the third site (`sections.ts#L14-L18`): that range is the constant’s own doc comment, whose prose says “longer than this” rather than repeating the number — the only literal there is the declaration itself, which has to stay.

PR #1362, opened 2026-08-10
Fix session timeline scroll containment

ColeMurray:
## Summary

- contain the resizable timeline and desktop workspace panels so their implicit content wrappers cannot become competing scroll containers
- add `min-h-0` through the session flex hierarchy and use `overflow-clip` at the session root
- run bottom-following in a layout effect for both event and processing-state changes while keeping history-prepend compensation in its own earlier layout effect
- add focused regressions for panel overflow containment and processing-indicator height changes

## Why PR #1340 was incomplete

PR #1340 correctly removed streaming-triggered `scrollIntoView` calls and scoped auto-follow writes to `timeline.scrollTop`. That prevented direct page-level scrolling, but it did not eliminate the other scroll containers surrounding the timeline.

`react-resizable-panels` v4 wraps each `Panel` child in an implicit element whose default `overflow: auto` can make it independently scrollable. When nested sub-task activity grows beneath an earlier Task row, browser scroll anchoring can adjust those panel wrappers even though the timeline code only writes to the intended timeline element. This is most visible while a sub-task is thinking because activity is inserted retroactively rather than only appended at the end.

The timeline also performed bottom-following in a passive effect keyed only by `events`. Toggling `isProcessing` adds or removes the Thinking indicator without necessarily changing the event array, so its height change was not corrected before paint. The follow effect now uses `useLayoutEffect` and depends on both `events` and `isProcessing`.

History prepend compensation remains a separate layout effect that runs first. This preserves the reader's viewport and marks that render so the following bottom-follow effect does not race the prepend adjustment when processing state changes at the same time.

## Testing

- `npm test -w @open-inspect/web -- src/components/session-timeline.test.tsx src/components/session-desktop-layout.test.tsx` (29 passed)
- `npm run lint -w @open-inspect/web`
- `git diff --check`

The unrelated pre-existing `package-lock.json` modification is not included.

---
*Created with [Open-Inspect](https://open-inspect-prod.vercel.app/session/c0a07b8b0a40607f2b209aafade7bd0a)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session workspace sizing and overflow handling to keep content contained within its panels.
  * Improved timeline scrolling so new activity and processing updates remain visible without disrupting the user’s current position.

* **Tests**
  * Added coverage for workspace overflow and nested panel behavior.
  * Added coverage for timeline auto-scrolling, processing updates, and preserving the user’s scroll position.

### DIFF
--- a/packages/web/src/app/(app)/session/[id]/page.tsx
+++ b/packages/web/src/app/(app)/session/[id]/page.tsx
@@ -259,9 +259,13 @@ export default function SessionPage() {
   }, [focusDetailsTrigger, isBelowLg]);
 
   const sessionWorkspace = (
-    <div className="flex h-full flex-1 flex-col overflow-hidden">
+    <div className="flex h-full min-h-0 flex-1 flex-col overflow-hidden">
       <PanelGroup orientation="vertical" id="session-terminal">
-        <Panel defaultSize={showTerminal ? "70%" : "100%"} minSize="30%">
+        <Panel
+          defaultSize={showTerminal ? "70%" : "100%"}
+          minSize="30%"
+          style={{ minHeight: 0, overflow: "hidden" }}
+        >
           <SessionTimeline
             events={events}
             sessionId={sessionId}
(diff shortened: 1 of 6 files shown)


ColeMurray:
@open-inspect[bot] please re-review and approve

PR #1405, opened 2026-08-13
feat(sandbox): add E2B prebuilt-image (snapshot) support

tomasvarga:
Adds prebuilt-image (snapshot) support to the E2B sandbox provider, so E2B sessions can boot from an image built once by `.openinspect/setup.sh` instead of re-running setup on every spawn — the same capability Modal, Vercel and OpenComputer already have.

Follows on from #1008, which added the E2B provider itself.

## How it works

An E2B snapshot template id doubles as a `templateID`, so a prebuilt spawn is just a create with that id in place of the base template. The image-build workflow runs setup once in a build sandbox, then bakes its filesystem into a reusable snapshot.

Baking needs care, because a reusable E2B snapshot captures live process memory. Snapshotting the running build sandbox directly would bake the build supervisor and its secret env into every image, and resume that stale process on spawn instead of a fresh launcher. So `takePrebuiltImageSnapshot` first does `pause(memory: false)` — which drops all memory and persists only the filesystem — then `connect`, which cold-boots from disk and re-runs the launcher into its env-wait state. The snapshot captures that clean state, so sandboxes spawned from it start a fresh supervisor with their own per-session env.

Sessions restored from a *session* snapshot get the inverse treatment: that snapshot deliberately preserves process memory, so the captured supervisor would wake holding the previous session's credentials. `restoreFromSnapshot` creates the sandbox with outbound networking off, drops the captured memory, cold-boots, and only then re-enables the network and delivers fresh session env.

The generic `takeSnapshot` used by the lifecycle manager after a prompt is left alone — it snapshots a live session in place, without pausing or reconnecting it.

## Integration

E2B registers as an image-build provider (`provider-policy`, `provider-factory`, `E2BImageBuildAdapter`) using the shared provider-session lifecycle and `buildImageBuildEnvVars` / `buildImageBuildCallbackEnv` helpers, so no E2B-specific plumbing enters the workflow. The web layer adds `e2b` to the prebuilt-image capability checks.

`E2B_SANDBOX_VERSION` and the template's `SANDBOX_VERSION` are bumped to `v57-vnc-opencode-1-18-11`, matching the Vercel/OpenComputer constants and the image's pinned opencode. This is required, not cosmetic: the previous value parsed below `MIN_COMPATIBLE_RUNTIME_VERSION`, which would have made spawn-time selection reject every E2B prebuilt image. **The E2B template needs a rebuild when this deploys.**

## Testing

Unit tests cover the adapter, the REST client, the provider paths and the lifecycle interaction. Beyond those, the template was rebuilt and the changed API surface was exercised against live E2B:

- `PUT /sandboxes/{id}/network` behaves as used by the restore path
- `DELETE /templates/{id}` accepts a tagged snapshot id (`abc:default`) verbatim — E2B strips the tag server-side
- `pause(memory: false)` → `connect` cold-boots and returns the envd access token the post-boot env write needs
- a snapshot id spawns as a `templateID`; `takeSnapshot` leaves a live session `running`; `restoreFromSnapshot` completes its quarantine sequence and boots

Not covered by automation: the bridge ↔ control-plane WebSocket path, which needs a running control plane.

## Summary by CodeRabbit

* **New Features**
  * Added E2B support for sandbox image builds, prebuilt images, snapshots, and snapshot restoration.
  * Added provider image creation and cleanup for completed or failed builds.
  * Added configurable internet access and network updates for E2B sandboxes.
  * Preserved refreshed access credentials when reconnecting to sandboxes.
* **Bug Fixes**
  * Unsupported snapshot and restore operations now safely fall back or skip when unavailable.
  * Improved cleanup and handling of missing resources during sandbox and image operations.
* **Infrastructure**
  * Updated the E2B runtime image and startup behavior for restored prebuilt environments.

### DIFF
--- a/packages/control-plane/src/image-builds/e2b-adapter.test.ts
+++ b/packages/control-plane/src/image-builds/e2b-adapter.test.ts
@@ -1,0 +1,131 @@
+import { describe, expect, it, vi } from "vitest";
+import type { E2BSandboxProvider } from "../sandbox/providers/e2b-provider";
+import { E2BImageBuildAdapter } from "./e2b-adapter";
+import { resolveImageBuildProviderSessionTimeoutSeconds } from "./timeouts";
+import type { ImageBuildPlan } from "./types";
+
+function createProvider(): E2BSandboxProvider {
+  return {
+    triggerImageBuild: vi.fn(async () => undefined),
+    takeSnapshot: vi.fn(async () => ({ success: true, imageId: "snap-abc:default" })),
+    deleteSandbox: vi.fn(async () => undefined),
+    deleteProviderImage: vi.fn(async () => undefined),
+  } as unknown as E2BSandboxProvider;
+}
+
+function createPlan(): ImageBuildPlan {
+  return {
+    buildId: "build-1",
+    scope: { kind: "repo", id: "acme/repo" },
+    repositories: [{ repoOwner: "acme", repoName: "repo", baseBranch: "develop" }],
+    repositoriesFingerprint: "fp-1",
+    callbackUrl: "https://worker.test/image-builds/build-complete",
+    failureCallbackUrl: "https://worker.test/image-builds/build-failed",
+    callbackToken: "callback-token",
+    cloneAuth: { type: "credential_helper", token: "clone-token" },
+    buildTimeoutMs: 1_800_001,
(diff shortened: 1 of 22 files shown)


ColeMurray:
superseded by https://github.com/ColeMurray/background-agents/pull/1535

Issue #1374, opened 2026-08-12
OpenAI Slack target classification cannot work: chatgpt.com blocks the CF-Worker header


ColeMurray:
## Summary

Slack target classification with `slack_classification_model = "openai/gpt-5.6-luna"` cannot work in **any** Open-Inspect deployment.

The control plane calls the Codex subscription endpoint (`https://chatgpt.com/backend-api/codex/responses`) directly from a Cloudflare Worker. The Workers runtime stamps a `CF-Worker` header on every outbound subrequest, and `chatgpt.com` returns **403 with an HTML bot-block page** to any request carrying it. The header is added by the runtime after user code runs and cannot be removed.

Because the control plane is always a Cloudflare Worker, this affects every deployment, not a particular configuration.

## Symptom

Slack falls back to the target picker on every message that reaches the LLM tier. Logs:

```
slack-bot     classifier.classify        Target classification request failed with 502
control-plane classifier.upstream_error  upstream_status: 403
```

## Root cause

One Node process, one credential, two back-to-back requests, single variable:

```
node, no CF-Worker         → 200  api-response
node + CF-Worker header    → 403  HTML block page
```

Reproduce by adding `"CF-Worker": "anything.workers.dev"` to an otherwise-working Codex request.

To see what the runtime emits, POST from a Worker to `https://httpbin.org/anything` and read back the echoed headers — `Cf-Worker: <script>.<zone>` appears alongside yours. Four suppression attempts (omit / set `""` / spoof another value / `Headers.delete()` then re-set) all still arrived stamped with the real value.

## What this is not

| Theory | Ruled out by |
| --- | --- |
| Stale or wrong token | Token had ~4 days of validity; it is the same global credential working sessions use |
| Missing/incorrect `User-Agent` | Five UA values from a Worker (none, `opencode/...`, `codex_cli_rs/...`, Chrome, `node`) — all 403 |
| `session-id` vs `session_id` | codex-rs (`codex-api/src/requests/headers.rs`) and opencode (`plugin/openai/codex.ts`) both use the hyphen form, matching the control plane |
| Responses-Lite vs normal API | Blocked identically; the block precedes request parsing |
| Cloudflare **egress / IP reputation** | `wrangler dev` **without** `--remote` runs workerd locally and is still blocked, while plain Node on the same machine succeeds. Same IP, different runtime. |

That last row is the trap — an edge-only test looks like an egress problem. It isn't.

## Why sandbox sessions are unaffected

[...]

PR #1408, opened 2026-08-13
feat(bots): let the Slack and Linear classifiers run on OpenAI, and bound the request

rhlsthrm:
## Problem

The Slack and Linear bots classify each inbound message to decide which repository or environment a coding session should target. Both classifiers are pinned to Anthropic:

- `packages/slack-bot/src/classifier/index.ts` builds an Anthropic client and forces a `classify_target` tool call.
- `packages/linear-bot/src/classifier/index.ts` calls `api.anthropic.com/v1/messages` directly and **hardcodes** `claude-haiku-4-5` with no env override at all.

Two consequences:

1. **Single-provider coupling.** An Anthropic outage, rate limit, or billing lapse degrades routing on every deployment, with no way to point the classifier elsewhere — even for deployments whose coding agents already run OpenAI models. We hit exactly this: an Anthropic billing lapse dropped both bots to "pick a target yourself" until it was noticed.
2. **Unbounded requests.** Neither classifier passes an abort signal, so a stalled or queued provider request holds the Slack thread / Linear webhook open until the platform kills the invocation. The classifiers already fail soft to a target picker, so a *fast* failure is cheap — it was the unbounded wait that hurt.

## What this does

Lets an operator pick the classifier's provider, requires **only that provider's** credential, and binds exactly one provider key to the bots.

| `classification_model` | Provider | Credential required |
|---|---|---|
| `anthropic/<x>` or bare `claude-*` (default) | Anthropic, existing tool-calling request | `classification_anthropic_api_key`, falling back to `anthropic_api_key` |
| `openai/<x>` or bare `gpt-*` | OpenAI Chat Completions, strict `json_schema` | `classification_openai_api_key` |

The prefix rule reuses the convention already encoded in `normalizeModelId`/`MODEL_CATALOG` in `packages/shared/src/models.ts`, so there is no second setting that can disagree with the model id. The bare id is sent to the provider. An unrecognised prefix throws into each classifier's existing `catch`, which already degrades to asking the user to pick — no new failure mode.

Both providers funnel through the existing validators (`normalizeModelResponse` in slack-bot, `classifyToolInputSchema` in linear-bot), so the downstream contract is untouched.

`CLASSIFICATION_REQUEST_TIMEOUT_MS = 15_000` now bounds **both** providers, following the existing convention (`REPOS_FETCH_TIMEOUT_MS`, `OUTBOUND_REQUEST_TIMEOUT_MS`): milliseconds in the name, defined once, and asserted in tests by identity of the signal object rather than just its shape.

### Scope of the credential choice — please read

This is deliberately **classifier-scoped**, not a deployment-wide provider switch. `anthropic_api_key` is left exactly as it is on `main` (`nullable = false`, non-blank validation) because it has consumers unrelated to classification: the Modal sandbox's `llm-api-keys` secret (`modal.tf`) that Claude coding sessions use, and the opencomputer control-plane path. The diff to `variables.tf` is purely additive — it does not touch that variable.

So: choosing the OpenAI classifier means you supply `classification_openai_api_key` and the bots receive **only** that key. It does not make the deployment OpenAI-only, and this PR makes no claim to. Making sandbox provider credentials uniformly optional is a separate, larger change tied to the default coding model, and I have not attempted it here.

## Backward compatibility

**Nothing changes for an existing deployment that sets no new value.**

- `classification_model` defaults to `claude-haiku-4-5` — today's value.
- `classification_anthropic_api_key` defaults to blank and falls back to `anthropic_api_key`, so existing deployments keep working untouched.
- The Anthropic request body is unchanged; the timeout is passed as `messages.create(body, { signal })`, so the body itself is untouched.
- `ANTHROPIC_API_KEY` stays required, the `@anthropic-ai/sdk` dependency stays, `CLASSIFY_TARGET_TOOL` stays.
- No Claude entries removed anywhere — `packages/linear-bot/src/model-resolution.ts` (`MODEL_LABEL_MAP`) is untouched, so `model:opus`-style Linear labels keep working.
- Anthropic-classifier deployments keep exactly the bot secret bindings they had; no empty secret is introduced and no worker version churns from this change.
- The Anthropic SDK client is now constructed lazily, so an OpenAI-configured deployment never reaches `new Anthropic({ apiKey: undefined })`.

The Linear bot gains a `CLASSIFICATION_MODEL` binding it never had; its default makes the previously hardcoded `claude-haiku-4-5` explicit, so the effective model is unchanged.

## Configuration

```hcl
# Default — Anthropic, using the key you already supply
# classification_model = "claude-haiku-4-5"

# Or classify on OpenAI; the bots then receive only this key
classification_model          = "gpt-5.4-mini"
classification_openai_api_key = "sk-proj-..."
```

Each provider's key is validated non-blank **when that provider is selected and a classifier bot is enabled** — so an OpenAI deployment is never asked for an Anthropic classifier key, a deployment running neither bot is never asked for either, and a selected provider can't ship credential-less. That last guard matters because GitHub Actions renders an unset secret as an empty string, which would otherwise plan and apply cleanly and leave a classifier rejecting every message. For the same reason the workflow maps the model with an explicit fallback (`${{ vars.CLASSIFICATION_MODEL || 'claude-haiku-4-5' }}`, matching the existing `ENABLE_SLACK_BOT || 'true'` pattern), and the configuration additionally refuses a blank override rather than silently treating it as "use the default".

## Verification

Terraform (`terraform test`, mock providers) — **18 passed, 0 failed**, including a new `tests/classifier_provider.tftest.hcl` whose 8 runs cover every branch:

- Anthropic default binds `ANTHROPIC_API_KEY` and **no** `OPENAI_API_KEY` on both bots (the backward-compatibility guarantee, asserted rather than assumed)
- OpenAI model binds `OPENAI_API_KEY` and **no** `ANTHROPIC_API_KEY` — exactly one provider credential reaches the bots, asserted in both polarities
- `gpt-5.4-mini` and `openai/gpt-5.4-mini` both resolve to OpenAI; `anthropic/claude-haiku-4-5` resolves to Anthropic
- OpenAI model with a blank key → plan **fails**
- OpenAI model with both bots disabled and a blank key → plan **succeeds**
- unknown provider prefix → plan **fails**; blank model → plan **fails**

The pre-existing `anthropic_api_key_blank` guard in `tests/auth_provider_configuration.tftest.hcl` still passes unchanged. `terraform fmt -check -recursive` clean; `terraform validate` success.

TypeScript: `npm run typecheck` exit 0; `eslint --max-warnings 0` clean on both changed packages.

Unit suites (clean upstream-main baseline → this branch): slack-bot 421 → **425**, linear-bot 223 → **230**; unchanged elsewhere: shared **601**, github-bot **130**, control-plane **2518**, web **956**. Control-plane integration (workerd + real D1): **778 passed**.

New tests per bot cover: the OpenAI request contract (`max_completion_tokens` present, `max_tokens` absent, `temperature: 0`, `strict: true`, bare model id, `additionalProperties: false`, all fields `required`, nullable id typed `["string","null"]`), non-2xx degrading to the picker, the timeout signal being the exact `AbortSignal.timeout` object, the Anthropic default path still firing when nothing is set, and an unrecognised prefix degrading without calling either provider.

## Notes for reviewers

- **`max_completion_tokens` is required and `max_tokens` is rejected** by the gpt-5 family (`Unsupported parameter: 'max_tokens' is not supported with this model`) — verified against the live API, and pinned by a test in each bot so it cannot regress silently.
- Each bot implements its own small OpenAI request function rather than sharing one: two call sites with different schemas, and it keeps each Worker self-contained. Happy to extract into `packages/shared` if you would prefer that.
- The provider is derived from the model id rather than a separate `CLASSIFICATION_PROVIDER` variable, to avoid a setting that can disagree with the model. If you would rather support OpenAI-compatible gateways (Azure, OpenRouter, proxies) whose ids are not `gpt-*`, an explicit provider override is the natural follow-up — happy to add it here or later.
- `classification_anthropic_api_key` exists mainly so the two providers are symmetric and the classifier's credential is separable from the sandbox's. If you would rather the Anthropic classifier just always read `anthropic_api_key` and drop that variable, that is a one-line simplification — say which you prefer.
- The `docs/GETTING_STARTED.md` diff looks larger than it is: adding `CLASSIFICATION_ANTHROPIC_API_KEY` widened the Actions-secret table's first column, so Prettier (which your `lint-staged` runs on Markdown) realigned every row. `git diff -w` on that file shows only the six sample lines, the two new table rows, and the widened separator.

## Summary by CodeRabbit

* **New Features**
  * Added configurable classification model selection for Slack and Linear bots.
  * Added OpenAI and Anthropic classification support with provider-specific credentials.
  * Added structured response validation and 15-second request timeouts.
  * Added graceful handling for unsupported models, provider errors, and missing credentials.

* **Documentation**
  * Updated setup and deployment guidance for models and API keys.

* **Tests**
  * Expanded coverage for provider selection, validation, timeouts, credentials, and fallback behavior.

### DIFF
--- a/packages/linear-bot/src/classifier/index.test.ts
+++ b/packages/linear-bot/src/classifier/index.test.ts
@@ -1,40 +1,12 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import type { RepoConfig } from "@open-inspect/shared/types/repository-catalog";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { anthropicMessagesResponseSchema, classifyRepo, classifyToolInputSchema } from "./index";
 import {
-  anthropicMessagesResponseSchema,
-  CLASSIFIER_REQUEST_TIMEOUT_MS,
-  classifyRepo,
-  classifyToolInputSchema,
-} from "./index";
+  CLASSIFICATION_REQUEST_TIMEOUT_MS,
+  OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS,
+} from "@open-inspect/shared/classification";
+import { clearReposLocalCache } from "./repos";
 import { createFakeKV, makeLinearBotEnv } from "../test-helpers";
-
-const { getAvailableRepos, buildRepoDescriptions } = vi.hoisted(() => ({
-  getAvailableRepos: vi.fn(),
-  buildRepoDescriptions: vi.fn(),
-}));
-
-vi.mock("./repos", () => ({ getAvailableRepos, buildRepoDescriptions }));
-
-const repos: RepoConfig[] = ["api", "web"].map((name) => ({
-  id: `acme/${name}`,
-  owner: "acme",
-  name,
-  fullName: `acme/${name}`,
-  displayName: name,
-  description: `${name} repository`,
-  defaultBranch: "main",
-  private: true,
-}));
-
-beforeEach(() => {
-  getAvailableRepos.mockResolvedValue(repos);
-  buildRepoDescriptions.mockResolvedValue("- acme/api\n- acme/web");
-});
-
-afterEach(() => {
-  vi.restoreAllMocks();
-  vi.unstubAllGlobals();
-});
+import type { Env } from "../types";
 
 describe("anthropicMessagesResponseSchema", () => {
   it("parses a response with the consumed tool block fields", () => {
@@ -99,10 +71,168 @@ describe("classifyToolInputSchema", () => {
   });
 });
 
-describe("classifyRepo", () => {
+describe("classifyRepo provider dispatch", () => {
+  const traceId = "trace-classify";
+
+  function twoRepoControlPlane(): Fetcher {
+    return {
+      fetch: vi.fn(async () =>
+        Response.json({
+          repos: [
+            {
+              id: 1,
+              owner: "acme",
+              name: "alpha",
+              fullName: "acme/alpha",
+              description: "Alpha service",
+              private: false,
+              defaultBranch: "main",
+              archived: false,
+              language: "TypeScript",
+              metadata: {},
+            },
+            {
+              id: 2,
+              owner: "acme",
+              name: "beta",
+              fullName: "acme/beta",
+              description: "Beta service",
+              private: false,
+              defaultBranch: "main",
+              archived: false,
+              language: "TypeScript",
+              metadata: {},
+            },
+          ],
+          cached: false,
+          cachedAt: "2026-08-02T00:00:00.000Z",
+        })
+      ),
+    } as unknown as Fetcher;
+  }
+
+  function classify(env: Env) {
+    return classifyRepo(
+      env,
+      "Fix the login bug",
+      "Users cannot log in",
+      ["bug"],
+      "Core",
+      "Platform",
+      "PLAT",
+      undefined,
+      traceId
+    );
+  }
+
+  function anthropicToolResponse(repoId: string) {
+    return vi.fn<typeof fetch>(async () =>
+      Response.json({
+        content: [
+          {
+            type: "tool_use",
+            name: "classify_repository",
+            input: { repoId, confidence: "high", reasoning: "Matches", alternatives: [] },
+          },
+        ],
+      })
+    );
+  }
+
+  beforeEach(() => {
+    clearReposLocalCache();
+    vi.unstubAllGlobals();
+  });
+
+  it("sends a spec-compliant OpenAI request when CLASSIFICATION_MODEL selects an OpenAI model", async () => {
+    const { kv } = createFakeKV();
+    const env = makeLinearBotEnv(kv, {
+      CONTROL_PLANE: twoRepoControlPlane(),
+      CLASSIFICATION_MODEL: "openai/gpt-5.4-mini",
+      OPENAI_API_KEY: "openai-key",
+    });
+
+    const fetchMock = vi.fn<typeof fetch>(async () =>
+      Response.json({
+        choices: [
+          {
+            message: {
+              content: JSON.stringify({
+                repoId: "acme/alpha",
+                confidence: "high",
+                reasoning: "Matches",
+                alternatives: [],
+              }),
+            },
+          },
+        ],
+      })
+    );
+    vi.stubGlobal("fetch", fetchMock);
+
+    const result = await classify(env);
+
+    expect(result.repo?.id).toBe("acme/alpha");
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    const [url, init] = fetchMock.mock.calls[0];
+    expect(url).toBe("https://api.openai.com/v1/chat/completions");
+    expect(init!.headers).toMatchObject({ Authorization: "Bearer openai-key" });
+
+    const body = JSON.parse(init!.body as string);
+    expect(body.model).toBe("gpt-5.4-mini");
+    // gpt-5-family models reject an explicit temperature with HTTP 400.
+    expect(body).not.toHaveProperty("temperature");
+    expect(body.max_completion_tokens).toBe(OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS);
+    expect(body).not.toHaveProperty("max_tokens");
+    expect(body.response_format.type).toBe("json_schema");
+    expect(body.response_format.json_schema.strict).toBe(true);
+    const schema = body.response_format.json_schema.schema;
+    expect(schema.additionalProperties).toBe(false);
+    expect(schema.required).toEqual(["repoId", "confidence", "reasoning", "alternatives"]);
+    expect(schema.properties.repoId.type).toEqual(["string", "null"]);
+  });
+
+  it("degrades to a clarification result with alternatives on a non-2xx OpenAI response", async () => {
+    const { kv } = createFakeKV();
+    const env = makeLinearBotEnv(kv, {
+      CONTROL_PLANE: twoRepoControlPlane(),
+      CLASSIFICATION_MODEL: "gpt-5.4-mini",
+      OPENAI_API_KEY: "openai-key",
+    });
+
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(async () => new Response("server exploded", { status: 500 }))
+    );
+
+    const result = await classify(env);
+
+    expect(result.needsClarification).toBe(true);
+    expect(result.repo).toBeNull();
+    expect(result.alternatives?.length).toBeGreaterThan(0);
+  });
+
+  it("bounds every classification request with the shared timeout signal", async () => {
+    const { kv } = createFakeKV();
+    const env = makeLinearBotEnv(kv, { CONTROL_PLANE: twoRepoControlPlane() });
+
+    const fakeSignal = {} as AbortSignal;
+    const timeoutSpy = vi.spyOn(AbortSignal, "timeout").mockReturnValue(fakeSignal);
+    const fetchMock = anthropicToolResponse("acme/alpha");
+    vi.stubGlobal("fetch", fetchMock);
+
+    await classify(env);
+
+    expect(timeoutSpy).toHaveBeenCalledWith(CLASSIFICATION_REQUEST_TIMEOUT_MS);
+    const [, init] = fetchMock.mock.calls[0];
+    expect(init!.signal).toBe(fakeSignal);
+
+    timeoutSpy.mockRestore();
+  });
+
   it("falls back to clarification when the classifier request times out", async () => {
     const timeoutSignal = AbortSignal.abort(new DOMException("timed out", "TimeoutError"));
-    const timeoutSpy = vi.spyOn(AbortSignal, "timeout").mockReturnValue(timeoutSignal);
+    vi.spyOn(AbortSignal, "timeout").mockReturnValue(timeoutSignal);
     vi.stubGlobal(
       "fetch",
       vi.fn(async (_input, init) => {
@@ -112,25 +242,80 @@ describe("classifyRepo", () => {
     );
     const { kv } = createFakeKV();
 
-    const result = await classifyRepo(
-      makeLinearBotEnv(kv),
-      "Update service",
-      null,
-      [],
-      null,
-      "Engineering",
-      "ENG",
-      null
-    );
+    const result = await classify(makeLinearBotEnv(kv, { CONTROL_PLANE: twoRepoControlPlane() }));
 
-    expect(timeoutSpy).toHaveBeenCalledWith(CLASSIFIER_REQUEST_TIMEOUT_MS);
-    expect(result).toEqual({
+    expect(result).toMatchObject({
       repo: null,
       confidence: "low",
-      reasoning:
-        "Could not classify repository automatically. Please reply with the repository name (e.g., `owner/repo`).",
-      alternatives: repos,
       needsClarification: true,
     });
+    expect(result.alternatives).toHaveLength(2);
   });
+
+  it("fires the Anthropic default path when CLASSIFICATION_MODEL is unset", async () => {
+    const { kv } = createFakeKV();
+    const env = makeLinearBotEnv(kv, { CONTROL_PLANE: twoRepoControlPlane() });
+    expect(env.CLASSIFICATION_MODEL).toBeUndefined();
+
+    const fetchMock = anthropicToolResponse("acme/beta");
+    vi.stubGlobal("fetch", fetchMock);
+
+    const result = await classify(env);
+
+    expect(result.repo?.id).toBe("acme/beta");
+    const [url, init] = fetchMock.mock.calls[0];
+    expect(url).toBe("https://api.anthropic.com/v1/messages");
+    expect(init!.headers).toMatchObject({ "x-api-key": "anthropic-key" });
+    const body = JSON.parse(init!.body as string);
+    expect(body.model).toBe("claude-haiku-4-5");
+  });
+
+  it("degrades rather than throwing on an unrecognised classification model prefix", async () => {
+    const { kv } = createFakeKV();
+    const env = makeLinearBotEnv(kv, {
+      CONTROL_PLANE: twoRepoControlPlane(),
+      CLASSIFICATION_MODEL: "mistral/mistral-large",
+    });
+
+    const fetchMock = vi.fn();
+    vi.stubGlobal("fetch", fetchMock);
+
+    const result = await classify(env);
+
+    expect(result.needsClarification).toBe(true);
+    expect(result.repo).toBeNull();
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it.each([
+    {
+      binding: "OPENAI_API_KEY",
+      model: "gpt-5.4-mini",
+      overrides: { OPENAI_API_KEY: undefined },
+    },
+    {
+      binding: "ANTHROPIC_API_KEY",
+      model: "claude-haiku-4-5",
+      overrides: { ANTHROPIC_API_KEY: undefined },
+    },
+  ])(
+    "degrades without calling out when $model is selected but $binding is unbound",
+    async ({ model, overrides }) => {
+      const { kv } = createFakeKV();
+      const env = makeLinearBotEnv(kv, {
+        CONTROL_PLANE: twoRepoControlPlane(),
+        CLASSIFICATION_MODEL: model,
+        ...overrides,
+      });
+
+      const fetchMock = vi.fn();
+      vi.stubGlobal("fetch", fetchMock);
+
+      const result = await classify(env);
+
+      expect(result.needsClarification).toBe(true);
+      expect(result.repo).toBeNull();
+      expect(fetchMock).not.toHaveBeenCalled();
+    }
+  );
 });
--- a/packages/linear-bot/src/classifier/index.test.ts
+++ b/packages/linear-bot/src/classifier/index.test.ts
@@ -1,11 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import {
-  anthropicMessagesResponseSchema,
-  classifyRepo,
-  classifyToolInputSchema,
-  CLASSIFICATION_REQUEST_TIMEOUT_MS,
-  openaiChatCompletionResponseSchema,
-} from "./index";
+import { anthropicMessagesResponseSchema, classifyRepo, classifyToolInputSchema } from "./index";
+import { CLASSIFICATION_REQUEST_TIMEOUT_MS } from "@open-inspect/shared/classification";
 import { clearReposLocalCache } from "./repos";
 import { createFakeKV, makeLinearBotEnv } from "../test-helpers";
 import type { Env } from "../types";
@@ -68,22 +63,6 @@ describe("classifyToolInputSchema", () => {
       confidence: "certain",
       reasoning: "Invalid confidence value.",
     });
-
-    expect(parsed.success).toBe(false);
-  });
-});
-
-describe("openaiChatCompletionResponseSchema", () => {
-  it("parses a response with the consumed message fields", () => {
-    const parsed = openaiChatCompletionResponseSchema.safeParse({
-      choices: [{ message: { content: "{}", refusal: null } }],
-    });
-
-    expect(parsed.success).toBe(true);
-  });
-
-  it("rejects a response without choices", () => {
-    const parsed = openaiChatCompletionResponseSchema.safeParse({});
 
     expect(parsed.success).toBe(false);
   });
@@ -198,7 +177,8 @@ describe("classifyRepo provider dispatch", () => {
 
     const body = JSON.parse(init!.body as string);
     expect(body.model).toBe("gpt-5.4-mini");
-    expect(body.temperature).toBe(0);
+    // gpt-5-family models reject an explicit temperature with HTTP 400.
+    expect(body).not.toHaveProperty("temperature");
     expect(body.max_completion_tokens).toBe(2000);
     expect(body).not.toHaveProperty("max_tokens");
     expect(body.response_format.type).toBe("json_schema");
@@ -281,4 +261,36 @@ describe("classifyRepo provider dispatch", () => {
     expect(result.repo).toBeNull();
     expect(fetchMock).not.toHaveBeenCalled();
   });
+
+  it.each([
+    {
+      binding: "OPENAI_API_KEY",
+      model: "gpt-5.4-mini",
+      overrides: { OPENAI_API_KEY: undefined },
+    },
+    {
+      binding: "ANTHROPIC_API_KEY",
+      model: "claude-haiku-4-5",
+      overrides: { ANTHROPIC_API_KEY: undefined },
+    },
+  ])(
+    "degrades without calling out when $model is selected but $binding is unbound",
+    async ({ model, overrides }) => {
+      const { kv } = createFakeKV();
+      const env = makeLinearBotEnv(kv, {
+        CONTROL_PLANE: twoRepoControlPlane(),
+        CLASSIFICATION_MODEL: model,
+        ...overrides,
+      });
+
+      const fetchMock = vi.fn();
+      vi.stubGlobal("fetch", fetchMock);
+
+      const result = await classify(env);
+
+      expect(result.needsClarification).toBe(true);
+      expect(result.repo).toBeNull();
+      expect(fetchMock).not.toHaveBeenCalled();
+    }
+  );
 });
--- a/packages/linear-bot/src/classifier/index.test.ts
+++ b/packages/linear-bot/src/classifier/index.test.ts
@@ -1,6 +1,9 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { anthropicMessagesResponseSchema, classifyRepo, classifyToolInputSchema } from "./index";
-import { CLASSIFICATION_REQUEST_TIMEOUT_MS } from "@open-inspect/shared/classification";
+import {
+  CLASSIFICATION_REQUEST_TIMEOUT_MS,
+  OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS,
+} from "@open-inspect/shared/classification";
 import { clearReposLocalCache } from "./repos";
 import { createFakeKV, makeLinearBotEnv } from "../test-helpers";
 import type { Env } from "../types";
@@ -179,7 +182,7 @@ describe("classifyRepo provider dispatch", () => {
     expect(body.model).toBe("gpt-5.4-mini");
     // gpt-5-family models reject an explicit temperature with HTTP 400.
     expect(body).not.toHaveProperty("temperature");
-    expect(body.max_completion_tokens).toBe(2000);
+    expect(body.max_completion_tokens).toBe(OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS);
     expect(body).not.toHaveProperty("max_tokens");
     expect(body.response_format.type).toBe("json_schema");
     expect(body.response_format.json_schema.strict).toBe(true);
--- a/packages/linear-bot/src/classifier/index.test.ts
+++ b/packages/linear-bot/src/classifier/index.test.ts
@@ -1,40 +1,14 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import type { RepoConfig } from "@open-inspect/shared/types/repository-catalog";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 import {
   anthropicMessagesResponseSchema,
-  CLASSIFIER_REQUEST_TIMEOUT_MS,
   classifyRepo,
   classifyToolInputSchema,
+  CLASSIFICATION_REQUEST_TIMEOUT_MS,
+  openaiChatCompletionResponseSchema,
 } from "./index";
+import { clearReposLocalCache } from "./repos";
 import { createFakeKV, makeLinearBotEnv } from "../test-helpers";
-
-const { getAvailableRepos, buildRepoDescriptions } = vi.hoisted(() => ({
-  getAvailableRepos: vi.fn(),
-  buildRepoDescriptions: vi.fn(),
-}));
-
-vi.mock("./repos", () => ({ getAvailableRepos, buildRepoDescriptions }));
-
-const repos: RepoConfig[] = ["api", "web"].map((name) => ({
-  id: `acme/${name}`,
-  owner: "acme",
-  name,
-  fullName: `acme/${name}`,
-  displayName: name,
-  description: `${name} repository`,
-  defaultBranch: "main",
-  private: true,
-}));
-
-beforeEach(() => {
-  getAvailableRepos.mockResolvedValue(repos);
-  buildRepoDescriptions.mockResolvedValue("- acme/api\n- acme/web");
-});
-
-afterEach(() => {
-  vi.restoreAllMocks();
-  vi.unstubAllGlobals();
-});
+import type { Env } from "../types";
 
 describe("anthropicMessagesResponseSchema", () => {
   it("parses a response with the consumed tool block fields", () => {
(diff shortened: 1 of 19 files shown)


ColeMurray (commented):
[Automated Review] Code-quality pass. Overall this is a strong PR — the contract-pinning tests, the both-polarity Terraform assertions, and the fail-closed credential gating are all appreciated. Two of the inline comments below answer questions you raised in the PR description (shared extraction: yes please; `classification_anthropic_api_key`: drop it). Comments marked **nit** are non-blocking.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -54,6 +60,69 @@ const CLASSIFY_TARGET_TOOL: Anthropic.Messages.Tool = {
...
+
+/**
+ * Resolve which provider serves a classification model id, and the bare id to
+ * send that provider (any `anthropic/`/`openai/` prefix stripped).
+ *
+ * There is no separate provider env var — the model id alone selects the
+ * provider, so `CLASSIFICATION_MODEL=gpt-5.4-mini` routes to OpenAI while the
+ * default `claude-haiku-4-5` keeps routing to Anthropic.
+ */
+function resolveClassificationProvider(model: string): {
```
ColeMurray: [Automated Review] Answering the offer in the PR description: yes — please extract this layer into `packages/shared`. `@open-inspect/shared/models` already exports `extractProviderAndModel`, which encodes this exact prefix convention; this function and its twin in `packages/linear-bot/src/classifier/index.ts` reimplement it, and the copies have already diverged at birth (different error strings, zod vs `as`-cast envelope parsing, explicit key guard vs `?? ""`, duplicated `CLASSIFICATION_REQUEST_TIMEOUT_MS` — AGENTS.md asks that each default be defined exactly once). The "keeps each Worker self-contained" rationale doesn't quite hold: both bots already depend on shared (this very file's sibling imports from it, and linear-bot's classifier imports `@open-inspect/shared/types/repository-catalog`).

Suggested scope, deliberately bounded:
- a throwing `resolveClassificationProvider` in `shared/models.ts` next to `extractProviderAndModel` (keep your throw-on-unknown semantics — they're better here than `extractProviderAndModel`'s silent anthropic fallback, and your tests already pin them)
- one `callOpenAIStructured(apiKey, model, prompt, { name, schema }): Promise<unknown>` plus the timeout constant in a shared module; each bot keeps its own output validation (`normalizeModelResponse` / `classifyToolInputSchema`) on the returned `unknown`
- Anthropic transports stay per-bot as they are — please don't unify those here.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -54,6 +60,69 @@ const CLASSIFY_TARGET_TOOL: Anthropic.Messages.Tool = {
...
 };
 
+/**
+ * Plain JSON Schema mirror of {@link CLASSIFY_TARGET_TOOL}'s input shape, used
+ * for OpenAI's strict structured-output mode (`response_format.json_schema`).
+ * Kept in sync with the Anthropic tool's `input_schema` by hand — the two
+ * provider SDKs use incompatible schema types, so there's no single shared
+ * declaration to derive both from.
+ */
+const CLASSIFY_TARGET_JSON_SCHEMA = {
```
ColeMurray: [Automated Review] This mirror is character-for-character identical to `CLASSIFY_TARGET_TOOL.input_schema` above, and the justifying comment doesn't hold up: the OpenAI call is a raw `fetch` with `JSON.stringify`, so there is no OpenAI SDK schema type in play — the only typed constraint anywhere is Anthropic's, which the tool already satisfies. This can be one line:

```ts
schema: CLASSIFY_TARGET_TOOL.input_schema,
```

identical wire bytes, and the hand-sync burden (plus this 26-line duplicate and its comment) disappears.

packages/linear-bot/src/classifier/index.ts
```diff
@@ -35,6 +40,67 @@ export const anthropicMessagesResponseSchema = z.object({
...
+  if (modelId.startsWith("claude-")) return { provider: "anthropic", model: modelId };
+  if (modelId.startsWith("gpt-")) return { provider: "openai", model: modelId };
+  throw new Error(`Unrecognized classification model id: ${modelId}`);
+}
+
+/**
+ * JSON schema for the classification result, shared by the Anthropic tool
+ * definition and the OpenAI strict structured-output schema.
+ */
+const classifyRepoJsonSchema = {
```
ColeMurray, line 54: [Automated Review] The doc comment says this schema is "shared by the Anthropic tool definition and the OpenAI strict structured-output schema", but `callAnthropic` below still carries its own inline near-copy (the `input_schema` at ~L180–202) — the sharing it describes isn't implemented. Since the Anthropic call here is also raw `fetch`, there's no type friction: wire it in (`input_schema: classifyRepoJsonSchema`), or if you want the Anthropic request body byte-stable, keep a base schema and spread `additionalProperties: false` onto the OpenAI side only. Either way the comment and the code should agree.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -352,26 +492,35 @@ export class RepoClassifier {
...
+                    {
+                      role: "user",
+                      content: prompt,
+                    },
+                  ],
+                },
+                { signal: AbortSignal.timeout(CLASSIFICATION_REQUEST_TIMEOUT_MS) }
+              )
+            )
+          : await callOpenAI(this.env.OPENAI_API_KEY ?? "", model, prompt);
```
ColeMurray: [Automated Review] Two things on this dispatch:

1. The entire Anthropic request is buried three levels deep inside one ternary arm. Linear-bot got this shape right — named `callAnthropic` / `callOpenAI` helpers and a short dispatch. Extracting a `callAnthropic` here (method or free function) makes this read the same way:

```ts
const llmResult =
  provider === "anthropic"
    ? await this.callAnthropic(model, prompt)
    : await callOpenAI(openaiKey, model, prompt);
```

2. `this.env.OPENAI_API_KEY ?? ""` ships a doomed request with an empty bearer token and surfaces as a confusing OpenAI 401 in the logs. Linear-bot throws a clear `requires OPENAI_API_KEY to be set` before calling out — please use the same explicit guard here so both bots fail identically for the same misconfiguration.

packages/slack-bot/src/types/index.ts
```diff
@@ -36,6 +36,7 @@ export interface Env {
   SLACK_SIGNING_SECRET: string;
   SLACK_APP_TOKEN?: string;
   ANTHROPIC_API_KEY: string;
+  OPENAI_API_KEY?: string;
```
ColeMurray, line 49: [Automated Review] With this PR, the Terraform binding for `ANTHROPIC_API_KEY` becomes conditional — an OpenAI-classifier deployment binds no Anthropic key on the bots — but the line above still declares it as required `string`. The type is now knowingly false in a supported configuration (the lazy-client comment in the classifier says as much), and in linear-bot the possibly-undefined value flows into `callAnthropic(apiKey: string, …)`, safe only by control flow the types can't see. Please make it `ANTHROPIC_API_KEY?: string` here and in `packages/linear-bot/src/types.ts`, and guard the anthropic dispatch branch symmetrically to the OpenAI one.

terraform/environments/production/variables.tf
```diff
@@ -303,6 +303,59 @@ variable "anthropic_api_key" {
...
+      startswith(var.classification_model, "anthropic/") ||
+      startswith(var.classification_model, "claude-") ||
+      startswith(var.classification_model, "openai/") ||
+      startswith(var.classification_model, "gpt-")
+    )
+    error_message = "classification_model must be an Anthropic id (\"anthropic/...\" or \"claude-...\") or an OpenAI id (\"openai/...\" or \"gpt-...\")."
+  }
+}
+
+variable "classification_anthropic_api_key" {
```
ColeMurray: [Automated Review] Answering the question in the PR description: prefer dropping this variable entirely. `anthropic_api_key` is `nullable = false` with an unconditional non-blank validation, so this block's final disjunct (`trimspace(var.anthropic_api_key) != ""`) is always true in any plan that survives the primary validation — the validation here is unreachable dead logic, which is also the tell that the variable isn't earning its keep. Provider symmetry alone doesn't pay for a variable + fallback local + two workflow lines + docs rows (dropping it also erases the entire `GETTING_STARTED.md` table realignment, since the wide column name goes away). It can be reintroduced the day someone actually needs a separate classifier billing identity.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -191,18 +260,89 @@ function extractStructuredResponse(response: Anthropic.Messages.Message): LLMRes
   return normalizeModelResponse(toolUseBlock.input);
 }
 
+/**
+ * Call OpenAI's Chat Completions API with strict JSON-schema structured
+ * output, then funnel the parsed object through the same
+ * {@link normalizeModelResponse} validation as the Anthropic tool-use path.
+ */
+async function callOpenAI(apiKey: string, model: string, prompt: string): Promise<LLMResponse> {
+  const response = await fetch("https://api.openai.com/v1/chat/completions", {
```
ColeMurray: [Automated Review] Context you couldn't have known, plus one pre-merge gate: the previous OpenAI classifier (#1366) shipped with all upstreams mocked and was reverted the same day (#1375, #1374) because workerd stamps an unremovable `CF-Worker` header on every subrequest and the old endpoint's edge 403s it — something no mock can see. `api.openai.com` is a different origin and expected to be fine, but that has never been verified from workerd. One request from `wrangler dev` (local mode stamps the header too, no `--remote` needed) settles it before merge. Discriminator: a real application rejection comes back as JSON `invalid_request_error`; an edge block comes back as an HTML page. We'll run this check on our side — but if you have a key handy and beat us to it, pasting the result here works too.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -191,18 +260,89 @@ function extractStructuredResponse(response: Anthropic.Messages.Message): LLMRes
...
+    method: "POST",
+    headers: {
+      Authorization: `Bearer ${apiKey}`,
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      model,
+      temperature: 0,
+      max_completion_tokens: 2000,
```
ColeMurray: [Automated Review] **nit**: 2000 here vs Anthropic's `max_tokens: 500` is an unexplained 4× delta, inlined in both bots' copies. If the reason is that gpt-5-family reasoning tokens are billed inside `max_completion_tokens` (so 500 could truncate before the JSON emerges), a short comment saying so would pin it down. Moot if these collapse into a shared helper per the other comment.

packages/slack-bot/src/classifier/index.test.ts
```diff
@@ -671,4 +672,119 @@ describe("RepoClassifier", () => {
...
+    it("sends the OpenAI chat-completions contract for a gpt-* model", async () => {
+      const fetchMock = vi.fn().mockResolvedValue(
+        openAiFetchResponse({
+          targetId: "acme/prod",
+          confidence: "high",
+          reasoning: "Mentions prod.",
+          alternatives: [],
+        })
+      );
+      vi.stubGlobal("fetch", fetchMock);
```
ColeMurray, line 712: [Automated Review] **nit**: these tests stub the global `fetch` but the suite never restores it — the linear-bot suite calls `vi.unstubAllGlobals()`. Worth adding here too so any test added after this describe block doesn't inherit a stale fetch stub.

ColeMurray (changes requested):
-

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Done in `75b2beb6` / `e11daf0c`, to the bounded scope you sketched.

`packages/shared/src/classification.ts` (new `@open-inspect/shared/classification` subpath) now owns exactly three things:

- `resolveClassificationProvider` — your throw-on-unknown semantics kept, with a doc comment saying explicitly why it does not reuse `extractProviderAndModel`'s silent anthropic fallback.
- `callOpenAIStructured(apiKey, model, prompt, { name, schema }): Promise<unknown>` — returns `unknown`; each bot still validates with its own `normalizeModelResponse` / `classifyToolInputSchema`.
- `CLASSIFICATION_REQUEST_TIMEOUT_MS`, `DEFAULT_CLASSIFICATION_MODEL`, and `OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS`, each defined once.

Anthropic transports stayed per-bot as you asked.

One deliberate call on the "zod vs `as`-cast" divergence you flagged: the shared transport takes linear-bot's zod envelope parsing (`openAiChatCompletionEnvelopeSchema`), not slack-bot's cast — unifying on the weaker of two behaviours would have been a silent regression for linear-bot. New coverage in `packages/shared/src/classification.test.ts` (14 cases).

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Fixed in `e11daf0c` — you were right that the justifying comment did not hold up. `CLASSIFY_TARGET_JSON_SCHEMA` and its 26-line comment are gone; the OpenAI call now passes `schema: CLASSIFY_TARGET_TOOL.input_schema` directly, exactly as you wrote it. Identical wire bytes: the Anthropic tool schema already carried `additionalProperties: false`, which is precisely what OpenAI's `strict: true` requires.

packages/linear-bot/src/classifier/index.ts
rhlsthrm: Fixed in `e11daf0c`, taking your second option so the Anthropic request body stays byte-stable.

`callAnthropic`'s inline near-copy is deleted; it now uses `input_schema: classifyRepoJsonSchema`. Because that base schema deliberately omits `additionalProperties`, the Anthropic bytes are unchanged from before this PR. A one-line `classifyRepoStrictJsonSchema = { ...classifyRepoJsonSchema, additionalProperties: false }` feeds the OpenAI side, which needs the flag for `strict: true`. The doc comment now describes what the code actually does, including why the flag is spread on rather than baked in.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Both done in `e11daf0c`.

1. Extracted `RepoClassifier.callAnthropic(apiKey, model, prompt)`, so the dispatch now reads the way linear-bot's does — a short ternary over two named helpers, no request buried three levels deep in a ternary arm.
2. `?? ""` is gone. Both bots now share an identical `requireProviderKey(key, binding, modelId)` that throws ``Classification model "<id>" requires <BINDING> to be set`` before any request leaves the Worker — and it guards the Anthropic branch symmetrically, not just OpenAI. Covered by a parameterised test in each bot asserting the classifier degrades to the picker/clarification result and `fetch` is never called, for both an unbound `OPENAI_API_KEY` and an unbound `ANTHROPIC_API_KEY`.

packages/slack-bot/src/types/index.ts
rhlsthrm: Fixed in `e11daf0c`. `ANTHROPIC_API_KEY` is now `?: string` in both `packages/slack-bot/src/types/index.ts` and `packages/linear-bot/src/types.ts`, with a shared comment noting that the deployment binds exactly the one `CLASSIFICATION_MODEL` selects.

The type change is what forced the guard to be real rather than decorative: with the key optional, `callAnthropic(apiKey: string, …)` no longer typechecks on control flow the compiler cannot see, so the Anthropic branch had to go through `requireProviderKey` too. The types and the runtime guards now agree in both bots.

terraform/environments/production/variables.tf
rhlsthrm: Agreed and dropped, in `e58ab813`. You are right that the final disjunct was unreachable — `anthropic_api_key` is `nullable = false` with an unconditional non-blank validation, so `trimspace(var.anthropic_api_key) != ""` is true in any plan that gets that far.

Removed: the variable, `local.classification_anthropic_api_key`, both `TF_VAR_` workflow lines, the `terraform/README.md` row, and the `tfvars.example` entry. An Anthropic classifier is now served by the existing required `anthropic_api_key`, so no deployment needs a new value.

As you predicted, this also erased the entire `GETTING_STARTED.md` table realignment — with the wide column name gone, the secrets table is restored to its original widths and the docs diff is now purely additive (+13 lines, no reflow).

Backwards compatibility is pinned rather than asserted: `tests/classifier_provider.tftest.hcl` gained an assertion that the Anthropic binding's value is `var.anthropic_api_key`. `terraform test` — 18 passed, 0 failed.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Thank you for the #1366 / #1374 / #1375 context — that is exactly the kind of failure no mock can reach, and it is a fair pre-merge gate.

**I could not run it, and this is the one item on this PR I cannot close myself.** No authorized OpenAI credential is available to me: there is no `OPENAI_API_KEY` in this environment, the `CLASSIFICATION_OPENAI_API_KEY` Actions secret is not readable (`gh secret list` on `ColeMurray/background-agents` returns HTTP 403), and Cloudflare Worker secrets are write-only. I am not going to source a key from anywhere else to satisfy a check — so the `wrangler dev` → `api.openai.com` smoke remains **a human-owned blocker on your side**, as you offered.

To make it a copy-paste when someone with a key picks it up — local mode stamps `CF-Worker` too, no `--remote` needed:

```
curl -sS -o /tmp/openai-smoke.txt -w '%{http_code} %{content_type}\n' \
  http://127.0.0.1:8787/<any route that reaches the classifier>
```

Discriminator, as you framed it: a real application rejection returns JSON `invalid_request_error`; an edge block returns an HTML page. `head -c 200 /tmp/openai-smoke.txt` settles which.

Everything else in this review is addressed and verified; this is the only outstanding item.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Pinned rather than left unexplained — and, as you guessed, it did collapse into the shared helper. `OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS` is now declared once in `packages/shared/src/classification.ts`, with the reason you inferred written down:

> Four times the Anthropic tool-call budget because gpt-5-family reasoning tokens are billed inside `max_completion_tokens`: a tighter cap can be spent on reasoning before the structured JSON is emitted, truncating the response.

Both bots' inlined `2000` literals are gone.

packages/slack-bot/src/classifier/index.test.ts
rhlsthrm: Added in `e11daf0c` — `afterEach(() => { vi.unstubAllGlobals(); })` on the `provider selection` describe block, matching the linear-bot suite. It is load-bearing now rather than hygienic: the two new missing-provider-key tests in that same block stub `fetch` and assert it is never called, so a stub leaking out of the block would be exactly the kind of pass-for-the-wrong-reason you were guarding against.

packages/linear-bot/src/types.ts
```diff
@@ -33,6 +33,8 @@ export interface Env {
   LINEAR_WEBHOOK_SECRET: string;
   LINEAR_API_KEY?: string; // kept for backward compat / fallback
   ANTHROPIC_API_KEY: string;
+  CLASSIFICATION_MODEL?: string; // Optional override, e.g. "gpt-5.4-mini"; defaults to Anthropic claude-haiku-4-5
```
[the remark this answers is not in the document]
rhlsthrm: Fixed in `e11daf0c` — the comment now reads `// Optional override; defaults to DEFAULT_CLASSIFICATION_MODEL`, referencing the named constant (now in `@open-inspect/shared/classification`) rather than restating the literal.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -191,18 +260,89 @@ function extractStructuredResponse(response: Anthropic.Messages.Message): LLMRes
...
+      messages: [{ role: "user", content: prompt }],
```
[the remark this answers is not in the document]
rhlsthrm: Confirmed and fixed in `75b2beb6`. Both request builders collapsed into the shared `callOpenAIStructured`, which sends no `temperature` at all, with the reason recorded at the call site. Both bots' assertions flipped from `expect(body.temperature).toBe(0)` to `expect(body).not.toHaveProperty("temperature")`, plus the same assertion on the shared transport's own test.

packages/slack-bot/src/classifier/index.ts
```diff
@@ -352,26 +492,35 @@ export class RepoClassifier {
...
-            role: "user",
-            content: prompt,
-          },
-        ],
-      });
-
-      const llmResult = extractStructuredResponse(response);
+      const { provider, model } = resolveClassificationProvider(
+        this.env.CLASSIFICATION_MODEL || "claude-haiku-4-5"
+      );
```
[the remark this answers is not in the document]
rhlsthrm: Done, and taken all the way to the shared home you identified. `DEFAULT_CLASSIFICATION_MODEL`, `CLASSIFICATION_REQUEST_TIMEOUT_MS`, and `resolveClassificationProvider` now live once in `packages/shared/src/classification.ts`; both bots import them. The inline `"claude-haiku-4-5"` at the Slack call site is gone.

packages/slack-bot/src/classifier/index.ts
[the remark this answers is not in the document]
rhlsthrm: Fixed in `e11daf0c`, and made symmetric. `?? ""` is replaced by a shared-shape `requireProviderKey` helper present in both bots, which throws ``Classification model "<id>" requires <BINDING> to be set``. It guards the Anthropic branch as well as the OpenAI one — the Anthropic key is now optional in both `Env` types, since a deployment binds only the provider `CLASSIFICATION_MODEL` selects. Covered by a parameterised test per bot for both unbound bindings.

packages/linear-bot/src/classifier/index.test.ts
```diff
@@ -63,3 +67,230 @@ describe("classifyToolInputSchema", () => {
...
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    const [url, init] = fetchMock.mock.calls[0];
+    expect(url).toBe("https://api.openai.com/v1/chat/completions");
+    expect(init!.headers).toMatchObject({ Authorization: "Bearer openai-key" });
+
+    const body = JSON.parse(init!.body as string);
+    expect(body.model).toBe("gpt-5.4-mini");
+    // gpt-5-family models reject an explicit temperature with HTTP 400.
+    expect(body).not.toHaveProperty("temperature");
+    expect(body.max_completion_tokens).toBe(2000);
```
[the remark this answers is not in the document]
rhlsthrm: Fixed in `11921442` — both test files now import `OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS` from `@open-inspect/shared/classification` and assert against it, so the literal `2000` exists in exactly one place.

terraform/environments/production/terraform.tfvars.example
```diff
@@ -222,6 +222,17 @@ linear_webhook_secret = ""   # Webhook Signing Secret from the application confi
 # From: https://console.anthropic.com/
 anthropic_api_key = ""
 
+# OpenAI credential for the Slack/Linear bot classifiers. Required only when
+# classification_model names an OpenAI model; an Anthropic classifier model is
+# served by anthropic_api_key above.
+# From: https://platform.openai.com/api-keys
+classification_openai_api_key = ""
```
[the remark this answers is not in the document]
rhlsthrm: Fixed in `11921442`. Both now state the actual condition — an OpenAI `classification_model` **and** the Slack or Linear bot enabled — matching the variable validation, which short-circuits when both bots are disabled (pinned by the existing `openai_model_without_bots_needs_no_key` test).

terraform/environments/production/variables.tf
```diff
@@ -303,6 +303,41 @@ variable "anthropic_api_key" {
...
+    )
+    error_message = "classification_model must be an Anthropic id (\"anthropic/...\" or \"claude-...\") or an OpenAI id (\"openai/...\" or \"gpt-...\")."
+  }
+}
+
+variable "classification_openai_api_key" {
+  description = "OpenAI API key used specifically by the Slack and Linear bot classifiers. Required when classification_model is an OpenAI model and a classifier bot is enabled."
+  type        = string
+  sensitive   = true
+  default     = ""
```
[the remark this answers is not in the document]
rhlsthrm: Both valid, fixed in `11921442`.

`classification_model` and `classification_openai_api_key` are now `nullable = false`, so an explicit `null` cannot reach a `trimspace`/`startswith` expression.

The prefix hole was real: `startswith("claude-", "claude-")` is true, so a bare prefix passed validation and would have reached the bots as a value their resolver accepts and forwards to the provider verbatim. The condition now requires a non-blank id after the matched prefix:

```hcl
condition = anytrue([
  for prefix in ["anthropic/", "claude-", "openai/", "gpt-"] :
  startswith(var.classification_model, prefix) &&
  trimspace(substr(var.classification_model, length(prefix), -1)) != ""
])
```

Four boundary cases added to `tests/classifier_provider.tftest.hcl`: `claude-` rejected, `openai/` rejected, `anthropic/   ` (whitespace-only) rejected, and `claude-x` still accepted so the tightening does not overshoot. Full suite: `terraform test` — **22 passed, 0 failed**.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Current tip 82e36abd retains the bounded shared extraction in 75b2beb6 and e11daf0c: resolver, defaults and timeout, and OpenAI transport live in @open-inspect/shared; bot validation and Anthropic transports remain local. Fresh evidence: shared build, 644 shared tests, 224 Linear tests, 426 Slack tests, and all relevant typechecks and lints passed.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Current tip 82e36abd retains the schema deduplication from e11daf0c: Slack passes CLASSIFY_TARGET_TOOL.input_schema directly to the shared OpenAI transport. Fresh Slack evidence: 426 tests, typecheck, lint, and Prettier passed.

packages/linear-bot/src/classifier/index.ts
rhlsthrm: Current tip 82e36abd retains the Linear schema deduplication from e11daf0c: Anthropic uses classifyRepoJsonSchema and OpenAI derives only the strict additionalProperties flag. Fresh Linear evidence: 224 tests, typecheck, lint, and Prettier passed.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Current tip 82e36abd retains the short named-provider dispatch and symmetric requireProviderKey guards from e11daf0c. Missing Anthropic or OpenAI bindings fail before fetch and are covered in both bot suites; 224 Linear and 426 Slack tests passed.

packages/slack-bot/src/types/index.ts
rhlsthrm: Current tip 82e36abd keeps both provider Env bindings optional and guards the selected branch symmetrically. Fresh evidence: shared, Linear, and Slack typechecks passed, plus the missing-key regression cases in both bot suites.

terraform/environments/production/variables.tf
rhlsthrm: Current tip 82e36abd contains no classification_anthropic_api_key path; Anthropic classification reuses required anthropic_api_key and only one provider secret is bound. Fresh Terraform evidence: fmt and validate passed; 22 tests passed, 0 failed, including both binding polarities and backwards compatibility.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Live-smoke blocker: no authorized existing OpenAI credential is available in this environment. OPENAI_API_KEY and CLASSIFICATION_OPENAI_API_KEY are absent; GitHub and Cloudflare secrets are not readable credentials. Therefore the requested real wrangler dev to api.openai.com smoke remains human-owned and was not fabricated. Current tip: 82e36abd.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Current tip 82e36abd retains the single shared OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS constant and its gpt-5 reasoning-token rationale. Fresh shared evidence: build and 644 tests passed.

packages/slack-bot/src/classifier/index.test.ts
rhlsthrm: Current tip 82e36abd retains vi.unstubAllGlobals cleanup and adds the missing empty-choices transport regression. Fresh evidence: 426 Slack tests and 644 shared tests passed, with lint and formatting clean.

packages/linear-bot/src/classifier/index.ts
rhlsthrm: Rebased onto current upstream c3ec1015. The schema fix remains at head a922d3ac: Anthropic uses classifyRepoJsonSchema directly, while OpenAI adds only additionalProperties: false for strict mode. Linear tests pass (231/231).

packages/slack-bot/src/types/index.ts
rhlsthrm: Rebased onto current upstream c3ec1015. Both Env bindings remain optional at head a922d3ac, and both provider branches still pass through requireProviderKey before fetch. Shared, Linear, and Slack typechecks pass; both missing-key paths remain covered.

packages/slack-bot/src/classifier/index.test.ts
rhlsthrm: Rebased onto current upstream c3ec1015. The provider-selection block still restores global fetch stubs with afterEach(vi.unstubAllGlobals) at head a922d3ac. Slack tests pass (426/426).

rhlsthrm:
Rebased onto current upstream `c3ec1015`; new head is `a922d3ac`.

The rebase exposed upstream #1509's new Linear-only 10s timeout beside this PR's shared 15s classifier timeout. I reconciled that semantic conflict at the shared boundary: `CLASSIFICATION_REQUEST_TIMEOUT_MS` is now the single 10s source used by both providers and both bots, and retained upstream's timeout-to-clarification regression coverage.

Fresh verification:
- shared: 650 tests
- linear-bot: 231 tests
- slack-bot: 426 tests
- all three package typechecks
- affected-file ESLint and Prettier
- Terraform fmt + validate + test: 22 passed

The requested live `wrangler dev` → `api.openai.com` smoke is still not executable here because no authorized `OPENAI_API_KEY` is available. That external pre-merge gate remains human-owned; no result was fabricated.

rhlsthrm:
@ColeMurray, please re-review current head `a922d3ac`. All code findings are retained after the rebase; the real workerd/OpenAI smoke remains the only external gate I could not execute without an authorized key.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Addressed in 7462850b: provider/model resolution, the shared 10s deadline, and strict OpenAI transport now live once in `@open-inspect/shared/classification`; each bot retains only its Anthropic transport and result validation.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Addressed in 7462850b: the duplicate Slack OpenAI JSON-schema mirror is deleted; both Anthropic and OpenAI now use `CLASSIFY_TARGET_TOOL.input_schema`.

packages/linear-bot/src/classifier/index.ts
rhlsthrm: Addressed in 7462850b: `callAnthropic` now consumes `classifyRepoJsonSchema` directly, while the OpenAI-only strict wrapper adds `additionalProperties: false`; the comment and implementation now agree.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Addressed in 7462850b: Slack Anthropic dispatch is a named `callAnthropic` method and the provider branch is now a short two-arm dispatch. Provider credentials fail closed through the shared `requireClassificationProviderKey` guard.

packages/slack-bot/src/types/index.ts
rhlsthrm: Addressed in 7462850b: both Worker Env types mark provider keys optional, matching conditional Terraform bindings; the shared runtime guard requires the selected provider key before any outbound request.

terraform/environments/production/variables.tf
rhlsthrm: Addressed in 7462850b: `classification_anthropic_api_key` was removed completely. Anthropic classification reuses the existing required `anthropic_api_key`; OpenAI uses the scoped optional classification key only when selected.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: The requested live gate now passes on 7462850b: I ran the rebased shared classifier through local workerd (`wrangler dev`) against the real OpenAI API. `gpt-5.4-mini` strict `json_schema` returned HTTP 200 with `{"ok":true}`; this exercises the Cloudflare Worker subrequest path that previously edge-403ed.

packages/slack-bot/src/classifier/index.ts
rhlsthrm: Addressed in 7462850b: `OPENAI_CLASSIFICATION_MAX_COMPLETION_TOKENS` is single-sourced in shared code with the rationale that gpt-5 reasoning tokens consume the same completion budget; both bot tests import the constant.

packages/slack-bot/src/classifier/index.test.ts
rhlsthrm: Addressed in 7462850b: the Slack classifier suite now runs `vi.unstubAllGlobals()` in `afterEach`, matching the Linear suite and preventing fetch leakage.

rhlsthrm:
Rebased and squashed onto current `main`; force-pushed as `7462850b`.

All nine requested changes are present, including the shared classifier transport/provider layer, single shared credential guard, removed schema/key duplication, optional Env keys matching conditional bindings, strict Terraform validation, test cleanup, and documented token budget. The new classifier Terraform suite is now part of the unconditional CI validate job rather than a dead test file.

Verification:
- shared: 708 tests;
- Slack bot: 426 tests;
- Linear bot: 231 tests;
- Terraform: 26 tests, validate/fmt clean;
- package typechecks, ESLint, and Prettier clean;
- live local-workerd → OpenAI strict-json-schema request returned HTTP 200 with the expected object.

I replied to each review thread with its disposition. I left the human-authored threads open for maintainer review.

ColeMurray (approved):

ColeMurray:
thanks!

PR #1426, opened 2026-08-14
feat: add completed workflow run automations

roelofb:
Adds support for `workflow_run.completed` GitHub Event automations, including workflow-name and
conclusion filters, rerun-aware deduplication, UI controls, tests, and setup documentation.

## Review feedback addressed

**Conditions are now scoped per event type.** Previously every GitHub condition was offered for
every GitHub event, so `workflow_name = CI` could be saved on `pull_request.opened` and never
match. `GITHUB_WEBHOOK_EVENT_CATALOG` is now the canonical map from event type to supported
conditions; `eventType` is threaded through both the form and server-side `validateConditions`, and
switching event type drops conditions the new one cannot answer rather than leaving them behind.

**Workflow-run concurrency keys on the run, not the workflow name.** `concurrencyKey` is
`workflow_run:${run.id}`; `run_attempt` stays in the trigger key. Two runs of the same workflow no
longer collide, while reruns of one run still share a concurrency scope.

**One canonical `conclusion` field.** Check suites and workflow runs both populate `conclusion`;
the separate `checkConclusion` normalized field is gone. The GitHub condition handlers and
conclusion constants moved out of the cross-source registry into `triggers/github/conditions.ts`.

## Worth flagging

**Path-pattern filtering is no longer offered for GitHub events.** No GitHub webhook payload
carries a file list, so `changedFiles` was never populated and a `path_glob` condition could only
ever prevent an automation from firing. Making the catalog canonical surfaced this, so the handler
now declares an empty `appliesTo`. Its schema variant is kept so persisted configs still parse.
This was not part of the review feedback — flagging it because it removes a filter the UI
previously offered.

**`check_conclusion` is retained as a legacy alias.** It reads the canonical `conclusion` field for
persisted-config compatibility and is no longer offered for new automations. Collapsing it away
entirely would need either a D1 migration or canonicalization at the parse boundary — happy to do
that here if you would prefer it in this PR rather than a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added GitHub workflow-run completion automations.
  * Added workflow name and run conclusion filters, including success, failure, cancelled, and timed-out outcomes.
  * Added workflow-run details to automation event context.
  * Added event-specific GitHub condition filtering and configuration guidance for required permissions and webhook events.

* **Bug Fixes**
  * Improved validation for GitHub event conditions and conclusions.
  * Prevented duplicate processing when workflow runs are rerun.
  * Automatically removes incompatible conditions when changing event types.

* **Documentation**
  * Updated automation and setup documentation for GitHub event triggers and workflow-run automations.

### DIFF
--- a/packages/shared/src/triggers/github/normalizer.ts
+++ b/packages/shared/src/triggers/github/normalizer.ts
@@ -9,6 +9,7 @@ import {
   buildIssueContextBlock,
   buildPullRequestContextBlock,
   buildReviewCommentContextBlock,
+  buildWorkflowRunContextBlock,
 } from "./context";
 import {
   GITHUB_WEBHOOK_EVENT_CATALOG,
@@ -17,12 +18,14 @@ import {
   issuesEventSchema,
   pullRequestEventSchema,
   pullRequestReviewCommentEventSchema,
+  workflowRunEventSchema,
   type CheckSuitePayload,
   type GitHubEventBase,
   type IssueCommentPayload,
   type IssuesPayload,
   type PullRequestPayload,
   type PullRequestReviewCommentPayload,
+  type WorkflowRunPayload,
 } from "./webhook-types";
 
 // ─── Supported event type map ─────────────────────────────────────────────────
@@ -101,6 +104,12 @@ export function normalizeGitHubEvent(
       const parsed = checkSuiteEventSchema.safeParse(payload);
       if (!parsed.success) return null;
       return normalizeCheckSuite(eventType, parsed.data);
+    }
+
+    case "workflow_run": {
+      const parsed = workflowRunEventSchema.safeParse(payload);
+      if (!parsed.success) return null;
+      return normalizeWorkflowRun(eventType, parsed.data);
     }
 
     case "issues": {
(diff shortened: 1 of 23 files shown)


packages/shared/src/triggers/registry.ts
```diff
@@ -84,16 +103,30 @@ const sharedConditions = {
...
+  },
   check_conclusion: {
+    appliesTo: ["github"] as const,
+    validate: validateGitHubConclusion,
+    evaluate(c: { value: string }, event: AutomationEvent) {
+      if (event.source !== "github") return true;
+      return event.checkConclusion === c.value;
+    },
+  },
+  workflow_name: {
```
ColeMurray: [Automated Review] These conditions are scoped only to the broad `github` source, so the form offers `workflow_name`, `conclusion`, and `check_conclusion` for every GitHub event type and the server accepts those combinations. For example, `workflow_name = CI` can be saved on `pull_request.opened` but can never match; changing the selected event type also leaves incompatible conditions behind. Please make condition applicability event-type-aware and pass `eventType` through both UI and server validation. The GitHub event catalog could be the canonical mapping from event type to supported conditions, with tests that reject incompatible pairs.

packages/shared/src/triggers/github/normalizer.ts
```diff
@@ -263,6 +273,40 @@ function normalizeCheckSuite(eventType: string, payload: CheckSuitePayload): Git
...
+  payload: WorkflowRunPayload
+): GitHubAutomationEvent {
+  const run = payload.workflow_run;
+  const conclusion = run.conclusion ?? undefined;
+
+  return {
+    source: "github",
+    eventType,
+    triggerKey: `workflow_run:${run.id}:${run.run_attempt}`,
+    concurrencyKey: `workflow:${run.name}`,
```
ColeMurray: [Automated Review] This concurrency key groups every distinct run sharing a workflow display name. The scheduler treats an equal key as an overlap and records the later completion as skipped, so a second CI run that completes while the first automation session is active is silently lost after its webhook has been acknowledged. Please scope concurrency to the logical run, e.g. `workflow_run:${run.id}`, while keeping `run_attempt` in the trigger key. Add coverage showing that different run IDs with the same name can both be admitted and that attempts of one run share the intended concurrency scope.

packages/shared/src/triggers/github/normalizer.ts
```diff
@@ -254,6 +263,7 @@ function normalizeCheckSuite(eventType: string, payload: CheckSuitePayload): Git
     repoName: getRepoName(payload),
     branch: checkSuite.head_branch ?? undefined,
     actor: getActor(payload),
+    conclusion,
```
ColeMurray, line 266: [Automated Review] This introduces two normalized sources of truth for the same provider fact: check suites populate both `conclusion` and `checkConclusion`, while workflow runs populate only `conclusion`, and the registry duplicates the corresponding handlers. Please keep one canonical `conclusion` event field. The legacy `check_conclusion` condition can read that field for persisted-config compatibility while only the preferred condition is exposed for new configurations. This also gives the GitHub-specific handlers/constants a natural home in `github/conditions.ts` instead of further growing the cross-source registry.

ColeMurray:
superseded by #1633

PR #1428, opened 2026-08-14
test(slack-bot): make the Starting-status assertions deterministic

duboff:
## Summary

Two tests in `packages/slack-bot/src/index.test.ts` race a background promise against a 25 ms `setTimeout` and assert the background work wins. That budget holds on a developer machine and loses on a loaded CI runner, producing `expected 'blocked' to be 'complete'` — the timer winning rather than anything being wrong with the code.

We hit this on a fork, where it failed three merge-gating pipelines in one day while passing 423/423 locally on every run. The same code is here, so it will surface for you too — the tests date from #646, and the work later added to that background path (parent→child follow-ups, unread indicators) is what pushed it past the budget.

## A second bug, independent of the flake

The failing test's cleanup — resolving its deferred and restoring the global `fetch` spy — sits on the **last lines of the test body**:

```js
statusDeferred.resolve(...);
await new Promise((resolve) => setTimeout(resolve, 0));
slackFetch.mockRestore();
```

When the race assertion throws, none of that runs. The spy stays installed with a Slack status call in flight, and that call lands inside the *next* test. On our fork the symptom was `sets Starting status for repo-selection starts before session creation` failing with a `channel_id: 'D123'` body in its array — the DM channel from the previous test, not the `C123` it sends. One timing flake, two reported failures in different `describe` blocks.

This pattern is not limited to the two tests here: there are ten-plus tests in this file ending in `slackFetch.mockRestore()`, and no `afterEach` anywhere in it. Any of them can poison its neighbours by failing early.

## The change

- The property under test — "session creation does not wait for the Slack status call" — needs no clock. With the status deferred left unresolved, awaiting the background promise proves it: a chain that *did* wait could never settle, and vitest's own timeout reports that.
- `hasSettled()` answers "has this settled yet" by draining the microtask queue instead of sleeping. A promise blocked on an unresolved deferred cannot settle however many turns elapse, so the answer no longer depends on machine load.
- Teardown moves to an `afterEach`, so a test that throws part-way can no longer leave a global spy installed for its neighbours.

I left the ten-plus trailing `mockRestore()` calls in place — they are now redundant but harmless, and removing them felt like a separate cleanup for you to decide on.

## Test plan

- [x] `npm test -w @open-inspect/slack-bot` on this branch: 417 passed, 4 failed
- [x] Same command on clean `main`: 417 passed, **the identical 4 failed** — those failures are pre-existing (attachment/file tests in `attachments.test.ts` and `index.test.ts`) and unrelated to this change, which is neutral on them
- [x] The two target tests pass: `-t "Starting status"` → 7 passed

## One caveat

I could not reproduce the flake locally even with all cores saturated — the CI slowness involved is cgroup throttling and I/O rather than raw contention. So the case rests on the CI error message, which is definitionally the timer winning, plus the fact that the fix removes the clock the assertion depended on. The cross-test-pollution half is demonstrable independently of any flake.

## Summary by CodeRabbit

* **Tests**
  * Improved Slack bot event test reliability and consistency.
  * Added deterministic handling for asynchronous operations.
  * Ensured test mocks are restored between event tests.
  * Strengthened coverage for concurrent starting-status behavior.

### DIFF
--- a/packages/slack-bot/src/index.test.ts
+++ b/packages/slack-bot/src/index.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi, beforeEach } from "vitest";
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
 import type { Env } from "./types";
 import type * as SlackModule from "@open-inspect/shared/slack";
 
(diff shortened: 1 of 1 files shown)


ColeMurray:
the original test was low quality. rather than adding the for loop, let's remove it for now.
https://github.com/ColeMurray/background-agents/pull/1446

can reopen if desired

PR #1442, opened 2026-08-15
fix(control-plane): stop unarchivable rows pinning the draft sweep

ColeMurray:
Follow-up to #1420 and #1425. The abandoned draft sweep has archived **zero**
rows in production since it shipped, while 266 sessions sat at `created`.

## What was happening

`listAbandonedDraftSessionIds` reads its batch oldest-first:

```sql
SELECT id FROM sessions
WHERE status = 'created' AND updated_at < ?
ORDER BY updated_at ASC LIMIT 50
```

That ordering only drains a backlog while every visited row leaves the candidate
set. Two of the four outcomes left the row exactly as they found it, so any row
that declined *permanently* held the head and starved everything queued behind
it. The production log said so every hour, at `info`:

```
candidates:50  archived:0  not_draft:0  has_work:50  errored:0  truncated:true
```

All 50 were the oldest sessions in the deployment, holding messages while their
durable object still said `created` — a state current code cannot produce, since
`enqueuePromptCore` inserts the message and transitions in the same turn. They
would have declined forever, and nothing behind them was ever reached.

## The change

Every outcome now moves the row out of the candidate set.

| Outcome | Before | After |
| --- | --- | --- |
| `archived` | leaves `created` | unchanged |
| `not_draft` | repair could silently no-op | status-only write that always lands |
| `has_work` | **returns unchanged** | settles to `active` / `completed` |
| 404 | **throws, retried forever** | reported as `missing`; index row retired |

**`has_work` settles the session.** A queued prompt goes to `active` and is left
for the dispatch timeout. It is deliberately not archived: that would discard a
real request, and `archived` is not promptable, so the author could not resume it
either.

**404 becomes `missing`.** A 404 is definitive rather than transient — there is
no durable object session behind the row. The sweep retires the index row
directly, which is divergence-free precisely because no durable object state
exists. These rows are real; some were found while bulk-archiving.

**`not_draft` repairs without claiming activity.** Repairing through `transition`
sent the durable object's own `updated_at`, and `updateStatus` guards on
`updated_at <= ?` to keep concurrent writes ordered. D1 is the newer of the two
whenever `touchUpdatedAt` has run, so the guard matched no rows and returned
`false` — discarded, unlogged. The new `repairStatus` writes status alone, so
`updated_at` keeps meaning "last real activity" and the repair cannot be dropped.

**A stalled sweep is now loud.** `truncated` with every candidate errored is the
one shape where the next run reads identical rows. It logs at `error` instead of
hiding in an `info` line.

## Tests

Written first; the nine that mattered failed against `main` before the fix.

- `abandoned-draft-sweep.test.ts` — `missing` routes to the index archive; a
  failed orphan archive is isolated as `errored`; the stall alarm fires on a full
  errored batch **and stays quiet when a full batch was repaired rather than
  archived**, which is what encodes "repair is progress"; 404 resolves to
  `missing` while 500 still throws.
- `do-internal-routes.test.ts` — against the real durable object: a draft with a
  queued message now ends at `active`, one whose messages finished ends at
  `completed`, and neither is ever archived.
- `d1-session-index.test.ts` — against real D1: `repairStatus` writes status
  without moving `updated_at` and lands even when the index is ahead of the
  durable object (the regression the method exists for); `archiveOrphanedDraft`
  refuses a row that left `created`, and after it runs the next batch read
  returns the row *behind* the orphan — the end-to-end proof the head advances.

2572 unit, 801 integration, typecheck and lint clean.

## Not in this PR

Settling a stalled draft to `active` hands it to the pending-dispatch timeout,
which still does not exist — a prompt whose sandbox never connected has no exit
but an explicit cancel. That is the follow-up, and it has to run inside this same
endpoint, since a dormant durable object has no alarm armed to run it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved abandoned-draft cleanup by archiving sessions whose runtime is missing.
  - Drafts with messages now transition to **active**, **completed**, or **failed** based on their work state.
  - Prevented drafts with queued or processing work from being archived.
  - Repaired stale session statuses without changing session recency.
  - Improved reporting for missing sessions, cleanup failures, and stalled batches.
- **Tests**
  - Expanded coverage for orphaned drafts, status repairs, partial failures, and lifecycle outcomes.

### DIFF
--- a/packages/control-plane/src/db/session-index.ts
+++ b/packages/control-plane/src/db/session-index.ts
@@ -666,8 +666,9 @@ export class SessionIndexStore {
    *
    * `created` is the status a session holds until its first prompt is enqueued,
    * so a row still sitting there long after its last update was abandoned before
-   * any work started. Ordered oldest-first so a backlog drains deterministically
-   * across ticks rather than re-reading the same head each time.
+   * any work started. Ordered oldest-first, which drains a backlog only while
+   * every visited row leaves this set — see `archiveOrphanedDraft` and
+   * `repairStatus` for the two cases where that had to be made true.
    */
   async listAbandonedDraftSessionIds(staleBefore: number, limit: number): Promise<string[]> {
     const result = await this.db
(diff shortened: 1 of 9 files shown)


ColeMurray:
@open-inspect[bot] /fix-pr

PR #1523, opened 2026-08-19
fix(github-bot): never leave a review status stuck pending

OjasMor:
## Summary

A review writes a `pending` commit status when it starts, and only the agent's success path ever replaced it. Every other ending — a moved head, a lost ownership lease, a failed write, an agent that exits early — left that status pending forever, with nothing to clear it. A pending status is indistinguishable from a review still running, so the check never resolves and whoever is waiting on it waits forever.

This is not hypothetical. A PR in another repo has carried two pending `open-inspect` statuses on the *same* commit for over an hour:

```
pending | 17:23:12Z | Review in progress   <- a later retrigger
pending | 16:09:33Z | Review in progress   <- orphaned, nothing ever clears it
```

Two close-outs, matching the two ways a status gets orphaned:

- **The agent closes its own status** when step 5's guarded chain does not reach the status write (`prompts.ts`). Previously the instruction told it to exit without posting a review, inline comment, "or status update by any other means" — which is precisely what made the hang silent. The close-out is now the one status write permitted outside the chain, and it is required.
- **A push closes out the head it replaced** (`closeOutSupersededHeadStatus`, wired into both review-trigger handlers). That commit's review is cancelled by the supersession sweep and nothing else ever revisits its status. `synchronize` is the only action carrying `before`, so the payload schema gains it as optional; the all-zero sha GitHub sends when there is no prior head is ignored.

### Known remaining gaps

Named rather than fixed here, both deliberately:

- The agent's self-close cannot run if the agent never starts or dies hard — worker crash, sandbox death, timeout. Those endings still strand a pending status. Closing that needs a server-side reaper (a scheduled pass closing out pendings whose session row is terminal or older than some age), which is a larger change.
- `handlers.ts` posts the pending status *after* `createSession`, so a fast-failing session can have its terminal status overwritten by the later pending write.

## Testing

165 tests pass (161 before, 4 new), plus workspace `lint`, `format:check`, `test:lint-complexity`, and `typecheck`.

New coverage: the replaced head is closed out; no close-out when GitHub sends no `before`; the all-zero sha is ignored; and the review prompt requires a terminal status when the chain does not publish.

One test fix worth calling out: `webhook.test.ts` replaces the whole `github-auth` module with a mock factory, so the new named export that `prompts.ts` imports threw at import time and silently killed the review flow right after the sweep. The mock now mirrors the module's exports.

## Summary by CodeRabbit

* **New Features**
  * GitHub reviews now support reopened, synchronized, and ready-for-review events with commit-status updates and protection against stale or duplicate reviews.
  * Added operator session archiving with authorization, pagination, retry handling, and web access.
  * Slack and Linear classification now use OpenAI models with configurable model settings.
* **Bug Fixes**
  * Improved sandbox startup timeout handling and cleanup of superseded review sessions.
* **Documentation**
  * Updated setup guides with required permissions, API keys, operator access, and deployment configuration.


OjasMor:
Opened against the wrong remote by mistake — this targets our fork, not upstream. Closing; apologies for the noise.

PR #1601, opened 2026-08-24
fix(control-plane): don't stop a sandbox for inactivity mid-execution

duboff:
## Summary

The inactivity watchdog stops a sandbox while it is still executing a message, which kills the running tool call and fails the message with `Execution timed out (stuck processing)`.

`last_activity` is only refreshed by sandbox events — `tool_call`, `step_start`, `step_finish`, `artifact` (`session/sandbox-events.ts`). Heartbeats do not count. A single long tool call emits one event when it starts and one when it finishes, so anything quiet for longer than `SANDBOX_INACTIVITY_TIMEOUT_MS` (default 10 min) looks abandoned while it is actually working: a slow build, a long test suite, a poll loop waiting on CI.

With no client WebSocket attached — every automation and bot-driven session — `evaluateInactivityTimeout` returns `timeout`, the manager snapshots and stops the sandbox, and `createAlarmHandler` sees a non-`no_action` lifecycle result and calls `failStuckProcessingMessage()`. The user sees an execution-timeout error roughly 10 minutes into a turn, with `EXECUTION_TIMEOUT_MS` (default 7200s) nowhere near elapsed.

The `extend`-with-warning branch for connected clients shows the intent is to reap *abandoned* sessions. Reaping a live execution looks unintended.

### The change

- `InactivityState` gains a required `hasInFlightExecution` field. Required, not optional, so a future caller has to answer the question rather than silently inherit the old behaviour.
- When the threshold is crossed and an execution is in flight, return `schedule` at `minCheckIntervalMs` instead of `timeout`. The ordinary idle countdown resumes once the execution finishes. It takes precedence over the `extend` branch, whose "sandbox will stop in 5 minutes due to inactivity" warning is wrong mid-run.
- `handleAlarm` takes a `SandboxAlarmOptions` argument to receive it. `createAlarmHandler` already reads the processing message to evaluate the execution timeout, so the fact costs nothing extra to pass down, and the lifecycle manager stays independent of the message queue.

The execution timeout alarm remains the backstop for a run that never completes. The trade-off is that a session wedged in `processing` with a live heartbeat now holds its sandbox until that timeout rather than 10 minutes — deliberate, since the alternative is killing legitimate long-running work.

## Test plan

- [x] `evaluateInactivityTimeout` returns `schedule` when activity is stale by 3x the timeout and an execution is in flight
- [x] the in-flight check takes precedence over the connected-client extension
- [x] `handleAlarm({ hasInFlightExecution: true })` returns `no_action` and neither snapshots, stops, nor shuts down the sandbox
- [x] the alarm handler passes `hasInFlightExecution: true` when a message is processing and `false` when none is
- [x] all three new assertions fail with the guard removed (verified by reverting it)
- [x] `npx vitest run` in `packages/control-plane` — 203 files, 3171 tests pass
- [x] `npm run typecheck -w @open-inspect/control-plane`
- [x] eslint and prettier clean on the changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandboxes with executions in progress are no longer incorrectly considered idle.
  * In-flight executions are protected from inactivity-based shutdowns, even when no clients are connected.
  * Alarm handling now preserves active processing and avoids incorrectly marking it as stuck.
* **Tests**
  * Added coverage for timeout scheduling, execution completion, and connected-client behavior.

### DIFF
--- a/packages/control-plane/src/sandbox/lifecycle/manager.ts
+++ b/packages/control-plane/src/sandbox/lifecycle/manager.ts
@@ -294,10 +294,10 @@ export interface SandboxAlarmOptions {
   /**
    * Whether a message is still being executed. The caller owns message state
    * (the lifecycle manager deliberately does not depend on the message queue),
-   * so it has to be passed in. Defaults to `false` for callers with no message
-   * queue at all.
+   * so it has to be passed in. Required, with no default: a caller that could
+   * omit it would silently reintroduce stopping a sandbox mid-execution.
    */
-  hasInFlightExecution?: boolean;
+  hasInFlightExecution: boolean;
 }
 
 /**
@@ -1212,7 +1212,7 @@ export class SandboxLifecycleManager implements SandboxLifecycle {
   /**
    * Handle alarm for inactivity and heartbeat monitoring.
    */
-  async handleAlarm(options: SandboxAlarmOptions = {}): Promise<SandboxAlarmResult> {
+  async handleAlarm(options: SandboxAlarmOptions): Promise<SandboxAlarmResult> {
     const sandbox = this.storage.getSandbox();
     if (!sandbox) {
       this.log.debug("Alarm fired: no sandbox found");
@@ -1323,7 +1323,7 @@ export class SandboxLifecycleManager implements SandboxLifecycle {
       lastActivity: sandbox.last_activity,
       status: sandbox.status,
       connectedClientCount: connectedClients,
-      hasInFlightExecution: options.hasInFlightExecution ?? false,
+      hasInFlightExecution: options.hasInFlightExecution,
     };
 
     const inactivityDecision = evaluateInactivityTimeout(
--- a/packages/control-plane/src/sandbox/lifecycle/manager.ts
+++ b/packages/control-plane/src/sandbox/lifecycle/manager.ts
@@ -289,6 +289,17 @@ export type UnresponsiveSandboxTrigger =
 
 export type SandboxAlarmResult = "no_action" | "sandbox_failed" | "sandbox_terminated";
 
+/** Session state the alarm needs that the lifecycle manager cannot read itself. */
+export interface SandboxAlarmOptions {
+  /**
+   * Whether a message is still being executed. The caller owns message state
+   * (the lifecycle manager deliberately does not depend on the message queue),
+   * so it has to be passed in. Defaults to `false` for callers with no message
+   * queue at all.
+   */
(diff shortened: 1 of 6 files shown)


packages/control-plane/src/session/alarm/handler.ts
```diff
@@ -63,7 +63,11 @@ export function createAlarmHandler(deps: AlarmHandlerDeps): AlarmHandler {
         }
       }
 
-      const lifecycleResult = await deps.lifecycleManager.handleAlarm();
+      // `processing` is this session's in-flight execution: while it is set the
+      // sandbox is working, however long it has gone without emitting an event.
+      const lifecycleResult = await deps.lifecycleManager.handleAlarm({
+        hasInFlightExecution: processing !== null,
+      });
```
ColeMurray, line 72: [Automated Review] After the execution-timeout branch awaits `failStuckProcessingMessage()`, `processing` still references the row that was just failed, so this call passes `hasInFlightExecution: true` and defers otherwise-due inactivity cleanup for another `minCheckIntervalMs`. Re-read processing state after the awaited failure (so a concurrently-started replacement is also respected) and add a timed-out-path assertion for the value passed to `handleAlarm`.

packages/control-plane/src/sandbox/lifecycle/manager.ts
```diff
@@ -289,6 +289,17 @@ export type UnresponsiveSandboxTrigger =
...
 
+/** Session state the alarm needs that the lifecycle manager cannot read itself. */
+export interface SandboxAlarmOptions {
+  /**
+   * Whether a message is still being executed. The caller owns message state
+   * (the lifecycle manager deliberately does not depend on the message queue),
+   * so it has to be passed in. Defaults to `false` for callers with no message
+   * queue at all.
+   */
+  hasInFlightExecution?: boolean;
```
ColeMurray: [Automated Review] This optional field and the `handleAlarm(options = {})` fallback weaken the invariant introduced by this PR: a future caller can omit execution state and silently restore the mid-execution shutdown bug. The production alarm caller already owns this state, so make `hasInFlightExecution` required at the manager boundary and remove the `false` default.

ColeMurray:
@duboff, i implemented the intent behind this change using a bit different approach. the system makes use of the heartbeat activities for this and keeps the inactivity timer alive via the heartbeats if the message is still processing.

a bit simpler and avoids introducing the boolean to pass through the layers

https://github.com/ColeMurray/background-agents/pull/1611

PR #1621, opened 2026-08-26
fix(slack): keep activity status alive during long tool calls

### LINKED ISSUE #1262: Slack activity status expires during long-running tool calls
## Description

The Slack agent activity indicator disappears while an agent is still working on a long-running tool call.

Slack automatically clears `assistant.threads.setStatus` after two minutes without a new message or status update. Open-Inspect currently updates the status when a tool call starts but does not refresh it while that tool remains active.

This commonly occurs when an agent:

1. Starts a long-running process, such as generating embeddings.
2. Runs a long `sleep` command while waiting for that process.
3. Produces no further tool-call events for more than two minutes.

The agent and sandbox remain active, but Slack stops showing the working status.

## Steps to reproduce

1. Start an agent session from Slack.
2. Have the agent execute a command that takes longer than two minutes:

   ```bash
   sleep 600
   ```

3. Observe the `Working...` indicator in Slack.
4. Wait for two minutes without another tool call or agent response.

## Actual behavior

The Slack activity indicator disappears after approximately two minutes, even though the agent is still processing the request.

## Expected behavior

The activity indicator should remain visible for the duration of the active agent execution and clear when the execution completes, fails, or is cancelled.

## Cause

Tool-call notifications are delivered once per call ID and do not provide a periodic refresh:

- [`CallbackNotificationService.notifyToolCall`](https://github.com/ColeMurray/background-agents/blob/68606602d2ddd54979e8e5a7b72e716d93fd1008/packages/control-plane/src/session/callback-notification-service.ts#L322-L345)
- [`setAssistantThreadStatusBestEffort`](https://github.com/ColeMurray/background-agents/blob/68606602d2ddd54979e8e5a7b72e716d93fd1008/packages/slack-bot/src/activity-status.ts#L227-L262)

Slack documents a two-minute timeout for this status:

https://docs.slack.dev/reference/methods/assistant.threads.setStatus/

## Suggested direction

Refresh the Slack status periodically while a Slack-originated message remains in the processing state, at an interval safely below two minutes, such as every 60–90 seconds.

The refresh should stop when the execution completes, fails, or is cancelled.

hydraxman:
## Summary
- refresh Slack assistant-thread activity every 90 seconds while a Slack message is still processing
- use the persisted Durable Object alarm and a signed callback so refreshes survive runtime eviction without affecting other message sources
- stop re-arming automatically once the processing message completes, fails, or is cancelled

## Problem
Slack expires `assistant.threads.setStatus` after two minutes without an update. A quiet tool call longer than that leaves the agent running while its visible working status disappears.

## Testing
- `npm test -w @open-inspect/control-plane` — 3,194 passed
- `npm test -w @open-inspect/slack-bot` — 422 passed
- `npm run typecheck -w @open-inspect/control-plane` — passed
- `npm run typecheck -w @open-inspect/slack-bot` — passed
- `npm run lint -w @open-inspect/control-plane` — passed
- `npm run lint -w @open-inspect/slack-bot` — passed
- `npm run build -w @open-inspect/control-plane` — passed
- `npm run build -w @open-inspect/slack-bot` — passed

Fixes #1262

## Summary by CodeRabbit

* **New Features**
  * Slack assistant threads now remain marked as “Working…” while responses are processing.
  * Added automatic activity heartbeats approximately every 90 seconds.
  * Added signed callback handling to securely refresh Slack activity status.

* **Bug Fixes**
  * Activity refreshes continue through transient delivery failures.
  * Non-Slack and completed activities are excluded from unnecessary refreshes.
  * Invalid, expired, or future-dated callback requests are rejected.

* **Telemetry**
  * Added dedicated tracking for assistant activity heartbeat updates.

### DIFF
--- a/packages/control-plane/src/session/alarm/handler.ts
+++ b/packages/control-plane/src/session/alarm/handler.ts
@@ -55,6 +55,7 @@ export function createAlarmHandler(deps: AlarmHandlerDeps): AlarmHandler {
       // already failed (by lifecycle recovery or a prior alarm),
       // getProcessingMessageWithStartedAt() returns null.
       const processing = deps.repository.getProcessingMessageWithStartedAt();
+      let activityMessageId: string | null = null;
       if (processing?.started_at) {
         const now = deps.now();
         const executionTimeoutMs = deps.getExecutionTimeoutMs();
@@ -73,16 +74,11 @@ export function createAlarmHandler(deps: AlarmHandlerDeps): AlarmHandler {
           await deps.messageQueue.failStuckProcessingMessage();
         } else {
           // An earlier lifecycle or activity alarm has consumed the Durable
-          // Object's single alarm slot. Reassert the execution deadline first,
-          // then schedule another Slack refresh only when this message owns a
-          // valid Slack callback.
+          // Object's single alarm slot. Reassert the execution deadline first;
+          // a non-terminal lifecycle result below may then schedule another
+          // Slack refresh when this message owns a valid callback.
           await deps.alarmScheduler.schedule(processing.started_at + executionTimeoutMs);
-          const continueActivityHeartbeat = await deps.callbackService.notifyActivityHeartbeat(
-            processing.id
-          );
-          if (continueActivityHeartbeat) {
-            await deps.alarmScheduler.schedule(now + SLACK_ACTIVITY_HEARTBEAT_INTERVAL_MS);
-          }
+          activityMessageId = processing.id;
         }
       }
 
@@ -92,6 +88,13 @@ export function createAlarmHandler(deps: AlarmHandlerDeps): AlarmHandler {
       }
       if (lifecycleResult === "sandbox_terminated") {
         await deps.executionStop.resumeAfterSandboxTermination();
+      }
+      if (lifecycleResult === "no_action" && activityMessageId) {
+        const continueActivityHeartbeat =
+          await deps.callbackService.notifyActivityHeartbeat(activityMessageId);
+        if (continueActivityHeartbeat) {
+          await deps.alarmScheduler.schedule(deps.now() + SLACK_ACTIVITY_HEARTBEAT_INTERVAL_MS);
+        }
       }
       if (projectionFailure) throw projectionFailure.error;
     },
--- a/packages/control-plane/src/session/alarm/handler.ts
+++ b/packages/control-plane/src/session/alarm/handler.ts
@@ -92,7 +92,8 @@ export function createAlarmHandler(deps: AlarmHandlerDeps): AlarmHandler {
       if (lifecycleResult === "no_action" && activityMessageId) {
         const continueActivityHeartbeat =
           await deps.callbackService.notifyActivityHeartbeat(activityMessageId);
-        if (continueActivityHeartbeat) {
+        const stillProcessing = deps.repository.getProcessingMessageWithStartedAt();
+        if (continueActivityHeartbeat && stillProcessing?.id === activityMessageId) {
           await deps.alarmScheduler.schedule(deps.now() + SLACK_ACTIVITY_HEARTBEAT_INTERVAL_MS);
         }
       }
--- a/packages/control-plane/src/session/alarm/handler.ts
+++ b/packages/control-plane/src/session/alarm/handler.ts
@@ -6,6 +6,8 @@ import type { SessionMessageQueue } from "../message-queue";
 import type { ExecutionStopCoordinator } from "../execution-stop-coordinator";
 import type { MessageRepository } from "../message-repository";
 import type { SessionTerminalMessageProjection } from "../terminal-message-projection";
+import type { CallbackNotificationService } from "../callback-notification-service";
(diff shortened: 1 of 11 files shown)


packages/control-plane/src/session/alarm/handler.ts
```diff
@@ -58,10 +61,17 @@ export function createAlarmHandler(deps: AlarmHandlerDeps): AlarmHandler {
...
+          // An earlier lifecycle or activity alarm has consumed the Durable
+          // Object's single alarm slot. Reassert the execution deadline first,
+          // then schedule another Slack refresh only when this message owns a
+          // valid Slack callback.
           await deps.alarmScheduler.schedule(processing.started_at + executionTimeoutMs);
+          const continueActivityHeartbeat = await deps.callbackService.notifyActivityHeartbeat(
+            processing.id
+          );
+          if (continueActivityHeartbeat) {
+            await deps.alarmScheduler.schedule(now + SLACK_ACTIVITY_HEARTBEAT_INTERVAL_MS);
```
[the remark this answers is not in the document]
hydraxman: Fixed in bddee490. The handler now evaluates the lifecycle result before sending or re-arming activity and only refreshes on `no_action`. A regression test covers a processing message whose alarm returns `sandbox_failed`; the callback is not invoked and the message is failed. Focused alarm tests pass (9/9), and the full control-plane suite passes (3,195 tests).

packages/control-plane/src/session/callback-notification-service.ts
```diff
@@ -333,6 +333,90 @@ export class CallbackNotificationService {
...
+    try {
+      sessionId = this.getSessionId();
+      const callbackData = {
+        sessionId,
+        messageId,
+        timestamp: Date.now(),
+        context,
+      };
+      const signature = await this.signPayload(callbackData, secret);
+      const response = await binding.fetch("https://internal/callbacks/activity", {
```
ColeMurray: [Automated Review] This unbounded service-binding fetch runs inside the shared Durable Object alarm path. If it never settles, `createAlarmHandler` never reaches the re-arm in `handler.ts`, so the current alarm invocation can also prevent later execution-timeout and lifecycle work from running. Please bound the attempt with an `AbortSignal` (or reuse the existing 10-second `deliverWithRetry` policy), still allow the next heartbeat to be scheduled after a timeout, and add a deferred fetch test proving the alarm completes when the binding never resolves.

packages/control-plane/src/session/alarm/handler.ts
```diff
@@ -72,6 +78,13 @@ export function createAlarmHandler(deps: AlarmHandlerDeps): AlarmHandler {
       if (lifecycleResult === "sandbox_terminated") {
         await deps.messageQueue.resumeAfterSandboxTermination();
       }
+      if (lifecycleResult === "no_action" && activityMessageId) {
```
ColeMurray, line 92: [Automated Review] This branch relies on `activityMessageId` captured before several awaits. A completion event can interleave, mark that message terminal, and this path will still send a heartbeat and schedule another because neither `notifyActivityHeartbeat` nor the re-arm revalidates `processing`. Since `/callbacks/activity` returns before its `waitUntil(setStatus)` work finishes, the late update can also restore `Working...` after the completion reply cleared it. Please re-read that the same message remains processing immediately before delivery and again before re-arming, with a deferred completion-interleaving test.

packages/control-plane/src/session/callback-notification-service.ts
hydraxman: Fixed in 853190ae. Activity delivery now uses the existing bounded callback retry path with an AbortSignal on each attempt, and the regression test holds the service binding open until both attempts time out while proving the alarm path settles.

packages/control-plane/src/session/alarm/handler.ts
hydraxman: Fixed in 853190ae. The heartbeat now verifies that the same message is still processing immediately before delivery, and the alarm handler verifies it again after delivery before re-arming. A deferred-delivery regression completes the message mid-callback and proves no heartbeat alarm is scheduled afterward.

packages/slack-bot/src/callbacks.ts
```diff
@@ -39,6 +43,14 @@ const completionCallbackSchema = z.looseObject({
   context: slackCallbackContextSchema,
 });
 
+const activityCallbackSchema = z.looseObject({
+  sessionId: z.string(),
+  messageId: z.string(),
+  timestamp: z.number(),
```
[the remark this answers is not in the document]
hydraxman: Fixed in 8848247b. The activity route now verifies the signed heartbeat timestamp is within a five-minute clock-skew window before scheduling any Slack status update, rejecting stale and far-future callbacks with 401. Deterministic RED/GREEN coverage proves both directions are rejected without calling Slack; the full Slack bot suite passes (436 tests), along with workspace typecheck and formatting.

ColeMurray:
thanks! superseded by: https://github.com/ColeMurray/background-agents/pull/1867

PR #1657, opened 2026-08-28
perf(control-plane): put automation instructions before the event context block

24601:
## Summary

- Automation prompts put the per-event context block before the instructions, so the static half of the prompt never sits at a stable offset.
- Routes both composition sites through one helper and puts instructions first by default.
- `AUTOMATION_INSTRUCTIONS_FIRST="false"` restores the previous ordering.

## Problem

`scheduler.ts` composed event-triggered automation prompts like this:

```js
`${event.contextBlock}\n---\n\n${automation.instructions}`
```

`contextBlock` is described in `shared/src/triggers/types.ts` as "Human-readable context prepended to automation instructions", so the ordering is intended rather than accidental.

An automation's `instructions` are the same on every run. The event context differs each time — PR number, actor, payload. Provider prompt caches match byte-identical leading spans, so a variable prefix keeps the instructions off the cache boundary and every run pays full input price for them.

The instruction bodies are large enough for this to matter. Three automations in my deployment carry 5.5k, 11k and 15k characters of instructions, re-sent uncached on every event.

Composed prompt head, two events for the same automation:

```
before                          after
──────────────────────────      ──────────────────────────
Slack thread #eng-1204...       You are the review agent...
---                             ---
You are the review agent...     Slack thread #eng-1204...

Slack thread #eng-1207...       You are the review agent...
---                             ---
You are the review agent...     Slack thread #eng-1207...
```

The shared leading span is empty on the left and the whole instruction body on the right.

## Change

`composeAutomationPrompt(env, contextBlock, instructions)` is now the single composition point for both paths: the plain event path and the Slack path that enriches the context block with thread history. The second one built its prompt inline and is easy to miss when grepping for the variable name.

The separator is unchanged, and the opt-out branch is character-identical to the previous line.

## Testing

```
npm run lint                                  exit 0
npm run typecheck                             clean
npm test -w @open-inspect/control-plane       3334 passed
```

The flag is declared as `automation_instructions_first` (bool, default `true`) in `terraform/environments/production/variables.tf` and bound as a plain-text worker var with `tostring()`, following `UNSAFE_ALLOW_ALL_USERS`.

Four new tests in `compose-automation-prompt.test.ts` cover the default, the opt-out, unrecognised flag values falling back to the default, and the property the change exists for: two events with different context produce the same leading span. Three existing Slack assertions in `scheduler.test.ts` encoded the old order and now assert the new one.

These tests exercise the helper directly. They do not cover the scheduler path end to end, so they would stay green if `startInvocation` stopped passing `this.env`, and they do not verify the production binding. Building that seam looked heavier than this change warrants.

## What I have not measured

I verified the ordering and the composed output. I have not measured a cache-hit rate or a token-cost delta, because my sessions route through a gateway that does not report per-request cache hits, I will likely attempt to measure this at some point. Until then, the reasoning above is about how prefix caching works, not an observed saving. If you have a deployment that reports cache hits, that number would be worth having before this merges.

I also haven't formally been able to verify the impact this ordering change has on quality/performance/output from the LLM/Agents. Anecdotally, it has been transparent to me, but if you have eval baselines you use we can run this against, that's really going to be the proper way to measure that. 

## One thing this does not reorder

On the Slack path, `appendSlackSessionInstructions` appends workspace-level instructions after composition, so that static text still trails the dynamic context block. Moving it would change that function's contract, which felt out of scope here. Worth a follow-up if the ordering matters to you.

## Draft

Opened as a draft. Happy to rename the flag, move it to a per-automation column instead of a deployment env var, or drop the flag and swap the order outright if you would rather not carry the configuration.

## Summary by CodeRabbit

* **New Features**
  * Event-triggered automations now place instructions before event context by default, improving prompt clarity.
  * Added a configuration option to restore context-first ordering when needed.
  * Slack-based automation runs now follow the same prompt ordering behavior.

* **Tests**
  * Added coverage for default and configurable prompt ordering across event and Slack contexts.

### DIFF
--- a/packages/control-plane/src/scheduler/compose-automation-prompt.test.ts
+++ b/packages/control-plane/src/scheduler/compose-automation-prompt.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from "vitest";
+import { composeAutomationPrompt } from "./scheduler";
+import type { Env } from "../types";
+
+const env = (value?: string) => ({ AUTOMATION_INSTRUCTIONS_FIRST: value }) as unknown as Env;
+
+describe("composeAutomationPrompt", () => {
+  it("puts instructions first by default", () => {
+    expect(composeAutomationPrompt(env(undefined), "CTX", "INSTRUCTIONS")).toBe(
+      "INSTRUCTIONS\n---\n\nCTX"
+    );
+  });
+
+  it("keeps the same leading span when only the context changes", () => {
+    const a = composeAutomationPrompt(env(undefined), "event one", "INSTRUCTIONS");
+    const b = composeAutomationPrompt(env(undefined), "event two", "INSTRUCTIONS");
+    const shared = "INSTRUCTIONS\n---\n\n".length;
+    expect(a.slice(0, shared)).toBe(b.slice(0, shared));
+  });
+
+  it('restores context-first ordering when the flag is "false"', () => {
+    expect(composeAutomationPrompt(env("false"), "CTX", "INSTRUCTIONS")).toBe(
+      "CTX\n---\n\nINSTRUCTIONS"
+    );
+  });
+
+  it("treats any other value as the default", () => {
+    expect(composeAutomationPrompt(env("true"), "CTX", "INSTRUCTIONS")).toBe(
+      "INSTRUCTIONS\n---\n\nCTX"
+    );
+    expect(composeAutomationPrompt(env(""), "CTX", "INSTRUCTIONS")).toBe(
+      "INSTRUCTIONS\n---\n\nCTX"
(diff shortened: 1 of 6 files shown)


24601:
Oh, @coderabbit, "slop"? Them be fightin' words, clanker! 

Yeah, I used claude code, bro, what's it to you, code bunny? 

j/k

ColeMurray:
this looks good. i think we can remove the flag and have this be the default as from a prompt caching perspective it makes more sense.

ColeMurray:
thanks!

superseded by https://github.com/ColeMurray/background-agents/pull/1667

PR #1685, opened 2026-08-31
Close out a review whose session died before it could

OjasMor:
## Summary

The github-bot writes a `pending` commit status when a review starts; the agent writes the terminal one when it finishes. A session killed by a supervisor timeout runs no code of its own, so it writes **neither** — and `pending` is indistinguishable from a review still running, so the check never resolves and any merge rule keyed on it waits forever.

Observed today on AITaaS #2181: session `151ee214…` ended at 13:33:34 with `"Execution timed out (stuck processing)"`, and the `open-inspect` status was still "Review in progress" at 17:05 — three and a half hours later. open-inspect itself was healthy throughout; it reviewed #2194, #2193, #2190, #2188 and #2182 successfully in the same window. A bot-level liveness check would have been green through the whole incident, which is why this is fixed per-review rather than per-bot.

Every existing close-out is agent-driven — `prompts.ts` instructs the agent to post one before it exits — which is precisely what a killed sandbox cannot do. This closes out from the control plane instead, which already knows the session went terminal, already holds the GitHub App credentials, and already runs a per-minute reaper over the same table. **No new service binding, secret, cron trigger or migration.**

## What it does

`closeOutDeadReviewSessions` runs from the existing minute cron, after `reapSupersededReviewSessions`, and posts `state: error` / `context: open-inspect` / `"Review did not run — push again to retry"` against the row's `head_sha`, then drops the row.

Two conditions, because a session fails in two ways that look different from outside:

- its session status is `failed` or `cancelled`, or
- the row sat past `REVIEW_STALE_PENDING_MS` (20 min) without reaching either.

The second is the case the first cannot cover: a runtime that dies before recording anything leaves a row no status transition will ever resolve. Twenty minutes is set against measured healthy reviews, which ran 100–620 s of active time on 2026-08-31.

## Boundaries

- **Superseded rows stay the reaper's job.** It cancels the session first, and doing both in one tick would post twice for one head — hence the `generation = latest_generation` predicate and the ordering in the cron.
- **An unexpired submission lease defers**, for the same reason `cancelStaleReviewSession` defers: the holder is mid-GitHub-write and a status posted now would race the verdict it is about to publish. The row is retried once the lease lapses.
- **A failed write retains the row** so the next tick retries. A 422 does not — a force-push left no commit to describe.
- **No-ops entirely** when the GitHub App is unconfigured.

`REVIEW_STATUS_CONTEXT` moves to `@open-inspect/shared`. Two workers now write this status, and a copy in each package is a string that can drift silently — the symptom being a PR carrying two `open-inspect` statuses with one of them pending forever. github-bot re-exports it, so its existing importers are unchanged.

## Testing

Six new cases in `github-reviews.test.ts` covering both conditions, the lease deferral, the retained-on-failure path, the 422 path, and the unconfigured no-op.

`lint`, `format:check` and `typecheck` are clean. Backend suites pass (control-plane 3470 tests). The 38 failures in `packages/web` are pre-existing and unrelated — verified by running them on a clean `origin/main` tree, where they fail identically.

Filed as COD-696.

## Summary by CodeRabbit

* **New Features**
  * Added personal access tokens with creation, expiration, listing, one-time display, and revocation.
  * Added a read-only MCP server for inspecting sessions and automation runs.
  * Expanded GitHub reviews to support reopened, synchronized, and ready-for-review pull requests, with commit statuses and supersession handling.
  * Added operator session archiving with paginated processing.
  * Added an onboarding guide for deploying Open-Inspect instances.
* **Bug Fixes**
  * Improved OAuth credential cleanup and OpenAI subscription fallback behavior.
* **Documentation**
  * Updated setup, permissions, secrets, model, GitHub review, and MCP documentation.


OjasMor:
Opened against the wrong fork by mistake — this belongs on opencodos/background-agents. Closing; no review needed.

PR #1804, opened 2026-09-06
ci: support Actions variables for non-secret deployment config

csaroff:
## Summary

Allow existing non-secret deployment settings to live in GitHub Actions **Variables**, while keeping credentials in **Secrets** and preserving existing secret-only deployments.

Today most deployment settings—including account IDs, feature flags, provider configuration, and branding—must be stored as secrets. That makes ordinary configuration write-only in the GitHub UI and masks it in deployment logs. The workflows already use `vars.CLASSIFICATION_MODEL`; this extends that distinction to the other existing non-secret settings.

## Changes

- Resolve 45 existing configuration names as `vars.NAME || secrets.NAME`, preserving each existing default where present.
- Apply the same resolution to Terraform plan and apply, R2 backend account selection, and the Vercel deployment gate/environment.
- Pass configuration consumed by shell commands through environment bindings rather than interpolating variable values into shell source.
- Document which settings can use Variables, how to bulk-upload them with `gh variable set`, and how fallback/migration works.

## Compatibility and scope

- Existing secret-only deployments require no changes.
- A non-empty variable takes precedence over the same-named secret. An empty variable falls back to the secret; it does not clear it.
- `false` and `0` are non-empty strings in Actions variables and retain their intended values.
- API credentials, OAuth client secrets, signing/private/encryption keys, R2 credentials, and both Modal token fields remain secret-only.
- Allowlists may still be kept in secrets when operators prefer masking personal information in logs.
- `CLASSIFICATION_MODEL` retains its existing variable-only behavior.
- No new Terraform inputs, provider behavior, branding assets, or sandbox image changes.

## Validation

- [x] `actionlint -shellcheck= .github/workflows/terraform.yml .github/workflows/deploy-web.yml`
- [x] Prettier check for both workflows and both documentation files
- [x] `git diff --check`
- [x] Local verification using `@actions/expressions`: 744 checks across all updated expression occurrences, covering variable precedence, secret fallback, empty/missing values, defaults, `false`, `0`, whitespace, and shell-like strings
- [x] Parsed-YAML checks for plan/apply environment parity, documented variable coverage, and credential isolation
- [x] Executed the Vercel gate with six configuration cases and both R2 initialization commands against a stub Terraform executable

The expression and shell checks were run from a temporary validation harness; no new runtime dependencies were added. Full application suites and a live deployment were not run for this workflow/documentation-only PR. GitHub CI will provide the repository checks.

Compare: https://github.com/ColeMurray/background-agents/compare/main...csaroff:background-agents:feat-actions-config-variables?expand=1

## Summary by CodeRabbit

- **Documentation**
  - Updated CI/CD setup guidance to distinguish non-sensitive configuration variables from credential secrets.
  - Documented configuration precedence: repository or organization variable, same-named secret, then existing default.
  - Added instructions for bulk-uploading variables with the GitHub CLI.
  - Clarified that secret-only deployments remain supported.
  - Updated Terraform documentation with the new Actions Secrets and Variables configuration model.

### DIFF
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -13,8 +13,8 @@ permissions:
   contents: read
 
 env:
-  VERCEL_ORG_ID: ${{ secrets.VERCEL_TEAM_ID }}
-  VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+  VERCEL_ORG_ID: ${{ vars.VERCEL_TEAM_ID || secrets.VERCEL_TEAM_ID }}
+  VERCEL_PROJECT_ID: ${{ vars.VERCEL_PROJECT_ID || secrets.VERCEL_PROJECT_ID }}
 
 jobs:
   deploy:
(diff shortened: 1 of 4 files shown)


ColeMurray:
thanks!

PR #1806, opened 2026-09-06
fix: retry webhook deliveries that failed to enqueue Autofix

benandbalto:
When `AUTOFIX_QUEUE.send` fails, the webhook previously returned 200 and finalized its delivery marker despite never enqueueing the revision. Return 503 and clear the receipt so the sender can retry. If KV cleanup also fails, keep the processing lease retryable (503) until it expires, instead of acknowledging it as completed.

Successfully processed duplicates still return 200. Other handling is intentionally deferred on an enqueue failure so the retry can process the complete delivery.

Validation: all 147 GitHub-bot tests pass, including queue failure with successful and failed KV cleanup. Independently reviewed with Claude Sonnet 5.

## Summary by CodeRabbit

- **Bug Fixes**
  - Webhook deliveries that are still being processed now return a retryable error instead of being acknowledged as duplicates.
  - Failed autofix queueing now allows the same delivery to be retried safely.
  - Failed deliveries return HTTP 503 and avoid forwarding incomplete events.

- **Tests**
  - Added coverage for queueing failures, retry behavior, and cleanup failures.

### DIFF
--- a/packages/github-bot/src/index.ts
+++ b/packages/github-bot/src/index.ts
@@ -73,6 +73,11 @@ app.post("/webhooks/github", async (c) => {
         event_type: event,
         dedupe_status: existing,
       });
+      // A processing lease is not proof of durable delivery. Retry after it
+      // completes or expires, including when failed-delivery cleanup lost KV.
+      if (existing === DELIVERY_STATUS_PROCESSING) {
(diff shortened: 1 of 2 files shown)


ColeMurray (changes requested):
-

Issue #1363, opened 2026-08-10
Sandbox connecting timeout permanently strands the session; bot-triggered reviews are silently lost


rhlsthrm:
## Summary

When a sandbox exceeds the 120 s connecting timeout, the session is left permanently stranded: the
sandbox is marked `failed` and its access state is cleared, but the queued prompt is never
delivered, the session row stays `active` with `message_count = 0` forever, and recovery is deferred
to "your next message" — which never arrives for a bot-triggered session.

For github-bot review sessions this means the review is silently never posted. Nothing is written
back to the PR, so from the outside it is indistinguishable from the bot never having been
triggered.

In one production install this accounts for **20 of 135 github-bot sessions (~15%)**.

## Observed behavior

Captured live via `wrangler tail` on the control plane while reproducing on demand (an
`@bot` mention comment on a PR):

```
sandbox.connecting_timeout   session_id=86b7cfa0…  elapsed_ms=124912  timeout_ms=120000
```

The boot exceeded the limit by **4.9 seconds**. Immediately after, the sandbox — which did come
up — is locked out of the control plane and retries in a loop:

```
POST /internal/verify-sandbox-token        → "Sandbox token verification failed: sandbox is dead"
POST /sessions/{id}/commit-signing         → auth.sandbox_failed
```

8 such rounds over the next 3.5 minutes, all from the provider's IP. The sandbox is alive and
burning compute while being refused.

The session row never moves again:

| field | value |
| --- | --- |
| `status` | `active` |
| `message_count` | `0` |
| `created_at` | 22:24:24Z |
| `updated_at` | 22:24:25Z (frozen at create + 1 s) |

Every stranded session in this install shows exactly this signature: `message_count = 0` and
`updated_at` within 0–2 s of `created_at`.

## Root cause

`packages/control-plane/src/sandbox/lifecycle/manager.ts:1133-1157` — on connecting timeout the
manager marks the sandbox `failed`, calls `clearSandboxAccessState()` (which is why the live
sandbox's token stops verifying), attempts `stopProviderSandbox("connecting_timeout")` **only if**
`canStopProviderSandbox()`, and broadcasts:

> "Sandbox failed to connect within the allowed time. It will be retried on your next message."

Three problems compound:

1. **The queued prompt is dropped, not retried.** Nothing re-drives the pending message after the
   sandbox is failed, so `message_count` stays 0.
2. **"Your next message" never comes for a bot session.** github-bot calls `sendPrompt` exactly once
[...]

PR #1821, opened 2026-09-07
fix(control-plane): widen the sandbox connect watchdog and tie the spawn staleness bound to it

rhlsthrm:
Two related reliability defects in sandbox spawn/connect handling.

## 1. The initial-connect watchdog was too tight

`DEFAULT_CONNECTING_TIMEOUT_CONFIG.timeoutMs` gave a sandbox two minutes to report connected. The boot sequence (git clone, setup.sh, start.sh, opencode, bridge connect) is 30-90 seconds for a small repo, but a large repo with a real setup script lands right at that limit.

Overrunning it is not a soft failure. `handleAlarm` marks the sandbox failed and calls `clearSandboxAccessState`, which locks out the sandbox that does eventually come up; the queued prompt is never re-driven; and the documented recovery ("it will be retried on your next message") cannot fire for bot-triggered sessions, which only ever send one prompt. So a boot that overran by a few seconds stranded its session permanently and lost the review with no signal on the PR.

This widens the bound to four minutes so slow-but-healthy boots have real headroom. It is a mitigation of the symptom, not a fix for the underlying recovery gap, which is #1363 - a session that trips the watchdog should be recoverable rather than terminal, and that is a bigger change than this.

## 2. The staleness bound and the watchdog could disagree

`DEFAULT_SPAWN_CONFIG.spawningTimeoutMs` decides when a sandbox sitting in `spawning`/`connecting` is stale enough that `evaluateSpawnDecision` stops skipping and lets a replacement spawn. Its comment said it matched the connecting-timeout watchdog, but the value was restated independently as a second literal.

That made the two silently separable, and widening the watchdog in (1) alone would have opened a two-minute window in which a healthy sandbox still inside the watchdog window is judged dead and a second sandbox is spawned alongside it - a duplicate provider sandbox per slow boot.

Both now derive from a single `CONNECT_WATCHDOG_MS`, so the invariant holds by construction rather than by comment, and the rationale for the value lives in one place.

## Why this is generally useful

Neither part is deployment-specific: no new configuration, no new environment variable, no provider-specific behavior. The first is a defaults change that only affects how much slack a slow boot gets before it is declared dead; the second removes a duplicated constant whose two copies had already drifted in meaning, which is the kind of coupling that reappears the next time either number is touched.

## Tests

`evaluateConnectingTimeout` tests that hardcoded offsets against the old two-minute value now derive them from `config.timeoutMs`, and the same in `manager.test.ts`, so they assert the behavior rather than the constant. Two new tests in `decisions.test.ts` (`connect watchdog and spawn staleness defaults`) exercise the shipped defaults together: a sandbox one millisecond inside the watchdog window must not get a replacement spawn, and once the watchdog has failed it, it must. Falsified by reintroducing the old independent literal `spawningTimeoutMs: 120_000`, which fails the first of the two (`expected 'spawn' to be 'skip'`); restored, both pass.

## Gates run locally

Run in a clean worktree off `main` at `265a5997`:

- `npm run build -w @open-inspect/shared` - exit 0
- `npm run typecheck -w @open-inspect/control-plane` (all four TS projects: default, `tsconfig.node.json`, `tsconfig.test.json`, `test/integration`) - exit 0
- `npm test -w @open-inspect/control-plane` - 275 test files passed, 4011 tests passed, exit 0
- `npm run test:integration -w @open-inspect/control-plane` - 102 test files passed, 1186 tests passed, 1 skipped, exit 0

No other package is touched, so package-scoped gates only.

## Adaptation to current main

The change was originally written against an older revision of this file. One test it edited, `calls onSandboxTerminating callback on connecting timeout` in `manager.test.ts`, no longer exists on `main` - that area is now the `terminateUnresponsiveSandbox` describe block - so only the surviving `detects connecting timeout and sets failed` case needed the derived offset. Nothing upstream was reverted and no upstream test or assertion was removed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Extended the sandbox connection and startup watchdog window to four minutes.
  * Sandboxes still within the connection window are now skipped, while replacements are spawned after the watchdog expires.
  * Updated timeout handling to use the configured watchdog duration consistently.

### DIFF
--- a/packages/control-plane/src/sandbox/lifecycle/decisions.test.ts
+++ b/packages/control-plane/src/sandbox/lifecycle/decisions.test.ts
@@ -17,6 +17,7 @@ import {
   isSandboxReconnectBlockedStatus,
   isSnapshotRuntimeCompatible,
   DEFAULT_CONNECTING_TIMEOUT_CONFIG,
+  DEFAULT_SPAWN_CONFIG,
   DEFAULT_EXECUTION_TIMEOUT_MS,
   type CircuitBreakerState,
   type CircuitBreakerConfig,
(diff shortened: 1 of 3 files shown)


ColeMurray:
thanks!

PR #1822, opened 2026-09-07
ci: thread declared-but-unwired Terraform variables through plan and apply

rhlsthrm:
## What this changes

`.github/workflows/terraform.yml` sets `TF_VAR_*` for most inputs of
`terraform/environments/production`, but eight variables that are **already declared** in
`variables.tf` were never passed. A CI-driven deployment therefore used their Terraform defaults
with no way to override them, from either an Actions variable or a secret.

This threads all eight through **both** the `plan` and the `apply` job, and documents the
corresponding Actions names in `terraform/README.md` and `docs/GETTING_STARTED.md`.

No Terraform was changed. Every variable, its validation, its consumers and its
`terraform.tfvars.example` entry already exist on `main` — only the workflow wiring was missing.

## Why this matters

The load-bearing one is `enable_service_bindings`. The control plane binds the Slack and Linear bot
workers, and each bot binds the control plane back, so a service binding cannot be created in the
same apply as the worker it points at. `terraform/README.md` already tells operators the first
deployment is two-phase: apply once with `enable_durable_object_bindings = false` and
`enable_service_bindings = false`, then apply again with both `true`. `enable_durable_object_bindings`
is wired into the workflow; `enable_service_bindings` was not. So a first deployment could not be
completed from CI at all — the documented procedure was only reachable by running Terraform by hand
or by editing the workflow.

`unsafe_allow_all_users` is the same shape of problem: an intentionally open deployment cannot
satisfy the `checks.tf` access-control gate from CI without it, because every allowlist is empty and
the escape hatch was unreachable.

The remaining six are ordinary configuration that simply had no CI path:
`cloudflare_zone_id`, `cloudflare_custom_domain`, `r2_media_bucket_name`, `r2_media_location`,
`sandbox_inactivity_timeout_ms`, `linear_api_key`.

## Nothing changes for existing deployments

Each entry uses the current `${{ vars.X || secrets.X || 'default' }}` precedence, and **every added
fallback literal is exactly that variable's own Terraform default**. That is the entire safety
argument for the diff, so I verified it mechanically rather than by eye: a script parsed the
`default =` line of each `variable` block out of `variables.tf`, parsed every `TF_VAR_*` line out of
the workflow, and compared them. Result — all eight appear exactly twice (once per job) with
byte-identical expressions, and each fallback matches its declared default:

| Variable | CI fallback | Terraform default |
| --- | --- | --- |
| `enable_service_bindings` | `'true'` | `true` |
| `unsafe_allow_all_users` | `'false'` | `false` |
| `r2_media_location` | `'ENAM'` | `"ENAM"` |
| `sandbox_inactivity_timeout_ms` | `'600000'` | `600000` |
| `r2_media_bucket_name` | none (empty string) | `""` |
| `linear_api_key` | none (empty string) | `""` |
| `cloudflare_zone_id` | none (empty string) | `null` |
| `cloudflare_custom_domain` | none (empty string) | `null` |

`linear_api_key` is threaded from `secrets` only, matching the other Linear credentials.

### The two `null` defaults

`cloudflare_zone_id` and `cloudflare_custom_domain` default to `null` in Terraform, and an unset
Actions variable/secret renders as an empty string, so those two are the only entries where the CI
value is not literally identical to the default. They are still behaviourally equivalent, which a
reviewer cannot derive from the diff alone:

- `locals.tf:35-36` normalizes both cases to the same value —
  `web_custom_domain = var.cloudflare_custom_domain == null ? "" : trimspace(var.cloudflare_custom_domain)`
  and the same shape for `web_custom_domain_zone_id`.
- Every consumer reads only those normalized locals. In particular the custom-domain gate in
  `checks.tf` is written against `local.web_custom_domain`/`local.web_custom_domain_enabled`, not the
  raw variables, so an empty string takes the same "unset" branch as `null` and the gate does not
  fire.
- `cloudflare_custom_domain`'s own validation already admits `trimspace(...) == ""`.

## Deliberately not threaded

- `project_root` — a path into the checked-out repository. CI must use the default; making it
  settable would only allow it to be set wrong.
- `control_plane_migration_tag`, `control_plane_migration_old_tag`,
  `control_plane_new_sqlite_classes`, `control_plane_deleted_classes` — staged by hand for an
  explicit Durable Object migration, as `terraform.tfvars.example` says. Two are `list(string)`,
  where the empty string an unset Actions variable renders would not be a valid list and would break
  every deploy; `control_plane_migration_old_tag` defaults to `null`, which is *not*
  empty-string-equivalent for its consumers.

## Verification

Run in `terraform/environments/production`:

- `terraform init -backend=false` — clean.
- `terraform validate` — `Success! The configuration is valid.`
- `terraform test` — `Success! 34 passed, 0 failed.` (unchanged from `main`; this PR adds no
  Terraform and so no new cases)
- Workflow YAML parses with `yaml.safe_load`, and the parsed document shows 76 `TF_VAR_*` keys on
  the `Terraform Plan` step and 76 on the `Terraform Apply` step, with
  `TF_VAR_enable_service_bindings` present in both.
- Default-parity script as described above: 8/8 variables present twice with identical expressions
  and matching defaults.

The diff is 41 insertions and 0 deletions across three files.

PRs from a fork leave upstream CI in `action_required` pending maintainer approval, so the gates
above are local runs rather than a green check on this PR.

## Summary by CodeRabbit

- **Documentation**
  - Expanded deployment setup guidance for R2 media location and bucket configuration.
  - Documented sandbox inactivity timeout, service bindings, unsafe access, and Linear API key settings.
  - Added Terraform configuration details for optional deployment and access settings.

- **Chores**
  - Deployment workflows now support these settings through environment variables and secrets, including defaults for media location, sandbox timeout, user access, and service bindings.


rhlsthrm:
Rebased onto upstream `main` at `30b885b5` and force-updated the existing branch.

Conflict resolution preserved the three bot default-model variables added by #1820 in both Terraform jobs, then re-applied this PR's eight declared-variable bindings. Plan/apply `TF_VAR_*` key sets remain identical.

Verification on the rebuilt branch:
- `terraform test`: 46 passed, 0 failed (with deployment-local `terraform.tfvars` excluded)
- `npm run lint`: passed
- one commit on `origin/main`
- diff scrub found no fork/deployment identifiers

Upstream Actions may remain `action_required` pending maintainer approval.

ColeMurray:

rhlsthrm:
Rebuilt on upstream `dfc4e89a` as `2995cf1a`, preserving this PR's existing head ref. One commit.

**The variable list was re-derived against current upstream rather than carried over.** Enumerating every `variable "<name>"` block under `terraform/environments/production/` (82 declared) and diffing against the `TF_VAR_` keys in each job found eleven declared but unwired. `cloudflare_zone_id` and `cloudflare_custom_domain` were dropped from the original list because `12da4d88` now threads both, so this PR threads six: `enable_service_bindings`, `linear_api_key`, `r2_media_bucket_name`, `r2_media_location`, `sandbox_inactivity_timeout_ms`, `unsafe_allow_all_users`. The commit body was updated from eight to six.

**Fallback parity.** Every literal `|| 'value'` fallback equals that variable's own `default =`: `enable_service_bindings` true, `unsafe_allow_all_users` false, `r2_media_location` `ENAM`, `sandbox_inactivity_timeout_ms` 600000. `r2_media_bucket_name` and the sensitive `linear_api_key` get no literal fallback and render empty when unset, matching their `""` defaults and the repository's convention for sensitive credentials. A throwaway module applied with the exact literals CI renders returned values identical to the Terraform defaults, so a deployment that configures none of these is behaviourally unchanged.

**Deliberately left unwired**, with reasons: `project_root` must keep its checked-out relative default or Terraform would be pointed at a path absent on the runner; `control_plane_migration_tag` and `control_plane_migration_old_tag` are hand-staged for explicit Durable Object migration steps, and an empty Actions string is not `null`; `control_plane_new_sqlite_classes` and `control_plane_deleted_classes` are `list(string)`, which an empty CI string does not satisfy.

**Local gates** (upstream CI is `action_required` for this fork): both jobs thread an identical `TF_VAR_` key set (extracted and diffed, empty); the workflow parses as YAML; `terraform validate` succeeds. No organization identifiers in the diff.

PR #1823, opened 2026-09-07
feat(control-plane): add an RBAC-authorized operator session archive

rhlsthrm:
## What this adds

`POST /operator/sessions/archive`: a cursor-paged maintenance endpoint for archiving sessions without requiring session participation. The authenticated web proxy preserves control-plane response statuses.

## Authorization revision requested by review

The open-inspect CHANGES_REQUESTED review on `db96f153` explicitly requested RBAC instead of a deployment allowlist. This revision adopts that recommendation: `sessions.archive_any` is enforced by `requirePermission(..., { service: "deny" })` before dispatch. Owner and Administrator receive the permission through the existing code-owned built-in role grants; custom roles may receive it. Members and Viewers do not receive it by default. Role changes grant/revoke access without deployment changes.

The underscore name follows the existing `skill_profiles.manage_own` spelling convention. The registry reserves `.any`/`.own` suffixes for paired scoped permissions; this administrative operation has no separate own-scoped grant.

All `OPERATOR_USER_IDS` binding, Node config, environment-example, Terraform and CI additions are removed. No migration is needed. The prior off-by-default allowlist behavior is intentionally replaced by the existing RBAC defaults. The handler no longer makes a second authorization decision after an allowed audit event. The integration regression checks that member denial produces `authorization.request_denied` with `operation_result = denied`, and that changing the assigned role grants and revokes access.

## Archive behavior

- The index supplies an oldest-first keyset page with a fixed cutoff; each session runtime decides whether it can archive. Page size 25, concurrency 5, timeout 10 seconds per runtime request.
- Cancelled sessions and sessions with queued work are skipped. Per-session archive continues using `sessions.lifecycle`; its authorization is unchanged.
- Missing runtimes with archived index rows appear in `missingArchivedIds`. Missing runtimes with non-archived index rows appear in the new `orphanedIds` bucket, rather than pinning the cursor forever. The index row is not silently rewritten or counted as archived.
- Successful and already-archived outcomes explicitly repair the index projection. A projection error returns 500, making the batch report a retryable failure instead of claiming success and advancing past a stale index row. Other runtime failures preserve the same page cursor for retry.
- Strict decimal cursor validation is retained. The malformed-percent and empty-ID test fixtures now reach their intended validation branches.

## Rebase and routing

Reconstructed first on `44580bd3`, then rebased onto the current upstream `d017795a` after main advanced. This is one feature commit with no personal-access-token/MCP or review-supersession changes.

The current Hono catalog mounts `sessionOperatorArchiveRoutes` through the existing sessions router. Authorization is declared through `admit`, and `dispatchSession` supplies the typed user context. The internal session handler is registered through `session/components.ts` and `session/http/routes.ts`.

Exactly one public route is added over `d017795a`: total and unique method/path count 177 → 178, unique paths 135 → 136. The conformance snapshot adds its human-user/RBAC policy, and the admission matrix adds owner=200 and anonymous=401. All other conformance snapshot changes are fixture-index renumbering. The old handler-gated owner-test exception is removed.

## Verification

Final local gate counts and the per-review disposition are recorded in the accompanying rebase comment. Upstream workflow execution on a fork PR may require maintainer approval; local results are not presented as upstream CI approval.

## Summary by CodeRabbit

- **New Features**
  - Added operator session archiving through the web and control-plane APIs.
  - Supports paginated processing, retry-safe results, orphan detection, and detailed archive outcomes.
  - Added the `sessions.archive_any` permission for workspace-wide session management.
  - Already archived sessions are handled safely without unnecessary state changes.
- **Security**
  - Restricted operator archiving to authenticated users with the required permission.
  - Added authorization auditing for denied requests.
- **Documentation**
  - Documented the operator session archive endpoint, authorization requirements, and behavior.


terraform/environments/production/variables.tf
```diff
@@ -722,3 +722,19 @@ variable "unsafe_allow_all_users" {
...
+# users who may act on sessions they do not own.
+variable "operator_user_ids" {
+  description = "Comma-separated list of canonical user IDs authorized for operator-only maintenance actions, such as archiving sessions they do not own. Leave empty to disable operator access."
+  type        = string
+  default     = ""
+
+  validation {
+    condition = alltrue([
+      for user_id in compact(split(",", var.operator_user_ids)) :
+      can(regex("^[0-9a-f]{32}$", trimspace(user_id)))
```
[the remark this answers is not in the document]
rhlsthrm: Fixed in 13cfd12e. Terraform now treats only an entirely blank allowlist as disabled and validates every trimmed comma-separated entry without `compact`, so trailing and doubled commas fail during planning instead of reaching the runtime 500 path. Added `operator_user_id_empty_entry`; the full Terraform suite passes 36/36.

packages/web/src/app/api/operator/sessions/archive/route.ts
```diff
@@ -0,0 +1,32 @@
+import { controlPlaneUserFetch } from "@/lib/control-plane";
+import { getServerAuthSession } from "@/lib/server-auth-session";
+import type { NextRequest } from "next/server";
+import { NextResponse } from "next/server";
+
+export async function POST(request: NextRequest): Promise<NextResponse> {
+  const session = await getServerAuthSession();
```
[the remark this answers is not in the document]
rhlsthrm: Verified the premise but not applying the suggested local catch. `getServerAuthSession()` can throw, but all 47 API routes that call it do so outside their route-local `try`; none catches `AuthenticationUnavailableError`. The nearest archive/title proxy routes follow this exact 401/400/500 shape. Handling this only here would create a second boundary convention; any change belongs in the shared auth/framework boundary across all callers.

rhlsthrm:
Rebuilt on upstream `4f3f7bcd` in `13cfd12e`, retaining the new upstream session-budget route. The Terraform allowlist validation now rejects empty entries, including doubled commas. I did not add a route-local `AuthenticationUnavailableError` catch: all 47 sibling API routes authenticate outside route-local `try` blocks, so that behavior belongs at the shared boundary rather than this route.

Local verification: affected typechecks; control-plane unit (4137 passed); integration (1227 passed, 1 skipped); focused web route tests (2 passed); Terraform suite (36 passed).

packages/control-plane/src/session/operator-archive.ts
```diff
@@ -0,0 +1,269 @@
...
+
+export function parseOperatorArchiveCursor(
+  raw: string | null | undefined
+): { ok: true; cursor: OperatorArchiveCursor | null } | { ok: false; error: "Invalid cursor" } {
+  if (raw === null || raw === undefined) return { ok: true, cursor: null };
+
+  const parts = raw.split(":");
+  if (parts.length !== 3) return { ok: false, error: "Invalid cursor" };
+  const cutoffCreatedAt = Number(parts[0]);
+  const createdAt = Number(parts[1]);
```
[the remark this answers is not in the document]
rhlsthrm: Fixed in cee2040d. Both numeric cursor fields must now match decimal digits before conversion; empty, whitespace, exponent, hex, fractional, unsafe, malformed-percent, ordering, and empty-resume cases are rejected. Every cursor emitted by `encodeOperatorArchiveCursor` remains accepted. Targeted cursor tests: 15 passed; typecheck, ESLint, Prettier, and diff check are clean.

rhlsthrm:
Rebuilt and force-pushed as `efdfb150`, preserving this PR's existing head ref. Reconstruction began on `44580bd3`; while gates ran, upstream advanced, so the final one-feature commit is based on `d017795a` instead. There are no personal-access-token/MCP or review-supersession changes and no new migration.

### Review dispositions

All three inline CodeRabbit threads were already resolved when inspected through GraphQL. Their dispositions have been rechecked:

- Web proxy auth-service catch: retained the prior convention-based decline. Authentication remains at the shared/framework boundary, matching adjacent archive/title proxies; this route does not introduce its own distinct AuthenticationUnavailableError boundary.
- Terraform empty-entry validation: superseded by the requested RBAC conversion. The entire operator allowlist binding/configuration/Terraform/CI addition is removed, not merely weakened or bypassed.
- Decimal cursor validation: preserved. The later review-body nitpick was valid: the empty-ID and malformed-percent fixtures now use `2:1:` / `2:1:%E0%A4%A`, so they reach the intended branches rather than failing the earlier cutoff-order check.

The open-inspect CHANGES_REQUESTED review body on `db96f153` was not stale. Its three concerns are addressed:

1. **Authorization and audit alignment:** replace the deployment allowlist with `sessions.archive_any`, granted through the existing Owner/Administrator defaults and available to custom roles. `requirePermission(..., { service: "deny" })` enforces it in Hono admission; no handler-owned allowlist decision remains. The integration regression exercises member denial, checks a denied (not allowed/applied) audit event, and proves role-based grant/revocation without deployment changes. The underscore suffix follows the existing `manage_own` convention rather than creating an unpaired `.any` permission.
2. **Missing-runtime cursor starvation:** non-archived index rows whose runtime returns 404 are surfaced as `orphanedIds` and do not pin the cursor. They are neither counted as successfully archived nor silently rewritten in the index.
3. **Suppressed index projection failures:** the operator handler calls the existing throwing `repairIndexStatus` path after archived/already-archived outcomes; projection failure produces HTTP 500 and a retryable batch failure, instead of a success that advances past stale index state.

### Wire-contract changes

The endpoint is no longer disabled until an env allowlist is configured: Owner and Administrator are authorized by RBAC, as requested in review. A missing permission returns the standard 403 `permission_required` response. Successful batch responses add `orphanedIds`; transient failures still preserve the page cursor for retry. Per-session archive authorization is unchanged.

The route fixtures now advance exactly one route from the final base: **177 → 178 routes/method-path pairs; 135 → 136 unique paths**. Read-back of the refreshed snapshots confirms the new human-user/RBAC admission policy, **owner=200**, **anonymous=401**, and no other semantic row changes (the conformance fixture's other changes are index renumbering). The former handler-gated owner-test exception is gone.

### Local gates on the final base

| Gate | Result |
| --- | --- |
| `npm run typecheck` | exit 0 across all workspaces, including control-plane Node/test/integration configs |
| `npm test -w @open-inspect/control-plane` | 287 files, **4246 passed** |
| `npm run test:integration -w @open-inspect/control-plane` | 106 files, **1255 passed, 1 skipped** |
| Scoped snapshot refresh | 2 files, 9 passed; both snapshot diffs inspected |
| Shared RBAC registry test | 1 file, 13 passed |
| Web operator archive proxy test | 1 file, 2 passed |
| Cursor/projection regression falsification | Reintroducing the two defects causes 3 failures; restoring the fix gives 44/44 passing targeted tests |
| Required org-leak scan of `git diff origin/main...HEAD` | empty |

Upstream workflow execution on this fork PR requires maintainer approval; these are local results, not a claim of upstream CI approval. The outstanding CHANGES_REQUESTED review decision needs reviewer re-evaluation; there were no unresolved inline threads at triage.

Publication check: GitHub reports **MERGEABLE** against `d017795a`; mergeStateStatus remains BLOCKED and reviewDecision remains CHANGES_REQUESTED pending reviewer re-evaluation. Per-thread REST replies were rejected with HTTP 422 because the current user already has multiple pending reviews. The GraphQL alternative only created a PENDING draft, which I removed; no existing review drafts were deleted or submitted. All per-thread dispositions are therefore recorded publicly above, and the three threads remain resolved.

packages/web/src/app/api/operator/sessions/archive/route.ts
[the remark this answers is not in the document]
rhlsthrm: Recording the disposition here as well as in the summary comment, now that a stale pending review draft is out of the way. Premise verified, suggestion declined on convention: `getServerAuthSession()` can throw, but every API route in this package calls it outside the route-local `try`, and none catches `AuthenticationUnavailableError` — the neighbouring archive/title proxies use exactly this 401/400/500 shape. Catching it only here would create a second boundary convention; the fix belongs in the shared auth boundary across all callers. Still accurate against the current head (efdfb150, rebased onto d017795a).

packages/control-plane/src/routes/session-operator-archive.ts
```diff
@@ -0,0 +1,67 @@
...
+  return json(result);
+}
+
+export const sessionOperatorArchiveRoutes = new Hono<ControlPlaneHonoEnv>();
+
+sessionOperatorArchiveRoutes.post(
+  "/operator/sessions/archive",
+  admit({
+    ...SCM_AGNOSTIC_HUMAN_USER_ROUTE,
+    authorization: requirePermission("sessions.archive_any", { service: "deny" }),
```
ColeMurray, line 64: [Automated Review] [P2] Keep bulk archiving in the sessions domain; the claimed participant boundary does not exist

The PR/README describes ordinary `/sessions/:id/archive` as participant-scoped, but `session-runtime-proxy.ts` admits it with workspace `sessions.lifecycle`, and `SessionLifecycleHandler.archive()` performs no participation check. I verified through the real public route that a Member who is not a session participant can archive it successfully. The existing integration suite also explicitly tests that the internal archive operation does not use participant identity for authorization.

This new capability can legitimately have stricter bulk-operation authorization, but it is not a distinct ability to archive other people's sessions. Adding `/internal/operator-archive`, its own request/outcome/status protocol, and an operator-specific method on the lifecycle handler creates a parallel operation for the same domain mutation. The ordinary handler now imports its outcome type from the batch orchestration module, reversing the natural dependency direction.

Please expose this as a sessions collection operation and share one canonical archive operation/result contract beneath route-level authorization. Preserve a separate bulk permission if intended, but remove the operator-specific lifecycle and correct the participant-scoped documentation.

packages/control-plane/src/session/http/handlers/session-lifecycle.handler.ts
```diff
@@ -138,19 +158,68 @@ export class SessionLifecycleHandler {
...
+    const sessionId = resolvePublicSessionId(session, this.durableObjectId);
+    const outcome = await this.getArchiveOutcome(session.status);
+
+    // `transition` swallows a failed index projection, so a caller driving a
+    // batch would count this session as done and advance past it while the
+    // index row still reads live. Re-projecting surfaces that failure: it is a
+    // no-op once the row already agrees, and throws when it still does not.
+    if (outcome === "archived" || outcome === "already_archived") {
+      try {
+        await this.statusService.repairIndexStatus();
```
ColeMurray, line 198: [Automated Review] [P1] This repair does not establish that a non-draft index row was archived

`SessionStatusService.repairIndexStatus()` calls `SessionIndexStore.repairStatus()`, whose SQL is restricted to `WHERE id = ? AND status = 'created' AND status != ?`. It deliberately repairs only draft projections. A no-op resolves successfully; it does not throw when another status disagrees.

Reproduced against real D1: archive the runtime, leave its index row `active` or `completed` with `updated_at` newer than the runtime's archived timestamp, then call this endpoint. The same-status transition cannot overwrite the newer index timestamp, this repair matches zero rows, and the endpoint returns 200 with the ID in `alreadyArchivedIds`, `failed: []`, and `hasMore: false`. D1 still contains the live status.

Please make the canonical projection operation establish agreement for the statuses this batch handles, or report a retryable failure before advancing. Preserve concurrency ordering: simply removing the draft guard could clobber a newer transition. Add a real-D1 regression for this no-op case; mocking `repairIndexStatus()` to reject only covers thrown database errors.

packages/control-plane/src/session/http/handlers/session-lifecycle.handler.ts
```diff
@@ -138,19 +158,68 @@ export class SessionLifecycleHandler {
...
+    } catch {
+      return Response.json({ error: "Invalid request body" }, { status: 400 });
+    }
+    const parsed = operatorArchiveRequestSchema.safeParse(raw);
+    if (!parsed.success) {
+      return Response.json({ error: "Invalid request body" }, { status: 400 });
+    }
+
+    const sessionId = resolvePublicSessionId(session, this.durableObjectId);
+    const outcome = await this.getArchiveOutcome(session.status);
```
ColeMurray, line 190: [Automated Review] [P1] Re-read lifecycle state after parsing the body

`session` was read at line 173, before `await request.json()` at line 180. This passes the stale status into the mutation decision.

Two controlled interleavings reproduce 200 responses where 409 is required:

- Read `completed`; another request cancels the session while parsing yields; resume and archive it, overwriting cancellation.
- Read `archived`; another request reopens the session and queues work; resume into `getArchiveOutcome`'s `archived` branch, which calls `transition('archived')` before checking queued work.

The second case archives a session with pending work and falsely classifies it as already archived. Parse/validate first, then read current session state and check eligibility immediately before the synchronous local mutation, with no intervening await. These protections belong in the canonical archive operation, not in a caller-supplied status snapshot.

packages/control-plane/src/session/operator-archive.ts
```diff
@@ -0,0 +1,263 @@
...
+        throw new Error(`Unhandled operator archive outcome: ${exhaustive}`);
+      }
+    }
+  }
+
+  // A failure is retryable, so the page is replayed from the same cursor until
+  // it clears. Only outcomes that can change on a retry may pin the cursor.
+  if (result.failed.length > 0) {
+    result.hasMore = true;
+    result.nextCursor = encodeOperatorArchiveCursor(cursor ?? { cutoffCreatedAt, resume: null });
```
ColeMurray, line 240: [Automated Review] [P2] Separate pagination progress from retries of failed sessions

Every rejection pins the entire page to the same cursor indefinitely. A persistently failing runtime (for example a runtime initialization/storage error or consistently invalid response) blocks every later session, while successfully processed siblings are called again on each attempt. The new orphan classification solves only the 404 instance of this liveness problem.

A probe with 26 candidates and a persistent failure in the first 25 makes 75 runtime calls across three attempts, returns the same cursor each time, and never reaches candidate 26. There is no failed-item retry contract or bounded escape from that page.

Please advance scan progress independently and return failed IDs for explicit retry, or maintain separate continuation/retry state. A bounded explicit-ID batch would make retrying only failures direct and avoid conflating `hasMore` with 'replay this page until every runtime succeeds'.

packages/control-plane/src/db/session-index.ts
```diff
@@ -773,6 +787,38 @@ export class SessionIndexStore {
     return (result.meta?.changes ?? 0) > 0;
   }
 
+  async listOperatorArchiveCandidates(options: {
+    cutoffCreatedAt: number;
+    cursor: OperatorArchiveCandidateCursor | null;
+    limit: number;
+  }): Promise<OperatorArchiveCandidatePage> {
```
ColeMurray, line 794: [Automated Review] [P2] Decompose the index store before adding another feature-specific concern

This change pushes `session-index.ts` from 981 to 1,027 lines and its test from 989 to 1,046. The added cursor/page/candidate types and `listOperatorArchiveCandidates` introduce an operator-specific query contract into an already broad store.

Under the requested strict maintainability bar, crossing 1,000 lines needs a compelling structural justification or decomposition first. Please keep any retained maintenance-selection query in a focused session query module, with domain-neutral types and focused tests. If this becomes an explicit-ID batch over existing session selection, the bespoke scan and cursor machinery can disappear entirely; that is preferable to merely relocating the same complexity.

ColeMurray (changes requested):
[Automated Review] Request changes

Reviewed `efdfb1502749bfa83201466293a87003760bee54` against `d017795a29de3541d58480ccecba78122240259e`, focusing on strict maintainability, architecture fit, authorization, and lifecycle correctness.

The `/operator/` concern is substantive, not just URL naming: this introduces a parallel internal archive operation and makes ordinary lifecycle code depend on the batch-specific contract. The stated distinction from participant-scoped archiving does not match the current system: a non-participant Member can already archive through the existing public sessions endpoint.

Five inline findings cover:

1. The unnecessary operator-specific archive domain and incorrect authorization premise (P2).
2. Successful batch results despite unrepaired non-draft index rows (P1).
3. Stale state across body parsing bypassing cancellation/queued-work protections (P1).
4. Permanent runtime failures indefinitely pinning pagination (P2).
5. Feature-specific store growth crossing the 1,000-line decomposition threshold (P2).

### Recommended shape

Use a sessions collection operation, such as `POST /sessions/batch-archive`, with bounded explicit targets, one canonical runtime archive operation, and typed per-session results. Keep a stricter bulk-operation permission if that is the intended policy; this does not require a separate operator lifecycle. If a server-side archive-all sweep is required, separate its selection/pagination from mutation and failed-item retries.

### Verification

247 existing focused tests passed across control-plane unit tests, operator authorization integration tests, shared RBAC, and the web proxy. Additional isolated review probes reproduced four failing cases: stale `active` and `completed` D1 projections, plus cancellation and reopen-with-queued-work interleavings. A real public-route integration probe confirmed that a non-participant Member can use ordinary archiving; another probe confirmed repeated cursor pinning for a persistently failing runtime.

The new route's RBAC denial and role grant/revocation checks pass. This is not a claim that the full test suite or upstream CI passed. No tracked implementation files were modified.

ColeMurray:
thanks! superseded by https://github.com/ColeMurray/background-agents/pull/1869

PR #1863, opened 2026-09-11
fix(sandbox): set git author globally and forbid cloning other repos

LarsBuur:
## Why

Two monorepo PRs (dreamplan-io/monorepo #8210 and #8214) were authored as "OpenCode Bot <bot@dreamplan.io>", and Vercel refused to deploy them because that email maps to no GitHub account on the team. Both came from sessions started on other repositories. The agent cloned monorepo by hand inside the sandbox, hit git's "please tell me who you are", and invented an identity.

Open-Inspect only applied the participant's noreply identity with `git config --local` on manifest repositories, so a checkout the agent created itself had none.

## What

- The signing runtime also writes `user.name` and `user.email` with `git config --global` on every prompt. Signing keys stay local to manifest repos. A stray clone now commits as the session user.
- A runtime-owned instruction file, `sandbox_runtime/instructions/open-inspect.md`, is loaded into every session through OpenCode's `instructions` config. It tells the agent to work only in the checked-out repositories, never clone or push elsewhere, and to stop and tell the user to start a session on the right repository when the code it needs is not present. It also says not to touch the git identity and to use the `create-pull-request` tool rather than `gh`.
- Runtime manifest bumped to generation 64 with a matching rebuild floor so prebuilt repo images pick up the new runtime.
- Test conftest points HOME at a temp dir so the global writes never touch a developer's real gitconfig.

## Testing

- sandbox-runtime: 948 passed, 3 skipped; ruff clean
- control-plane sandbox and image-builds suites: 774 passed
- New tests cover the global identity on both signing paths, including an actual commit in a repo outside the manifest, and the instruction wiring plus the bundled file.

## Deploy

`terraform apply` in `terraform/environments/dreamplan` rebuilds the sandbox image and redeploys Modal and the control plane. No web changes. The Modal backend is shared with the bandcizer environment; compatibility floor stays at generation 62.

## Summary by CodeRabbit

* **New Features**
  * Session attachments now support PDF and Markdown files alongside images.
  * Non-image attachments display file-type badges and filenames in previews.
  * Added manual production deployment support and configurable custom domains.
  * Added infrastructure options for Cloudflare, Vercel, Modal, Daytona, storage, and optional bot services.

* **Improvements**
  * Sandbox startup and connection timeouts increased to five minutes.
  * Git author identity is now applied consistently across repositories.
  * Added session guidance for repository usage and pull-request workflows.

* **Bug Fixes**
  * Improved attachment validation and clearer unsupported-type and size-limit messages.
  * Added deployment configuration validation for authentication and access controls.


LarsBuur:
Opened against the wrong repository by mistake; this belongs on a fork. Sorry for the noise.

PR #1884, opened 2026-09-13
refactor: key worker bindings by name

ColeMurray:
## Summary

- migrate the reusable Cloudflare Worker module's KV, service, D1, R2, queue producer, plaintext, secret, and Durable Object inputs from lists to maps keyed by Worker binding name
- remove repeated `binding_name`/`name` fields from map values and update all four production module callers plus the shared classifier secret binding local
- render the same Cloudflare binding names, types, and target values while preserving the service-binding and Durable Object enablement gates
- sort binding keys when projecting maps because the Cloudflare provider models `bindings` as a list, producing deterministic plans

## State and plan impact

Bindings remain nested `cloudflare_worker_version` data, not independently addressed Terraform resources, so no state moves or resource-address changes are required. Existing deployments may show a one-time binding reorder (and therefore a new Worker version) as prior caller order is normalized to lexical order within each binding type; binding semantics and payload content are unchanged.

The OpenNext web Worker does not use this module and continues to generate its bindings through Wrangler configuration.

## Validation

- `terraform fmt -check -recursive` from `terraform`
- `terraform init -backend=false` from `terraform/environments/production`
- `terraform validate -no-color` from `terraform/environments/production`
- `terraform test -no-color` from `terraform/environments/production` (46 passed, 0 failed)
- `git diff --check`

No apply or deployment was run.

---
*Created with [Open-Inspect](https://open-inspect-prod.vercel.app/session/754044b455691c9af0afaee4dceda002)*

## Summary by CodeRabbit

- **Refactor**
  - Standardized worker deployment bindings across storage, services, queues, text values, secrets, and Durable Objects.
  - Preserved existing resource references, conditional behavior, secret selection, and worker capabilities.
  - Improved consistency and predictability when processing bindings and generating infrastructure plans.
  - Updated binding outputs and configuration defaults to align with the standardized format.

- **Tests**
  - Updated queue-binding validation to recognize the revised configuration format.


ColeMurray:
Reviewed this end to end — the refactor is faithful and I'd merge it. I pushed two small follow-ups to the branch (f5200f590); details at the bottom.

## Verification

Rather than eyeball the translation, I checked it mechanically against `main`:

- **Binding name sets** across all four callers: identical, same declaration order, no duplicates.
- **Name→value pairings**: 100 bindings compared, **0 mismatches**.
- **Conditional guards** (`cond ? [...] : []` → `cond ? {...} : {}`): 25 groups, every guard preserved verbatim.
- **Module callers**: 4/4 updated; no `binding_name` left anywhere in the tree.
- **Output consumers**: all `contains()` / key lookup, so the lexical reorder can't reach them.
- `terraform fmt -check -recursive` clean, `terraform validate` success, `terraform test` **46 passed, 0 failed**.
- `scheduled-jobs.test.ts` parses `cron_triggers`, still a list — unaffected.
- `migrations` stays `null` in the steady state (`enable_durable_object_bindings` true, `deleted_classes` empty → second clause false), so no DO migration is re-emitted.

The riskiest construct here is `classifier_secret_bindings`: a ternary whose two branches are objects with *different* attribute names. Object-type unification could plausibly have leaked both keys. `classifier_provider.tftest.hcl` covers both branches with negative assertions (`!contains(..., "OPENAI_API_KEY")`) and passes — so that's genuinely safe rather than merely untested.

## One thing to confirm before applying

The `Plan` job skips on this PR (`check-secrets` finds none), so nothing in CI exercised a real plan — `Validate` and `terraform test` are config-level and insensitive to binding order by construction. The one-time reorder the description predicts is therefore unverified.

Worth reading an actual plan from a deployment that has credentials, and confirming it is a pure reorder: same binding count per worker, nothing dropped, and no `migrations` block appearing on the control plane. A new worker version restarts the Durable Objects it hosts, so this is not a free no-op apply.

## Non-blocking notes

**`merge` turns a loud duplicate into a silent one.** `concat` would have emitted two bindings sharing a name and let Cloudflare object; maps last-win silently — and not only across `merge` arguments. HCL accepts a duplicate key *inside a single object literal* with no error or warning:

```hcl
locals { m = { A = { value = "one" }, A = { value = "two" } } }   # validates; yields "two"
```

No collisions today (I checked all 26 merge arguments), but `merge(base, local.classifier_secret_bindings)` is exactly the shape where a future `ANTHROPIC_API_KEY` added to a base map would be silently overridden.

**The test regex got looser and position-dependent.** The old `binding_name = "X"` anchor was a token that could only occur inside a binding; `^\s*(\w+)\s*=\s*\{` matches any block header and leans on the `cloudflare_queue.` tail plus the module slice for specificity. It also now requires `queue_name` to be the *first* attribute in the map value — anything inserted ahead of it silently drops that binding from the parsed map. I left it alone because it fails **closed**: `expect(consumer.bindingName).toBe(JOB_QUEUE_BINDINGS[kind])` compares against a concrete string, so a parse miss fails the test rather than passing vacuously. Confirmed 3/3 parse today.

## Pushed to the branch (f5200f590)

**Dropped the redundant `sort(keys(...))`.** `keys()` already returns keys in lexicographical order, so the sort was a no-op — and the comment claiming otherwise would misdirect the next reader:

```
keys({ ZEBRA = 1, alpha = 2, MIKE = 3, BRAVO = 4 })  =>  ["BRAVO", "MIKE", "ZEBRA", "alpha"]
```

Each of the eight projections also loses a double map lookup:

```diff
-    [for binding_name in sort(keys(var.kv_namespaces)) : {
+    [for binding_name, binding in var.kv_namespaces : {
       type         = "kv_namespace"
       name         = binding_name
-      namespace_id = var.kv_namespaces[binding_name].namespace_id
+      namespace_id = binding.namespace_id
     }],
```

`new_sqlite_classes` goes back to `[for binding in var.durable_objects : binding.class_name]`, byte-identical to the pre-PR expression, which removes that line from the diff entirely.

**Removed a stray comma** in `queue_bindings` — the only map in the PR using comma separators. `terraform fmt` accepts both, so it would not have self-corrected.

Re-verified after both edits: fmt clean, validate success, `terraform test` 46/0, 100 bindings still equivalent to `main`, and the queue regex still parses 3/3.

PR #1885, opened 2026-09-13
feat(web): open changed files from agent output links

### LINKED ISSUE #1875: feat: Clicking a relative file path in agent output should open it in the changes panel
## Summary

When an agent references a repository file in its output, the markdown renderer turns it into an
ordinary link:

```html
<a href="docs/plans/daytona-prebuild-parity.md" target="_blank" rel="noopener noreferrer nofollow"
   class="text-accent hover:underline">daytona-prebuild-parity.md</a>
```

The href is repo-relative, so the browser resolves it against the session URL and the user lands on
`/session/docs/plans/daytona-prebuild-parity.md`, which 404s.

That file is almost always sitting right there in the **Changes** sidebar. Clicking the link should
select it in the changes panel instead of navigating away.

## Current behavior

`SafeMarkdown` renders every link identically — raw `href` plus `target="_blank"`
(`packages/web/src/components/safe-markdown.tsx:74-84`). Relative hrefs survive sanitization, because
the schema's `protocols.href` allowlist only constrains links that carry an explicit scheme
(`packages/web/src/components/safe-markdown.tsx:52-54`). The markdown layer has no knowledge of the
session or its diff manifest, so a repo-relative path renders as a dead external link.

## Desired behavior

- Clicking a relative link in agent output that matches a file in the session's diff manifest selects
  that file in the changes panel — same result as clicking the row in **Files changed**.
- Such a link no longer opens a new tab and no longer navigates.
- A relative link that does **not** match a changed file renders as inert text rather than a link
  (see *Scope* below). It is styled so it doesn't look clickable, and ideally carries a `title`
  explaining why (e.g. "Not in this session's changes").
- Absolute URLs (`https://…`), anchors (`#section`), and `mailto:` links keep their current behavior
  exactly.

## Notes / implementation considerations

The selection API already exists; most of this work is getting a callback to the markdown renderer
and mapping an href string onto a manifest entry.

**What's already there**

- `selectedDiff: DiffSelection | null` — `{ repositoryPosition, path }`
  (`packages/web/src/app/(app)/(sidebar)/session/[id]/page.tsx:193`, type at
  `packages/web/src/lib/session-diffs.ts:12`)
- `openDiff(repository, file)` — sets the selection, closes the mobile details sheet, and records
  focus-return state (`packages/web/src/app/(app)/(sidebar)/session/[id]/page.tsx:286-291`)
- `resolveDiffSelection(manifest, selection)` — resolves a `{position, path}` pair against the
  manifest (`packages/web/src/lib/session-diffs.ts:78`)
- `SessionChangesPanel` then fetches the patch by file id
  (`packages/web/src/components/session-changes-panel.tsx:254-256`)

**1. Plumbing — prefer context over prop threading**

`SafeMarkdown` renders at `packages/web/src/components/session-timeline.tsx:555`, inside a memoized
`EventItem` (`packages/web/src/components/session-timeline.tsx:669`), inside a virtualizer. Threading
a callback prop would touch `SessionTimeline` → `renderFlatItem` → `EventItem` → the `eventRenderers`
map → `SafeMarkdown`, and would break the `memo` on `EventItem` unless the callback is perfectly
stable.

A small context provider (e.g. `packages/web/src/lib/session-file-links.tsx`) exposing
`{ resolve(href), open(selection) }` is cleaner. Mount it in the session page, which already holds
both `diffState` and `openDiff`. `SafeMarkdown` reads it with a `useContext` that returns `null` when
there's no provider, so the other call site
(`packages/web/src/components/create-pull-request-event.tsx`) is untouched.

Memoize the context value on `diffState.current.revisionId` rather than on manifest object identity —
otherwise every SWR revalidation re-renders every visible markdown row in the timeline.

**2. Path resolution — the actual design work**

Worth its own pure, unit-tested module (e.g. `packages/web/src/lib/diff-file-links.ts`):

- Skip non-candidates: absolute URLs with a scheme, protocol-relative `//`, bare anchors `#…`,
  `mailto:`.
- Normalize: strip a leading `./`, strip query and hash, decode percent-escapes.
- Absolute sandbox paths: `/workspace/<repoName>/<rest>` maps cleanly, because checkouts live at
  `/workspace/{repoName}` and `repoName` is unique per session — enforced at
  `packages/shared/src/types/repositories.ts:137-153`.
- Match against the manifest on exact `file.path` per repository, and also on `file.oldPath` so
  renamed files resolve.
- Multi-repo sessions can produce the same path in two repositories. Prefer the lowest `position`
  (the primary), or require a `<repoName>/` prefix to disambiguate — either is acceptable, just make
  the choice deliberate and cover it with a test.
- Compare paths case-sensitively (git paths are), but compare `repoName` case-insensitively, matching
  `prArtifactBelongsToRepo` in `packages/shared/src/types/repositories.ts`.
- *Optional follow-up, not required here:* a unique path-boundary **suffix** match, for when the
  agent writes `docs/plans/x.md` but the manifest path is `packages/web/docs/plans/x.md`.
  `buildUniquePathLabels` (`packages/web/src/lib/session-diffs.ts:107`) is the same idea in reverse.

**3. Rendering in `SafeMarkdown`**

- Resolved link: drop `target="_blank"`, and `onClick` → `preventDefault()` + open the selection.
- Route clicks through `openDiff`, not `setSelectedDiff` directly — it also closes the mobile details
  sheet and records `diffReturnFocusRef`
  (`packages/web/src/app/(app)/(sidebar)/session/[id]/page.tsx:286-312`).
- No sanitizer changes are needed: the implementation only reads `href`, which is already allowlisted
  (`packages/web/src/components/safe-markdown.tsx:46`). Please avoid the alternative of a rehype
  plugin stamping `data-*` attributes onto nodes — that would force widening the sanitize schema.

**4. Focus and mobile**

- Focus-return on panel close looks for `button[data-diff-path]` in the sidebar, but the sidebar is
  hidden while the panel is open
  (`packages/web/src/app/(app)/(sidebar)/session/[id]/page.tsx:458`). It already falls back to
  `focusDetailsTrigger()`, so nothing is broken — but returning focus to the originating link would
  be nicer.
- Mobile needs no special handling: the panel opens as a full-screen `Sheet`
  (`packages/web/src/app/(app)/(sidebar)/session/[id]/page.tsx:541`).

## Scope

**Files that aren't in the diff render as inert text.** The control plane exposes only the diff
manifest and per-file patches (`packages/control-plane/src/routes/session-diffs.ts`) — there is no
read-file API into the sandbox, so a link to a file the agent merely *read* has nowhere to go.
Rendering it as text is still strictly better than today's 404. Adding a sandbox file-read endpoint
and a plain-file viewer mode is a much larger change and is **not** part of this issue.

Falling back to a forge blob URL (`https://github.com/owner/name/blob/...`) is also out of scope and
would be wrong as-is: `sessionRepositoryStateSchema`
(`packages/shared/src/types/repositories.ts:29-38`) carries owner, name, and branch but no SCM
provider or host, so it would silently mislink every GitLab session.

**Line anchors are out of scope.** Handling `file.ts:42` or `#L42` by scrolling the diff to that line
is a separate change — `PierreDiffRenderer` (`packages/web/src/components/pierre-diff-renderer.tsx`)
exposes no line-jump API today. Stripping a trailing line reference during normalization so the file
itself still resolves is welcome, but jumping to the line is not required.

**Deep linking is a possible follow-up, not required here.** `selectedDiff` is local state with no URL
sync. Encoding the selection in a query param would give shareable links and make cmd-click work
naturally, but it can land separately.

## Files

- `packages/web/src/components/safe-markdown.tsx` — the `a` renderer
- `packages/web/src/lib/diff-file-links.ts` *(new)* — href → `DiffSelection` resolution
- `packages/web/src/lib/session-file-links.tsx` *(new)* — context provider
- `packages/web/src/app/(app)/(sidebar)/session/[id]/page.tsx` — mount the provider, wire `openDiff`
- `packages/web/src/lib/diff-file-links.test.ts` *(new)* — resolver unit tests
- `packages/web/src/components/safe-markdown.test.tsx` *(new)* — rendering behavior

## Acceptance criteria

- [ ] Clicking a relative link matching a changed file selects that file in the changes panel, with
      no navigation and no new tab
- [ ] Renamed files resolve via `oldPath` as well as `path`
- [ ] `/workspace/<repoName>/<path>` absolute sandbox paths resolve
- [ ] Relative links with no matching changed file render as inert text, not as broken links
- [ ] Absolute URLs, anchors, and `mailto:` links behave exactly as they do today
- [ ] Multi-repo path collisions resolve deterministically, with a test covering the chosen rule
- [ ] `SafeMarkdown` outside a session (e.g. `create-pull-request-event.tsx`) is unaffected
- [ ] No changes to the rehype-sanitize schema
- [ ] Unit tests for the resolver and rendering tests for `SafeMarkdown`

burakeyler:
Closes #1875

## Summary

A repo-relative link in agent output (e.g. `docs/plans/daytona-prebuild-parity.md`) used to render as a `target="_blank"` link that resolved against the session URL and 404'd. Inside a session it now:

- **matches a changed file:** selects that file in the changes panel (no navigation, no new tab);
- **doesn't match:** renders as inert text with `title="Not in this session's changes"`.

Absolute URLs, anchors and `mailto:` behave exactly as before, and `SafeMarkdown` outside a session (`create-pull-request-event.tsx`) is unchanged.

## How it's built (following the notes in the issue)

**`lib/diff-file-links.ts`** (new, pure)
- `isRepositoryFileHref(href)` skips anything with a scheme (`https:`, `mailto:`), protocol-relative `//`, bare `#…`, and root-relative app paths other than `/workspace/…`.
- `resolveDiffFileLink(manifest, href)` normalizes first: strips `./`, query and hash, decodes percent-escapes (a malformed escape resolves to nothing), and strips a trailing `:42` / `:42:7` line reference so the file still resolves.
- Matching:
  - on `file.path` or `file.oldPath` (renames), case-sensitively;
  - `/workspace/<repoName>/<path>` resolves within that repository, with `repoName` compared case-insensitively;
  - repositories with `status: "unavailable"` are skipped.
- **Multi-repo collision rule:** a relative path present in several repositories resolves to the **lowest `position`** (the primary). A test covers it, with repositories deliberately listed out of order.

**`lib/session-file-links.tsx`** (new)
- Context exposing `{ resolve(href), open(selection) }`. `useSessionFileLinks()` returns `null` without a provider.
- The value is memoized on `manifest.revisionId` (with the manifest held in a ref), not on manifest identity, so SWR revalidation doesn't re-render every visible markdown row.

**`components/safe-markdown.tsx`**
- The `a` renderer became a `MarkdownLink` component that reads the context.
- Resolved links drop `target`/`rel` and `preventDefault()` on click.
- **No rehype-sanitize schema changes:** it only reads the already-allowlisted `href`.

**`session/[id]/page.tsx`**
- `openDiff` now delegates to a new `openDiffSelection(selection)`, so link clicks go through the same path as sidebar rows: set selection, record `diffReturnFocusRef`, close the mobile details sheet.
- The provider wraps `SessionTimeline`. No prop threading through `EventItem`, so its `memo` is untouched.

Not included (listed as optional or out of scope in the issue): suffix matching, returning focus to the originating link, line jumps, deep linking.

## Tests

- **`lib/diff-file-links.test.ts`:** candidate detection; relative match; `./` / query / hash / escapes / line refs; rename via `oldPath`; case-sensitive paths; `/workspace/<repoName>` (case-insensitive name, unknown repo, missing path); collision → lowest position; unavailable repositories skipped; non-matches and malformed escapes.
- **`components/safe-markdown.test.tsx`:** click opens the selection and prevents default with no `target`; a non-matching repo file renders as an inert `<span>` with the title; absolute URLs keep `target="_blank"` and `rel`; anchors unchanged; relative links unchanged without a provider.

## Checks (run locally)

- `vitest run` on the two new files plus `session-diffs.test.ts` and `create-pull-request-event.test.tsx`: 50 passed.
- `tsc --noEmit` in `packages/web`: clean.
- `eslint` and `prettier --check` on all touched files: clean.
- Not run: the full workspace `npm test` / `npm run typecheck` (with `next typegen`), and I haven't clicked through it in a running session.

## Summary by CodeRabbit

- **New Features**
  - Markdown links to changed repository files now open the corresponding file in the Changes panel.
  - Links support repository paths, renamed files, line references, encoded paths, and query or hash fragments.
  - Unresolved repository-file links remain readable without opening a file.
  - External URLs, anchor links, and other standard links continue to behave normally.
  - Markdown content can resolve relative links and safely omit images or display image placeholders.


ColeMurray (changes requested):
-

burakeyler:
Thanks for the review. Pushed 1b4d0b2b, which covers all five points:

- **Button instead of anchor:** a resolved file link now renders as a link-styled `<button>`, so middle-click and "Open link" have no session-relative href to follow.
- **Opt-in only:** file-link handling moved behind a `linkRepositoryFiles` prop on `SafeMarkdown`, set only by `AssistantMessageEvent`. Added a test that renders `CreatePullRequestEvent` inside `SessionFileLinksProvider` and checks its relative link stays a plain new-tab link.
- **Root-level line references:** `README.md:42` / `package.json:10:2` are no longer mistaken for a scheme. The resolver tries the literal decoded path first and only strips a line suffix as a fallback, so a real `notes:42` file still wins. Tests cover both.
- **No render-phase ref mutation:** the provider keeps one manifest per revision in state and builds an immutable resolver (path set + old-path map per repo) with ordinary `useMemo` dependencies; the lint suppression is gone and each link is a couple of map lookups.
- **Rename order:** exact current paths are checked across all repositories before any old-path alias, with a test for rename `z.ts → a.ts` plus a new `z.ts`.

`vitest` for `src/lib`, `safe-markdown`, `create-pull-request-event` and `session-timeline`: 582 passed; eslint and `tsc --noEmit` clean.

burakeyler:
Merged `upstream/main` in to resolve the conflict / bring the branch up to date (`c16a7a67`). `main` had independently added `baseUrl`/`imageMode` resolution to `SafeMarkdown` (used by `github-autofix-feedback.tsx`); the conflict was in `safe-markdown.tsx` where that overlapped with this PR's `linkRepositoryFiles` opt-in. Kept both: the `a` renderer now branches on `linkRepositoryFiles` first (repository-file button behavior, unchanged) and otherwise falls through to the pre-existing `baseUrl`-resolving link/img logic — the two features don't share any call sites (repository-file linking is session-only, `baseUrl`/`imageMode` are autofix-feedback-only), so no cross-wiring was needed.

Verified after the merge:
- `vitest run` on `safe-markdown.test.tsx`, `diff-file-links.test.ts`, `session-timeline.test.tsx` — 60 passed
- `tsc --noEmit` on `packages/web` — clean (after rebuilding `packages/shared`'s stale `dist/`, unrelated to this change)
- `eslint` on the resolved file — clean

@ColeMurray your changes-requested review didn't have a comment body — if there's feedback beyond the conflict and the five points from the earlier automated review (already addressed in 1b4d0b2b), could you point me at it?

ColeMurray:
thanks! https://github.com/ColeMurray/background-agents/pull/2032

PR #1896, opened 2026-09-14
fix(scripts): verify the D1 migration ledger after applying

rhlsthrm:
`scripts/d1-migrate.sh` can finish reporting success over a database whose schema is only partially migrated.

## The gap

Steps 1–3 trust each `wrangler` invocation. A lost client response, a provider that reports success on a partially applied batch, or any future early return leaves a migration unrecorded. Terraform's `null_resource` then records a successful deploy on top of a half-migrated database — and the next run skips nothing and repairs nothing, because the ledger is the only thing it consults. The failure is silent and sticky.

Step 3 already guards the *inverse* direction (it refuses to apply when a version is recorded under a different filename), so the script cares about ledger fidelity; it just never confirms the ledger after writing to it.

## The change

Re-read the ledger from the database after applying and compare it against the migration files on disk, failing the deploy on any divergence. Both directions are reported separately, because they mean different things:

- **expected but not recorded** — a migration silently did not land;
- **recorded but no such file** — the database carries schema this checkout cannot account for.

The comparison reuses `jq`, already a dependency of steps 2 and 3 (lines 62 and 83), so it adds no new tooling requirement.

## Verification

Exercised against a stubbed `wrangler` across three cases:

| Case | Ledger after applying | Result |
|---|---|---|
| Consistent | both migrations recorded | exit 0, `Verified: ledger matches all 2 migration file(s).` |
| Silent partial | `0002` missing | **exit 1**, names `0002_beta.sql` as missing |
| Ghost row | extra `0003` with no file | **exit 1**, names `0003_gone.sql` as unexpected |

The middle case is the one that matters: without this guard the script printed `Done. Applied 2 migration(s).` and exited 0 while the ledger was missing a migration it had just claimed to apply.

`bash -n` is clean and `shellcheck` reports nothing on the added block (it flags only two pre-existing `SC2001` style hits at lines 40 and 79).

## Relationship to #1819

This is the surviving half of #1819, which was closed as superseded by #1870. #1870 replaced the owner-bootstrap Wrangler parsing and result handling, which addressed the other half of that PR; this ledger postcondition in `d1-migrate.sh` is a separate concern and is still absent. Submitting it standalone so it can be judged on its own.

## Summary by CodeRabbit

- **Bug Fixes**
  - Database migration deployments now verify that applied migrations match the migration files.
  - Deployments fail with clear details when migrations are missing, unexpected entries are detected, or recorded versions differ from the expected files.
  - Successful deployments report the number of migrations verified.

- **Tests**
  - Added automated coverage for matching migration records and common mismatch scenarios.
  - Migration verification checks now run as part of continuous integration.


rhlsthrm:
Updated to `f4b0ba84`: the postcondition now ships with a permanent test instead of the throwaway harness I described above.

`scripts/d1-migrate.test.mjs` follows the pattern `scripts/deploy-aws.test.mjs` already uses for its rollback path — stub the CLI on `PATH` and drive the failure deterministically. The stub answers step 2's ordered read with an empty ledger so every file on disk is applied, then serves the case's ledger to the postcondition; that isolates the assertion from the apply loop, so a case can only fail because the final ledger disagrees with the files.

Four cases: a consistent ledger passes and emits the `Verified:` line; a ledger missing a just-applied migration fails and names it; a ledger carrying a row with no file fails and names it; and a version recorded under a different filename reports in both directions.

**The test was falsified against the unguarded script.** Replacing `scripts/d1-migrate.sh` with its pre-change version fails all four cases, including the consistent one. With the guard restored, all four pass.

Wired as `npm run test:d1-migrate` and added to the CI job beside the sibling script tests (`test:lint-sql-portability`, `test:deploy-aws`, `test:rbac-bootstrap-owner`). Prettier clean on all four touched files; `ci.yml` still parses.

Issue #1663, opened 2026-08-29
RFC: Deterministic execution guardrails for unattended sessions (secrets + shared App tokens in sandbox)


grimdalltech:
## The gap

The README's security model is refreshingly honest, but reading it against the automations feature, there's a gap I'd love maintainer thoughts on:

- Secrets are decrypted and injected as env vars at spawn (encrypted at rest, plaintext in-sandbox).
- The git credential helper brokers short-lived **shared GitHub App installation tokens** into the sandbox.
- Automations (webhooks, Sentry alerts, cron) start sessions **with no human in the loop**, and inbound payload content reaches the agent as prompt context.

So a malicious webhook/Sentry/Linear payload = indirect prompt injection into a sandbox holding (a) decrypted org secrets and (b) git credentials valid for every repo the App is installed on — with port tunnels and browser automation as exfil channels.

The single-tenant trust assumption covers *users*, but not *content arriving at users' agents*.

## Proposal (optional, default-off)

A deterministic pre-execution policy layer in `sandbox-runtime`:
1. **Secret denial** — block command args/env reads targeting injected secrets + credential helper.
2. **Egress allowlist** — fail-closed, subdomain-aware; curl/wget/tunnels restricted to known infra domains.
3. **Git remote allowlist** — push/fetch only to configured SCM hosts.
4. **Per-session spend cap** — including child-session fan-out.
5. **Signed block receipts** — emitted to the session event stream for audit.

Shadow mode by default, hard blocks via flag. Zero behavior change when off.

## Question for maintainers

Interest in this as an optional hardening layer? Happy to contribute the implementation (tests + docs included) if the direction is welcome — or open to pointers if you'd rather solve it at the sandbox-infra network level instead.

(Hardening proposal, not a vuln disclosure — anything sensitive routes via SECURITY.md.)

Silentpartnercoding:
The threat boundary is real. One adjustment seems necessary if `hard` mode is intended to claim containment: item 1 cannot be a deterministic hard block inside `sandbox-runtime` for a provider that has already injected the secret into an unrestricted agent process. Direct `os.environ` access or a helper subprocess bypasses command inspection, so a signed block receipt would prove only the paths the wrapper observed.

[...]

Issue #1266, opened 2026-08-04
[Feature]: Custom home-screens to lock down repo selection


Blue101black:
As a marketing person, I want to just visit open-inspect to make changes to my website and I don't care about doing anything else.

As a sales person, I want to visit open-inspect to generate a sales document and I don't care about doing anything else.

As customer success, I want to visit open-inspect to ask questions about codebase, make changes to website, generate a sales document.

I think this lends itself to having some ability to have custom home-screens that generate specific url's that will lock down the repo selection so it takes away the confusion for users.

Should filter side-bar sessions to only ones that have that repo as well etc.

Issue #856, opened 2026-06-29
Feature request: single-tenant multi-GitHub-or repo access with platform auth gate


rubenlangeweg:
## Why
A single instance deployment should be able to work across selected repositories in multiple Github orgs without becoming full multi-tenant. 

This matters for agencies, internal platform teams and delivery teams that do not own every repo they work in. 

Concrete example: you own a own GitHub org, but the client projects often live in client-owned GitHub orgs. The team using Open-Inspect is still trusted staff. The problem is not client-users access. The problem is that the repos are spread across multiple GitHub App installations. 

Today, the system assumes one deployment-wide GitHub App installation. That works when all repos live under one org. It breaks down when a trusted team needs one Open-Inspect setup, one agent setup, and one automation surface. 

## Target shape
Keep single-tenant. 

Only trusted members of the deployment should be able to sign in. GitHub should become the SCM access layer, not the main trust boundary. 

The platform auth layer can be Better Auth or an equiv. setup. Its job is to ansewr: 
- Is this person allowed to use this Open-Inspect deployment?

Github App installs answer a different question: 
- Which repos can this deployment access, and which installation token should be used for this app? 

## Usecases
- An agency runs one Open-Inspect deployment for its team, but client repos live in several client GitHub orgs. 
- A delivery leads wants to start sessions agains selected repos from multiple orgs without Switching deployments
- A scheduled automation needs to run agains curated repo set across orgs, with each child to run using the correct repo creds. 
- A GitHub webhook from a client repo should create a session using hat repo's installation context, not a global installation ID. 

## Proposed scope
 - Support multiple GitHub App installations in one single-tenant deployment.
  - Store GitHub App `installation_id` on repository catalog entries.
  - Persist the repo's installation context on sessions.
  - Use the session's installation context for repo lookup, branch listing,
  push auth, PR creation fallback, credential-helper tokens, webhooks, repo
  images, and automation matching.
  - Replace global repo caching with installation-aware or catalog-aware
  caching.
  - Gate product access through platform auth, for example Better Auth with
  Google or GitHub login plus rb2 membership checks.
  - Keep GitHub user OAuth as optional SCM identity for PR attribution, not as
  the only product identity model.

## Out of scope
- No full multi-tentant client workspace model
- No client users logging into the same deployment
- No mixed-provider deployment support beyond what is needed for this GitHub multi-installation path. 

## Security direction
The trusted-user assumption remains: admitted users are allowed to use the deployment.

[...]

PR #1901, opened 2026-09-14
feat: opt-in repository access gate at session creation (ENFORCE_REPO_ACL)

shrwnsan:
## Summary
Community asks #856 (repos spanning multiple GitHub App installations), #1266 (lock down repo selection per persona), and #1663 (guardrails for unattended sessions) all point at the same gap: any authenticated user can start a session against any repo the installation can see. This PR adds an **opt-in** repository access gate evaluated at session creation, before any session or sandbox state exists.

- New env flag `ENFORCE_REPO_ACL` (default **off** — unset means byte-for-byte identical behavior; critical for zero-risk adoption)
- When enabled with `REPO_ACL_ALLOWLIST="owner/repo,org/*"` (comma-separated; `*` wildcards per segment; bare `*` = allow all; entries without `/` rejected as typos), every requested repository must pass before the session is created
- Enforcement lives in `SessionInitHandler` as an optional predicate: denial returns `403` after request-shape validation (malformed requests keep their 400s) and before any session/sandbox/participant write — a rejected request leaves nothing to clean up. Covers both the scalar repo fields and every entry of the `repositories` list
- Allowlist parsing/matching is a small pure module (`repos/repo-acl.ts`), case-insensitive, GitLab-subgroup aware; `env-validation.ts` fails loudly at startup if enforcement is on with an empty or malformed allowlist
- No live GitHub API calls in the request path — deterministic and unit-testable; a future per-user grant layer can replace the predicate without touching the handler

## Testing
- `npx vitest run src/repos/repo-acl.test.ts src/env-validation.test.ts src/session/http/handlers/session-init.handler.test.ts` — 50/50 passed (flag-off passthrough, allowed proceeds, denial → 403 with zero writes, scalar + list coverage, wildcard matrix, startup validation)
- `npm run typecheck -w @open-inspect/control-plane` — clean
- `npx eslint` on changed files — clean

Refs #856, #1266, #1663.

## Summary by CodeRabbit

- **New Features**
  - Added optional repository access control for session initialization.
  - Configure enforcement with `ENFORCE_REPO_ACL` and a comma-separated `REPO_ACL_ALLOWLIST`.
  - Supports wildcard owners or repository names and case-insensitive matching.
  - Denied repositories receive a 403 response before session-related state is created.
  - Added configurable sandbox boot timeouts with `SANDBOX_BOOT_TIMEOUT_MS`.

- **Bug Fixes**
  - Invalid access-control configuration now fails clearly when enforcement is enabled without an allowlist.
  - Malformed allowlists no longer prevent startup when enforcement is disabled.


shrwnsan:
Addressed in 3c5a8e5:

**Allowlist parsed while the gate is off (major)** — `parseRepoAclConfig` ran `parseRepoAclAllowlist` unconditionally, so a stale or malformed `REPO_ACL_ALLOWLIST` failed startup even with `ENFORCE_REPO_ACL` disabled. The parser now short-circuits when enforcement is off — `{ enforce: false, allowlist: [] }` regardless of the allowlist's content — and validation only runs when the gate is enabled, where an empty or malformed allowlist still refuses to start. Session behavior is unchanged: the ACL components only consult the allowlist when `enforce` is true.

Tests: gate-off returns an empty allowlist for valid and malformed values (new case: malformed allowlist ignored while off); all gate-on contracts re-verified. 51/51 pass, `tsc` + eslint clean.

Issue #43, opened 2026-02-01
Agent defaults to Haiku 4.5


johangithub:
Even when choosing a different model (e.g., GLM 4.7), it defaults to Haiku 4.5. 

https://github.com/user-attachments/assets/2395c4d0-ec2c-496e-b5d3-c247d79b6fe3

The issue is in `page.tsx`
`const [selectedModel, setSelectedModel] = useState("claude-haiku-4-5");`

`selectedModel` does not sync with `sessionState.model` when the session loads.

Issue #42, opened 2026-02-01
Dark Mode Readability


johangithub:
Hard to read the agent response with the current color setup.
<img width="844" height="562" alt="Image" src="https://github.com/user-attachments/assets/c1a1c44e-6886-4aab-9934-9f18af174281" />

I'll submit a PR shortly

I'm on MacOS Chrome 143

ColeMurray:
this should be resolved with:
https://github.com/ColeMurray/background-agents/pull/30

please pull in the latest and redeploy vercel

PR #1914, opened 2026-09-15
feat(automations): bound concurrent runs per automation

rhlsthrm:
## What

`automations.max_concurrent_runs` turns the schedule/manual overlap guard from a boolean into a bound.

Today a schedule or manual firing is admitted only when the automation has **no** active run. That makes an automation's throughput the slower of its cron cadence and one run per cadence, so a queue-draining automation whose runs outlast its interval can never work more than one item at a time however deep the queue behind it.

The new column defaults to `1`, which reproduces the current behaviour exactly, and every existing automation keeps it. Raising the limit is an explicit per-automation decision about a workload that is safe to overlap — it is never inferred from a trigger type or a cadence.

## Why the predicate counts invocations, not runs

A fan-out automation posts one `automation_runs` row per target (up to ten). Counting rows would refuse a ten-repository automation's *second* firing at a limit of three while nothing overlapping had actually happened. The predicate therefore counts `DISTINCT invocation_id` over the automation's active runs, so the bound is about firings in flight; the tick's own child-launch budget is what paces their sessions.

It keeps the `WHERE NOT EXISTS (...)` shape, so the conditional insert stays atomic against a racing firing. It also reads the bound from the `automations` row **inside the same statement** rather than taking it as a parameter: a firing resolves repositories, authorization and provider auth between reading the automation row and reaching the insert, and a PATCH lowering the bound during that window would otherwise be overtaken by a firing still comparing against the old value.

`MAX(col, 1)` plus a `COALESCE` means a row carrying zero, a negative, or a deleted automation reads as the serialized default rather than as *no* bound. `automationConcurrencyLimit()` is the TypeScript twin of that, for the cheap pre-check.

## Auto-pause threshold 3 -> 5

Sized against concurrency rather than against a serialized automation. Above a limit of one, a shared cause (bad base branch, expired credential) can fail every firing in flight at once, and those strikes land consecutively with no success between them to reset the streak. Three strikes is one such round; five requires the failures to outlast it. Docs updated to match.

## Migration

`terraform/d1/migrations/0082_automation_max_concurrent_runs.sql`

`ALTER TABLE automations ADD COLUMN max_concurrent_runs INTEGER NOT NULL DEFAULT 1`, plus a partial index `idx_runs_active_by_automation ON automation_runs (automation_id, invocation_id) WHERE status IN ('starting','running')` so the new count is served by an index rather than a scan of the append-only runs table (status stays out of the key and in the partial clause as literals, per the reasoning in migration 0024).

Numbered 0082 rather than 0080: 0080 and 0081 are already taken by two other branches we have in
flight, so this number is reserved for this change and will not collide. The gap is harmless —
`scripts/d1-migrate.sh` sorts by filename and only rejects duplicate numeric prefixes, not gaps.

**Correcting an earlier version of this description, which offered a renumber-on-merge and claimed
nothing references the filename. That was wrong.** On current `main`, `scripts/d1-migrate.sh`
records the filename into `_schema_migrations.name` (line 99) and hard-errors when a version is
already recorded under a different name:

```
87  if [ "$RECORDED_NAME" != "$FILENAME" ]; then
88    echo "ERROR: version $VERSION is already recorded as $RECORDED_NAME." >&2
89    echo "Renumber this migration before applying it to this installation." >&2
```

This migration is already applied under this exact filename on at least one deployed installation,
so renaming it after the fact would trip that check on the next migrate run rather than just
producing a cosmetic gap. If you would rather it were renumbered, it needs a coordinated
`_schema_migrations` update on already-deployed installations — happy to do that, but as a
deliberate step rather than a silent rename.

## Scope

Schedule and manual (Trigger Now) firings only. Event firings continue to block per concurrency key; an automation-wide guard there would serialize unrelated events (PR #42 against PR #43). Lowering the setting never interrupts firings already in flight.

UI: a "Concurrent runs" select (1-10) on the schedule trigger section of the automation form, defaulting to 1.

## Gates

Run locally against this branch, each sequentially from the repo root:

| Gate | Result |
| --- | --- |
| `npm run build -w @open-inspect/shared` | pass |
| `npm run typecheck` | pass (0 errors) |
| `npm test -w @open-inspect/control-plane` | 293 files, 4393 tests pass |
| `npm run test:integration -w @open-inspect/control-plane` | 106 files, 1278 pass, 1 skipped |
| `npm test -w @open-inspect/shared` | 56 files, 912 tests pass |
| `npm test -w @open-inspect/web` | 203/204 files pass |
| `npm run lint` | pass |

The single web failure is `src/hooks/use-session-rename.test.tsx > useSessionRename > keeps the latest overlay until its authoritative title is cached`. It fails deterministically on pristine `main` at `dcd9677` as well, so it is not from this change. (Two further web files — `automation-form.test.tsx`, `slack-integration-settings.test.tsx` — timed out at the 5s default on one loaded run and pass on every repeat; `automation-form.test.tsx` was re-run 3x in isolation, 54/54 each time.)

## Falsification: the new tests bite

The new tests are worthless if they pass with the enforcement removed, so both enforcement points were disabled and the suites re-run:

- `AutomationStore.overlapPredicate` schedule/manual arm -> `SELECT 1 WHERE 0 AND (...)` (never blocks)
- `Scheduler` pre-check -> `: false && (await store.countActiveInvocations(...)) >= automationConcurrencyLimit(...)`

8 integration tests then failed:

```
automation-invocations.test.ts > insertInvocationGuarded
  > admits up to the automation's bound and refuses past it
  > reads the bound at insert time, so a lowered limit is not overtaken by a firing in flight
  > treats a bound below one as the serialized default rather than as no bound
  > suppresses the invocation and children when an active run exists, but still advances
scheduler.test.ts > Scheduler (integration)
  > scheduled tick > skips overdue automations with active runs (concurrency guard)
  > scheduled tick > skips an overdue firing at the automation's concurrency limit
  > manual trigger > admits exactly one run across two triggers and a concurrent tick
  > manual trigger > rejects when active run exists
```

and 2 unit tests:

```
src/scheduler/scheduler.test.ts > Scheduler
  > tick > records an atomic childless skip when a run is active (concurrency guard)
  > trigger > rejects when active run exists, recording nothing
```

Both suites return to fully green when the enforcement is restored. The four pre-existing tests in that list are the ones that pin the default-1 behaviour, so the falsification also shows the old guarantee is still enforced through the new code path.

The three new tests that did **not** fail under sabotage are the positive-admission ones — `admits an overdue firing below the automation's concurrency limit`, `starts at most one firing per tick however much headroom the limit leaves`, and `bounds invocations rather than runs, so one fan-out firing cannot exhaust the limit`. An always-allow guard admits those by construction; they guard the other direction — that the bound does not over-refuse, and that concurrency builds one interval at a time rather than all at once.

## Summary by CodeRabbit

- **New Features**
  - Added a configurable **Concurrent runs** setting for scheduled and manual automation triggers, supporting 1–10 simultaneous firings.
  - Added a concurrency selector to automation forms; existing automations default to one concurrent run.
  - Concurrency is counted per firing, including fan-out automations.

- **Bug Fixes**
  - Updated automation auto-pause behavior to occur after five consecutive failures instead of three.
  - Changes to concurrency limits apply to new firings without interrupting runs already in progress.

- **Documentation**
  - Updated automation documentation with concurrency behavior, limits, skipped-trigger handling, and the revised auto-pause threshold.


rhlsthrm:
Workflow runs on this PR are `action_required` (fork PR, awaiting maintainer approval), so CI has not executed here. Every gate was run locally against this exact head (`f9d7123`), sequentially, from a clean worktree checked out at `main` = `dcd9677` with the commit cherry-picked on top:

```
npm run build -w @open-inspect/shared              pass
npm run typecheck                                  pass, 0 errors
npm test -w @open-inspect/control-plane            293 files / 4393 tests pass
npm run test:integration -w @open-inspect/control-plane
                                                   106 files / 1278 pass, 1 skipped
npm test -w @open-inspect/shared                   56 files / 912 tests pass
npm test -w @open-inspect/web                      203/204 files pass
npm run lint                                       pass
```

The one web failure, `use-session-rename.test.tsx > keeps the latest overlay until its authoritative title is cached`, reproduces on pristine `main` at `dcd9677` and is unrelated to this change.

The integration suite applies `terraform/d1/migrations/` automatically, so `0082_automation_max_concurrent_runs.sql` is exercised by all 106 integration files, not merely parsed.

The falsification detail for the new tests (which specific tests fail when both enforcement points are disabled, and which do not and why) is in the PR description.

PR #1922, opened 2026-09-15
Use Better Auth credentials for bot-created GitHub PRs

### LINKED ISSUE #1912: [Bug]: Slack / Linear not using author's Github credentials
Noticed linear issue I delegated was making the pr as open-inspect instead of my credentials, even though in the web open-inspect version it had my name as the original prompt.

Fable 5.1 yap below:

**Cause: a bug, not config.** Your Linear session was correctly attributed to you (the linear-bot passes the Linear creator, and the control plane links it to your web user by email). The break is credential lookup. For bot principals the control plane uses the "legacy" authority, and `resolveGitHubEnrichment` in [[identity.ts](https://claude.ai/epitaxy/packages/control-plane/src/session/identity.ts)](packages/control-plane/src/session/identity.ts) reads OAuth tokens from the `user_scm_tokens` table. Since web sign-in moved to Better Auth, GitHub tokens are stored in Better Auth's account table (`user_identities.access_token`) and nothing writes `user_scm_tokens` any more. So every Linear, Slack, or GitHub-bot session started with no user token, `resolveAuthForPR` returned null, and the PR was created with the App installation token.

ColeMurray:
Closes #1912

## Summary

- route Slack, Linear, and GitHub bot service principals through Better Auth's trusted server API using the admitted canonical user and linked GitHub account
- validate the returned GitHub profile before attaching attribution or a re-encrypted current access token to a session
- resolve a current Better Auth access token again at PR creation, while retaining the participant-local token path only for already-running pre-cutover sessions
- remove the legacy `UserScmTokenStore` runtime implementation and centralized `user_scm_tokens` refresh path
- retain the legacy table migration temporarily so its production row can be removed and validated independently before a later schema cleanup

## Security

- credentials are selected with `{ providerId, accountId, userId }`, where the canonical user comes from route admission rather than request payload data
- multiple linked GitHub accounts and provider-profile substitution are rejected
- Better Auth remains the sole shared refresh-token owner; new session state receives only the current access token

## Rollout Note

Deployments with GitHub grants that exist only in `user_scm_tokens` must have those users sign in with or re-link GitHub so Better Auth owns their current grant. The production audit for this issue found one legacy row, and that user already has Better Auth GitHub credentials.

## Verification

- `npm run lint -w @open-inspect/control-plane`
- `npm test -w @open-inspect/control-plane` (292 files, 4,375 tests)
- `npm run test:integration -w @open-inspect/control-plane` (106 files, 1,271 passed, 1 skipped)
- targeted browser-auth integration verifies a service principal obtains the Better Auth GitHub token while `user_scm_tokens` is empty

---
*Created with [Open-Inspect](https://open-inspect-prod.vercel.app/session/1a135c54ddbe55f6c770df286e537998)*

## Summary by CodeRabbit

* **New Features**
  * Sessions resolve current GitHub credentials through linked authentication accounts.
  * Service-principal access is supported for GitHub credential resolution.
  * Pull-request authentication is resolved only when needed, with app authentication as a fallback.

* **Bug Fixes**
  * GitHub identity mismatches now stop session, prompt, and automation operations.
  * SCM credentials are no longer copied into new sessions or persisted during enrichment.

* **Documentation**
  * Updated encryption-key rotation guidance for credentials already copied into sessions.


packages/control-plane/src/session/identity.ts
```diff
@@ -82,62 +77,53 @@ const browserGitHubAccountInfoSchema = z.object({
...
-  const profile = browserGitHubAccountInfoSchema.parse(
-    await dependencies.getAccountInfo(selection)
+  const accountClient = getAccountClient();
+  let tokenResponse: unknown;
+  try {
+    tokenResponse = await accountClient.getAccessToken({ body: selection });
+  } catch (error) {
+    throw new BetterAuthGitHubTokenUnavailableError(error);
+  }
+  const token = betterAuthAccessTokenSchema.parse(tokenResponse);
```
ColeMurray: **[P1] Preserve app fallback for identities without an OAuth grant**

A bot-created GitHub identity can legitimately exist in `user_identities` without `access_token` or `refresh_token`. In that case Better Auth 1.6.25's `getAccessToken` resolves with `accessToken: ""`; it does not throw. This parse therefore raises a ZodError outside the retrieval-error wrapper, and `ParticipantService.resolveAuthForPR` turns it into HTTP 500 instead of using the GitHub App fallback. This regresses PR creation for GitHub-bot users who have never signed in, in deployments with GitHub OAuth configured.

Reproduced against real D1 and Better Auth by clearing the grant fields on an otherwise valid GitHub identity: the resolver rejects with `Too small: expected string to have >=1 characters`. Could we classify the documented empty-token result as unavailable credentials and return null, while retaining fail-closed handling for malformed responses and identity mismatches? Add coverage for a bot-only identity with no OAuth grant.

packages/control-plane/src/session/identity.ts
```diff
@@ -187,38 +169,30 @@ export async function resolveGitHubEnrichment(
...
   userId: string,
   authority: GitHubCredentialAuthority
 ): Promise<GitHubEnrichment | null> {
-  // One invariant for the whole boundary: both authorities encrypt with
-  // validated AES-256 material, regardless of which branch runs.
-  const tokenEncryptionKey = requireTokenEncryptionKey(env);
-  if (authority.kind === "legacy") {
-    return resolveGitHubEnrichment(env, db, userStore, userId);
-  }
+  const enrichment = await resolveGitHubEnrichment(userStore, userId);
```
ColeMurray, line 209: **[P2] Preserve GitHub commit attribution for fresh browser sign-ins**

Browser enrichment now uses only the canonical database metadata, but the Better Auth account mapping/adapter does not populate `user_identities.provider_login` on a fresh OAuth sign-in. Consequently this returns `scmUserId` but no `scmLogin` or GitHub noreply email. The web session-create route supplies no replacement SCM display fields, and `resolveParticipantGitIdentity` subsequently selects `agent-only` because the login is missing. New browser users therefore lose their Git commit attribution even though their GitHub account is authenticated.

Reproduced using the real callback integration: the previous `scmLogin: "octocat"` and `email: "583231+octocat@users.noreply.github.com"` assertions, removed in this PR, both receive undefined. Could we persist verified provider metadata or retain identity-only profile enrichment without copying tokens, and restore these integration assertions?

packages/control-plane/src/session/identity.ts
```diff
@@ -82,62 +77,53 @@ const browserGitHubAccountInfoSchema = z.object({
...
+    tokenResponse = await accountClient.getAccessToken({ body: selection });
+  } catch (error) {
+    throw new BetterAuthGitHubTokenUnavailableError(error);
+  }
+  const token = betterAuthAccessTokenSchema.parse(tokenResponse);
+  if (
+    token.accessTokenExpiresAt &&
+    token.accessTokenExpiresAt.getTime() <= Date.now() + GITHUB_ACCESS_TOKEN_EXPIRY_BUFFER_MS
+  ) {
+    return null;
```
ColeMurray: **[P2] Refresh near-expiry grants instead of silently changing PR attribution**

The installed Better Auth 1.6.25 implementation refreshes only when less than five seconds remain. This guard rejects everything within 60 seconds, leaving a window in which a valid, refreshable user grant returns null and `SessionPullRequestService` creates the PR with app authentication. Moving resolution after the branch push avoids the earlier lifetime problem, but this fallback still reproduces the wrong-author behavior the PR is intended to fix.

Reproduced with real D1/Better Auth by setting a valid grant's expiry to `Date.now() + 30_000` while retaining its refresh token: the resolver returns null. Could we refresh through Better Auth when the larger safety window requires it, rather than treating a refreshable grant as absent? The regression should assert user OAuth is retained for this case, not just assert the current null fallback.

ColeMurray (commented):
Review verdict: Request Changes.

Reviewed exact head `4b9260af62b0b5c8b45fa5230e7d380c8221ac89`. Shared build, control-plane typecheck, 438 focused unit tests, and 44 integration tests passed. Additional real-D1 probes using the installed Better Auth implementation reproduced the three correctness gaps in the inline comments below. Canonical account selection and removing token copying are sound improvements, but these regressions should be addressed before merging.

PR #1933, opened 2026-09-16
fix: preserve prebuilt images on transient spawn errors

ColeMurray:
## Summary
- only mark a selected prebuilt image restore as failed when `createSandbox` returns a permanent `SandboxProviderError`
- skip the inline base-image retry for transient provider failures and let the existing outer spawn failure path handle them
- add structured warning events with `error_type` for permanent restore failures and transient prebuilt spawn errors
- cover transient no-retry behavior and retain permanent fallback coverage for repo and environment images

## Testing
- `npm test -w @open-inspect/control-plane` (4,395 tests)
- `npm run lint -w @open-inspect/control-plane -- --no-warn-ignored src/sandbox/lifecycle/manager.ts src/sandbox/lifecycle/manager.test.ts`

Fixes COL-140.

---
*Created with [Open-Inspect](https://open-inspect-prod.vercel.app/session/1b25a4538dbf1babd95e47d874e6d550)*

## Summary by CodeRabbit

- **Bug Fixes**
  - Sandbox launches now retry from the base image only when a configured prebuilt or repository image is unavailable.
  - Other provider failures now fail startup directly instead of triggering an unnecessary retry.
  - Unavailable repository images now return a clear HTTP 410 response rather than a generic server error.
  - Improved error reporting for failed image restoration, including error type and message.
  - Added consistent handling of unavailable prebuilt images across supported providers.
  - Added coverage for image-provider and restore-failure scenarios.


packages/control-plane/src/sandbox/lifecycle/manager.ts
```diff
@@ -651,15 +652,24 @@ export class SandboxLifecycleManager implements SandboxLifecycle {
         result = await this.provider.createSandbox(createConfig);
       } catch (error) {
         if (!selectedImage) throw error;
-        // A provider restore failure is "no image" (design §7.3): fail the
-        // row so the cron rebuilds it and boot this session from base rather
-        // than failing the spawn. Unrelated create failures (quota, network)
-        // can false-positive here — the cost is one rebuild, and the base
-        // retry surfaces them through the normal failure path anyway.
+        if (!(error instanceof PrebuiltImageUnavailableError)) {
```
ColeMurray, line 655: This gate only reaps when a provider explicitly throws `PrebuiltImageUnavailableError`, and three of the four image-build providers now do — but `vercel` is in `IMAGE_BUILD_PROVIDER_IDS` too, and `vercel/provider.ts` has no mapping.

`VercelProvider.createSandbox` consumes `config.prebuiltImageId` (line 113) and its catch is an unconditional `throw this.classifyError("Failed to create Vercel sandbox", error)` (line 156). Vercel runtime snapshots expire (`VERCEL_SNAPSHOT_EXPIRATION_MS`, `docs/VERCEL_SANDBOX_PROVIDER.md`), so an expired prebuild goes 404 → `SandboxProviderError("permanent")` → rethrown here → the row stays `ready` → the rebuild cron never fires (it only rebuilds when there's no ready image) → every spawn for that scope hard-fails with the breaker incrementing. On `main` the same input degraded to a base boot, so this is a regression for Vercel rather than a pre-existing gap.

Suggest mirroring what the file already does for restore/snapshot/stop (lines 209, 234, 248–251) inside the `createSandbox` catch:

```ts
} catch (error) {
  if (error instanceof SandboxProviderError) throw error;
  if (config.prebuiltImageId && error instanceof VercelSandboxApiError && error.status === 404) {
    throw new PrebuiltImageUnavailableError("Vercel prebuilt snapshot is unavailable", error);
  }
  throw this.classifyError("Failed to create Vercel sandbox", error);
}
```

The pass-through on the first line matters on its own: without it any inner `SandboxProviderError` (including the new subclass) gets rewrapped by `classifyError` and this gate never sees it — same reason the OC change at `opencomputer-provider.ts:746` was needed.

A provider test along the lines of the new `modal-provider.test.ts` / `e2b-provider.test.ts` cases (404 with `prebuiltImageId` set → `PrebuiltImageUnavailableError`; 404 without it → plain permanent) would lock it in.

PR #1938, opened 2026-09-17
docs: correct r2_media_bucket_name to describe managed bucket

rhlsthrm:
## Problem

`r2_media_bucket_name` is documented as the way to point the deployment at a bucket that was
pre-created out-of-band, but the configuration cannot do that.

`terraform/environments/production/r2.tf` manages the bucket unconditionally — no `count`, no
`for_each`, no `lifecycle` block:

```hcl
resource "cloudflare_r2_bucket" "media" {
  account_id = var.cloudflare_account_id
  name       = var.r2_media_bucket_name != "" ? var.r2_media_bucket_name : "open-inspect-media-${local.name_suffix}"
  location   = var.r2_media_location
}
```

So the variable only changes the *name* of the bucket Terraform creates and owns; it never adopts an
existing one. An operator who follows the current advice gets a create against a name that already
exists and the apply fails. The advice also can't achieve its stated purpose: credentials that
cannot create R2 buckets still cannot apply this configuration, because the resource is still a
create.

Three places carried that framing:

- `terraform/environments/production/variables.tf` — "Set this when the bucket must be pre-created
  out-of-band (e.g. when the Terraform credentials cannot create R2 buckets)."
- `docs/GETTING_STARTED.md` deployment settings table — "Optional media bucket name override for a
  pre-created bucket".
- `terraform/environments/production/terraform.tfvars.example` — "Set this when the bucket must be
  pre-created out-of-band …". This file is the only one that goes on to document the required
  `terraform import`, so its own opening framing contradicted its own instructions.

## Change

Docs only. All three now say what the input does — it names the bucket Terraform creates and
manages — and point at the import step required to use a bucket that already exists.

In `terraform.tfvars.example` only the five-line opening framing was rewritten; the existing manual
setup requirements, the `wrangler r2 bucket create` snippet, the `terraform import` command, the
rename/data-loss warning, and the token-scope notes are unchanged, and no example value was altered
or uncommented.

The variable description cites the resource address from `r2.tf` and the import ID format documented
for `cloudflare_r2_bucket` in the pinned provider (`5.19.1`):
`<account_id>/<bucket_name>/<jurisdiction>`, with `default` as the ordinary jurisdiction — the same
form `terraform.tfvars.example` already used.

No variable name, default, type, resource, or behaviour changed. The documented default name
(`open-inspect-media-<deployment_name>`) is unchanged and still matches `r2.tf`, since `locals.tf`
sets `name_suffix = var.deployment_name`.

## Verification

- `npx prettier --check docs/GETTING_STARTED.md` → exit 0 ("All matched files use Prettier code
  style!"). The replacement table cell is shorter than the table's widest cell, so column alignment
  is untouched.
- Prettier does not cover `terraform.tfvars.example`: the `lint-staged` globs are `*.{ts,tsx}` and
  `*.{js,jsx,json,md,yaml,yml,css}`, and `prettier --check` on the file exits 2 with "No parser
  could be inferred".
- `cd terraform/environments/production && terraform fmt -check` → exit 0.
- `terraform init -backend=false` → exit 0, then `terraform validate` → exit 0 ("Success! The
  configuration is valid.").
- Re-read `r2.tf` and `locals.tf` to confirm the documented default still matches the expression.

## Backward compatibility

None affected. No input, output, or resource changed — only comment and description text. Existing
deployments plan and apply identically; the only visible difference is the text `terraform plan`
prints when prompting for the variable.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that Terraform creates and manages the R2 media bucket, using a default name when none is provided.
  * Documented that existing buckets must be imported into Terraform before applying.
  * Added guidance that renaming a managed bucket requires migration and can replace the bucket, destroying its objects.


ColeMurray (changes requested):
-

rhlsthrm:
Fixed in 55597a46 — the finding was right and my wording was wrong.

Verified rather than assumed: `cloudflare_r2_bucket`'s `name` carries
`stringplanmodifier.RequiresReplace()` in the provider schema
([`internal/services/r2_bucket/schema.go`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/v5.19.0/internal/services/r2_bucket/schema.go),
and `id` is just the bucket name), and `grep -rn prevent_destroy terraform/` shows the only
`prevent_destroy` in the repo is on the AWS control-plane bucket — `cloudflare_r2_bucket.media` has
none. So a name change on a live deployment is a destroy-and-create, not a rename.

The description now says the variable *selects* the managed bucket's name, that the name is
RequiresReplace with no `prevent_destroy` so changing it destroys the current bucket and its
objects, that objects must be migrated first, and that an existing bucket must be imported. I made
the same correction to the lead sentence in `terraform.tfvars.example`, which had inherited the same
"only changes the name" phrasing and contradicted the WARNING block a few lines below it.

PR #1951, opened 2026-09-18
feat: identify provider accounts by seat, not just subscription

isakfagerlund:
## Problem

Adding a second ChatGPT subscription reported **"Existing ChatGPT account reconnected"** instead of creating a new account.

A provider account is keyed on `external_account_id`, which comes from the `chatgpt_account_id` claim. On a ChatGPT Business/Enterprise plan that claim identifies the **workspace**, not the person — every seat in it carries the same value. So a colleague connecting resolved to the first seat's account. The dedup was working exactly as written; the identity was too coarse.

## Fix

Record the seat alongside the subscription and key uniqueness on the pair.

| Field | Claim | Scope |
|---|---|---|
| `external_account_id` | `chatgpt_account_id` | workspace — still sent as `ChatGPT-Account-Id` |
| `external_principal_id` *(new)* | `chatgpt_user_id` | the individual seat |

The seat claim appears in both the `id_token` and the `access_token`, so it survives the token refresh that re-validates identity on every exchange.

```mermaid
flowchart TD
    A[Seat authorizes device code] --> B{Account bound to this seat?}
    B -->|yes| C[Reconnect it]
    B -->|no| D{Account on this subscription<br/>with no seat recorded?}
    D -->|yes| E[Adopt it, backfill the seat]
    D -->|no| F[Create a new account]
```

### Before / after

| Scenario | Before | After |
|---|---|---|
| Second seat connects | merged into the first seat's account | own account, titled from its email |
| Existing account, seat connects | n/a | adopted and backfilled, not duplicated |
| Reconnect authorized by the wrong seat | **silently replaced the credential** | refused |
| Runtime API requests | workspace id in header | unchanged |

## Migration

`0076` adds the column and moves the unique index to `(provider, external_account_id, COALESCE(external_principal_id, ''))`. The `COALESCE` matters: SQLite treats NULLs as distinct, so without it an installation could accumulate several unadopted rows per subscription. Forward-only, and it cannot fail to build — the old index already guaranteed at most one live row per subscription.

Existing accounts keep working with a NULL seat until that seat next connects.

## Tests

- 3 claim-extraction tests, using the claim layout auth.openai.com actually issues for a Business seat
- 3 finalizer seat-routing tests (new seat / adoption / wrong-seat refusal)
- 7 integration tests running the real SQL in workerd against D1 with the migration applied

Full suite: typecheck clean, 4177 unit + 1236 integration passing, lint and SQL-portability clean. One unrelated pre-existing failure in `use-session-rename.test.tsx`, confirmed failing on `main` too.

## Summary by CodeRabbit

* **New Features**
  * Linear sessions now provide progress updates, tool results, completion reasons, stop confirmations, and conversation-aware follow-ups.
  * Added repository selection prompts, automatic issue delegation, session stopping on unassignment, and handling for additional Linear webhook events.
  * Added session-level analytics with token usage, searchable tables, sorting, and pagination.
  * Provider accounts now distinguish multiple seats within a shared subscription.
  * Increased default E2B template capacity to 8 CPUs and 8 GB memory.
* **Bug Fixes**
  * Improved handling of stuck Linear sessions, callback failures, stale sessions, and repository suggestions.


isakfagerlund:
Opened against the wrong repository by mistake — this belongs on the autarcgmbh fork. Closing.

PR #1952, opened 2026-09-18
refactor: define the automation list page size and name limit once in shared

rhlsthrm:
## The duplication

The automation-list page size and the automation name limit are each declared more than once, in
separate packages, for the same concept:

- `packages/web/src/hooks/use-automations.ts:11` — `const AUTOMATION_LIST_PAGE_SIZE = 25;`, used to
  build the `limit` query parameter the client sends to `GET /automations`.
- `packages/control-plane/src/routes/automation-list.ts:20` —
  `const DEFAULT_AUTOMATION_LIST_PAGE_SIZE = 25;`, the server-side default for that same parameter.
- `packages/control-plane/src/routes/automation-list.ts:22` —
  `const MAX_AUTOMATION_LIST_PAGE_SIZE = 100;`, the ceiling the limit schema refuses above.
- `packages/control-plane/src/routes/automation-validation.ts:42` — `const MAX_NAME_LENGTH = 200;`,
  applied to automation names in `automation-crud.ts` and to the list `search` parameter in
  `automation-list.ts`.

The page size is the concrete bug: the web client and the control plane hard-code `25` independently
and neither imports it. Change one and the client silently requests a page size the server no longer
defaults to, with nothing failing to say so.

## Why this shape

`AGENTS.md:109`, under the coding conventions:

> - **Define each default value exactly once.** Extract to a named constant and import everywhere.

And `packages/shared/src/types/automations.ts` already holds the sibling automation limits and
exports them from the types barrel:

- `MAX_AUTOMATION_REPOSITORIES`
- `MAX_AUTOMATION_INSTRUCTIONS_LENGTH`
- `MAX_AUTOMATION_INVOCATION_LIST_LIMIT`

So this follows the pattern already in the repo rather than introducing one. The three constants
above were simply the outliers.

## The change

Moves the three constants into `packages/shared/src/types/automations.ts` beside their siblings,
exports them from `packages/shared/src/types/index.ts`, and migrates every consumer:

| Constant | Was | Now |
| --- | --- | --- |
| `MAX_AUTOMATION_NAME_LENGTH` | `MAX_NAME_LENGTH` in `automation-validation.ts` | `shared/types/automations` |
| `MAX_AUTOMATION_LIST_PAGE_SIZE` | local in `automation-list.ts` | `shared/types/automations` |
| `DEFAULT_AUTOMATION_LIST_PAGE_SIZE` | local in `automation-list.ts` **and** `AUTOMATION_LIST_PAGE_SIZE` in `use-automations.ts` | `shared/types/automations` |

Consumers migrated: `automation-validation.ts` (definition removed), `automation-list.ts`,
`automation-crud.ts` (two guards plus their two error strings), `automation-list.test.ts`, and
`use-automations.ts`.

Clean cutover: the old local definitions are deleted, with no alias or re-export left behind.
`MAX_NAME_LENGTH` and the web-local `AUTOMATION_LIST_PAGE_SIZE` have zero remaining references
outside build output.

**No value changed.** 200 stays 200, 100 stays 100, 25 stays 25 — this is a move, not a retune. No
routes are added or altered, so the route-catalog and admission-matrix counts are untouched.

## Verification

Run locally on this branch:

- `typecheck` — `@open-inspect/shared`, `@open-inspect/control-plane` (all four projects, including
  `test/integration`), `@open-inspect/web` — all exit 0.
- `packages/shared`: `vitest run` — 57 files, 918 tests passed.
- `packages/control-plane`: `vitest run src/routes/automation-*.test.ts` — 6 files, 144 tests passed
  (there is no `automation-crud.test.ts`; the CRUD name-length guards are covered by
  `automation-create.test.ts` and `automation-update.test.ts`).
- `packages/control-plane`: `vitest run --config vitest.integration.config.ts
  test/integration/hono-route-catalog-conformance.test.ts
  test/integration/route-admission-matrix.test.ts` — 2 files, 9 tests passed, confirming the route
  counts did not move.
- `packages/web`: `vitest run src/hooks/use-automations.test.tsx` — 1 file, 6 tests passed.

Note that CI on this PR will sit at `action_required` until a maintainer approves the run, since it
comes from an outside contributor's fork; the local gate results above stand in for it in the
meantime.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized automation name validation for creation, updates, and search, preserving the 200-character limit.
  * Standardized automation list pagination with a default of 25 items and a maximum of 100 items.

* **Refactor**
  * Centralized automation validation and pagination settings to ensure consistent behavior across the application.


rhlsthrm:
Good catch, and it was internally inconsistent with this PR's own argument — fixed in `c264981e`.

The point of the PR is that these values should be defined once, so leaving the test asserting
`25`/`101`/`201` as literals undercut it. All five sites in
`packages/control-plane/src/routes/automation-list.test.ts` now derive from the shared constants:

- the default-pagination assertion and the repository-filter assertion use
  `DEFAULT_AUTOMATION_LIST_PAGE_SIZE` instead of `25`
- both invalid-limit rows use `String(MAX_AUTOMATION_LIST_PAGE_SIZE + 1)` instead of `"101"`
- the second invalid-search row uses `MAX_AUTOMATION_NAME_LENGTH + 1` instead of `201`, matching
  what the earlier search case already did

The boundaries are unchanged, since `DEFAULT_AUTOMATION_LIST_PAGE_SIZE = 25`,
`MAX_AUTOMATION_LIST_PAGE_SIZE = 100` and `MAX_AUTOMATION_NAME_LENGTH = 200` — so the derived
inputs still resolve to exactly 25, 101 and 201. A grep for the old literals in that file now
returns nothing.

Verified: `npx vitest run src/routes/automation-list.test.ts` → 1 file / 17 tests passed, and
`tsc --noEmit -p packages/control-plane/tsconfig.test.json` exits 0. The change is amended into the
single refactor commit so the PR stays one coherent change rather than a fix-up on top.

ColeMurray:
thanks!

PR #1954, opened 2026-09-18
fix(control-plane): always fail the sandbox when a fatal report's timeline write throws

rhlsthrm:
## The defect

In `packages/control-plane/src/session/http/handlers/sandbox.handler.ts`, `sandboxError` lands the
failed boot phase on the timeline and then fails the sandbox:

```ts
// sandbox.handler.ts:163-177 on fff23f94
if (phase !== undefined) {
  await this.sandboxEventProcessor.processSandboxEvent({ /* boot_progress, status: "failed" */ });
}
await this.failSandbox(result.data.error);
return Response.json({ status: "ok" });
```

If `processSandboxEvent` rejects, the rejection propagates out of the handler and `failSandbox`
never runs.

That call is genuinely rejectable. `processSandboxEvent` reads
`messageRepository.getProcessingMessage()` and then dispatches `boot_progress` to
`SandboxRuntimeEventHandler.handleBootProgress`, which does
`sandboxRepository.recordBootProgress(...)` — a bare `this.sql.exec(...)`
(`sandbox-repository.ts:185`) — followed by `persistSandboxEvent` → `eventRepository.createEvent`,
another SQL insert whose `data` column carries `JSON.stringify(event)` including the `outputTail`,
and then `messenger.broadcast`. Any storage error on that path throws.

The awaited timeline write arrived in #1925 (`5e96f3ff`, "admit the sandbox bridge before boot
completes"), the same batch as #1946 which put those phases and the failing script's output in the
web UI. Before that commit `failSandbox` was reached unconditionally on every accepted report.

## User-visible consequence

The sandbox reported a **fatal** error, so nothing else is going to. `failSandbox` is what fences
the generation, terminates it and lets the session retry. Skipping it leaves a dead-but-live
sandbox: the row stays in a live status, nothing terminates it, nothing re-drives the pending
prompt, and the user sees a session that has silently stopped making progress. The trigger is a
storage failure — precisely the situation where the runtime is already unhealthy. Timeline
bookkeeping should not be able to suppress the failure path.

## The change

The timeline write becomes best effort and is logged when it fails, following the established shape
in this codebase — `SandboxLifecycleManager.reportSandboxError` (`manager.ts:952-967`) already wraps
its `setLastSpawnError` persist in exactly this try/catch so a storage failure cannot also cost the
broadcast. Event name follows the neighbouring `event: "sandbox.error_ignored"` /
`event: "sandbox.error_persist_failed"` convention.

### try/catch-then-fail rather than try/finally

Both run `failSandbox`. They differ in what the caller observes.

`try/finally` re-throws after the `finally` block, so the sandbox gets a 5xx and never reaches
`return Response.json({ status: "ok" })`. The report *was* accepted and the consequential half
*did* run, but the runtime cannot tell that: it sees a failed request and re-sends the fatal
report, which now arrives at a sandbox the previous call just failed and is rejected by the
`isDeadSandboxStatus` guard as `{ status: "ignored" }`. Pure noise, plus the timeline error escapes
as an unstructured stack instead of a queryable log line.

`try/catch` keeps the response contract identical to the happy path: the report is accepted, the
sandbox is failed, and the operator gets one structured `sandbox.error_timeline_failed` warning
naming what was lost. Losing the timeline detail is a degraded failure screen; losing the failure
coordination is a stuck session. Those are not equal, so they are not treated equally.

`failSandbox`'s own rejection still propagates, unchanged.

## Regression test

Added to the existing `sandbox.handler.test.ts` using that file's `createHandler()` harness.
Written first and confirmed failing on unmodified `fff23f94`:

```
 FAIL  src/session/http/handlers/sandbox.handler.test.ts > SandboxHandler > fails the sandbox even when landing the fatal report on the timeline throws
Error: storage write failed
 ❯ src/session/http/handlers/sandbox.handler.test.ts:212:43
    210|     } as SandboxRow);
    211|     isValidSandboxToken.mockResolvedValue(true);
    212|     processSandboxEvent.mockRejectedValue(new Error("storage write fai…
       |                                           ^
    213|
    214|     const response = await handler.sandboxError(

 Test Files  1 failed (1)
      Tests  1 failed | 50 skipped (51)
```

The handler propagated the rejection, so `failSandbox` was never called and no response was
returned.

Falsified after the fix: reverting only the handler change (test untouched) reproduces the failure
above at `1 failed | 50 passed (51)`; restoring it returns to `51 passed`.

## Local gates

| Gate | Result |
| --- | --- |
| `npx vitest run src/session/http/handlers/sandbox.handler.test.ts` | 51 passed (51) |
| `npx vitest run src/session/sandbox-events/processor.test.ts` | 38 passed (38) |
| `npx tsc --noEmit` (packages/control-plane) | exit 0 |

CI will land `action_required` here since this is from an outside contributor; the gates above were
run locally on this branch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Sandbox failures are now handled reliably even when recording the related timeline event fails.
  - Sandbox termination continues as expected, while the underlying timeline error is logged for troubleshooting.

- **Tests**
  - Added coverage for sandbox error handling when timeline storage is unavailable.


ColeMurray (changes requested):
fix merge conflict

rhlsthrm:
Rebased onto `9c92c2a5`; the conflict is resolved.

It was a real semantic one, not just context: #1964 removed `outputTail` from this path, so the
branch's `processSandboxEvent` call no longer forwards it. The rebase keeps current `main`'s field
set and wraps only the try/catch around it.

`npm test -w @open-inspect/control-plane -- src/session/http/handlers/sandbox.handler.test.ts`: 52
passed. `tsc -p packages/control-plane` and `eslint` clean.

PR #1973, opened 2026-09-19
fix(slack): keep target clicks working across a deployment

rhlsthrm:
#1936 gave the target-clarification controls request-id-bearing block ids
(`target_picker:<uuid>`, `target_quick_picks:<uuid>`) and #1958 carried the dispatcher guard
forward:

```ts
const parsedRequestId = action.block_id
  ? parseTargetInteractionRequestId(action.block_id, selectionSource)
  : undefined;
if (action.block_id && !parsedRequestId) return;
```

A clarification message posted **before** that deployment carries the previous block id —
`repo_quick_picks` (`target-clarification.ts:303` at `ee3f0dc4~1`) — or, for the picker, a
Slack-generated one. Neither parses, so the guard returns and the click does nothing at all: no
session, no error, no log. `handleTargetSelection` still supports exactly this case; it falls back
to `getLegacyPendingRequest(env, channel, threadKey)` (`target-selection.ts:74`) when `requestId` is
undefined. The guard just never lets it run.

The window is every clarification control in flight at deploy time, and the failure is silent —
the user clicks a repository and the thread stays dead.

This refuses only a block id this code minted and then failed to parse, which keeps the mismatched
and malformed cases rejected, and lets an unrecognised block id through to the channel/thread
fallback.

Tests: two dispatcher cases, one for the legacy `repo_quick_picks` payload and one for a
Slack-generated block id, both asserting the selection is routed with `requestId: undefined`. Both
fail against the current guard (verified by reverting the one-line change: `2 failed | 3 passed`)
and pass with it. The existing cross-source rejection test still passes unchanged.

`npm test -w @open-inspect/slack-bot`: 494 passed. `tsc -p packages/slack-bot` and `eslint` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Slack interaction handling for quick-pick and picker selections that lack an embedded request ID.
  - Legacy quick-pick controls and Slack-generated interaction identifiers now continue through normal request resolution instead of being discarded.
  - These selections are correctly routed for target selection using the appropriate interaction source.


ColeMurray:
thanks!

